PR #190, opened 2021-02-02
Updated Socket facade so that it is now possible to start TLS *after* the socket has opened.

SalomonBrys:
This allows to have TLS over Tor (because the socks5 handshake happens **before** the TLS handshake).

Also (Romain):
- ElectrumClient
    - Simplify the ElectrumClient state machine (less states, less rounds).
    - Prevent multiple connection attempt on the ElectrumClient (edge case)
- Improve ElectrumClient / Peer connectivity
    - nullable `TcpSocket.Builder`, avoiding the use of a default one (e.g. Phoenix startup with Tor activated  doesn't want to run on a classic one "transparently")
    - `disconnect` functions close the socket that will end the connection job.
- Small fixes

tested with `eclair-kmp-tests` / `simulator` / `device`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -26,10 +26,10 @@ allprojects {
 val currentOs = org.gradle.internal.os.OperatingSystem.current()
 
 kotlin {
-    val ktorVersion: String by extra { "1.4.1" }
+    val ktorVersion: String by extra { "1.5.0" }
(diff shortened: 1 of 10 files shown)


robbiehanson:
These changes were merged independently via #196 & #347

PR #228, opened 2021-03-16
Phoenix-kmm issue #123: Missing transactions when closing channels

robbiehanson:
Phoenix-kmm [issue #123](https://github.com/ACINQ/phoenix-kmm/issues/123):

> When the user chooses to close (or force-close) a channel, there should be a corresponding transaction. From the user's perspective, any operation that alters the ledger should have a corresponding transaction.

This PR builds atop of PR #219, which is still pending at the moment. So this PR is blocked until 219 is merged.

Logic overview:

- When a channel transitions to `Closing`, and the channel has a non-zero balance, we inject an OutgoingTransaction into the ledger. The transaction is marked as `Pending`
- When the corresponding channel transitions to `Closed`, we mark the OutgoingTransaction as `Completed.Succeeded.OnChain`, and we record the corresponding txid's & fees.

In terms of code changes, we needed to add a new class to `OutgoingPayment.Details`:

```kotlin
data class ChannelClosing(
   val closingAddress: String,
   val isLocalWallet: Boolean, /*...*/
) : Details()
```

And we refactored `OutgoingPayment.Status` to include support for on-chain payments:

```kotlin
sealed class Status {
   object Pending : Status()
   sealed class Completed : Status() {
      abstract val completedAt: Long
      data class Failed(/*...*/) : Completed()
         data class OffChain(/*...*/) : Completed()
         data class OnChain(
            val txids: List<ByteVector32>,
            val claimed: Satoshi,
            override val completedAt: Long
         ) : Completed()
      }
   }
}
```

Here's the code that runs when a channel transitions to the `Closing` state:

<details>
<summary>Code changes in Channel.kt</summary><p>

```kotlin
this is ChannelStateWithCommitments && newState is Closing -> {
   val channelBalance = commitments.localCommit.spec.toLocal
   if (channelBalance > 0.msat) {
      val defaultScriptPubKey = commitments.localParams.defaultFinalScriptPubKey
      val localShutdown = when (this) {
         is Normal -> this.localShutdown
         is Negotiating -> this.localShutdown
         is ShuttingDown -> this.localShutdown
         else -> null
      }
      if (localShutdown != null && localShutdown.scriptPubKey != defaultScriptPubKey) {
         // Non-default output address
         val btcAddr = Helpers.Closing.btcAddressFromScriptPubKey(
            scriptPubKey = localShutdown.scriptPubKey,
            chainHash = staticParams.nodeParams.chainHash
         ) ?: "unknown"
         actions + ChannelAction.Storage.StoreChannelClosing(
            amount = channelBalance,
            closingAddress = btcAddr,
            isLocalWallet = false
         )
      } else {
         // Default output address
         val btcAddr = Helpers.Closing.btcAddressFromScriptPubKey(
            scriptPubKey = defaultScriptPubKey,
            chainHash = staticParams.nodeParams.chainHash
         ) ?: "unknown"
         actions + ChannelAction.Storage.StoreChannelClosing(
            amount = channelBalance,
            closingAddress = btcAddr,
            isLocalWallet = true
         )
      }
   } else /* channelBalance <= 0.msat */ {
      actions
   }
}
```

</p></details>

From the user's perspective:

- they had a balance of X on the channel
- the channel was closed (one way or another)
- they see an OutgoingTransaction with amount -X
- this balances the ledger

Of course, the user may not actually receive X on-chain:

- pending HTLC's could go either direction
- millisatoshi's get truncated
- dust gets ignored

I'm taking a simplistic approach right now, and simply recording all on-chain amounts that we successfully "claimed". Any difference between the original amount and the claimed amount is regarded as "fees".

<details>
<summary>Code changes in Channel.kt</summary><p>

```kotlin
private fun completeChannelClosing(additionalConfirmedTx: Transaction?): ChannelAction.Storage.CompleteChannelClosing {
   // We want to give the user the list of btc transactions for their outputs
   val txids = mutableListOf<ByteVector32>()
   var claimed = 0.sat
   additionalConfirmedTx?.let { confirmedTx ->
      mutualClosePublished.firstOrNull { it == confirmedTx }?.let {
         txids += it.txid
         // NB: this code could be much simpler if we knew the localScriptPubKey
         var expectedAmount = commitments.localCommit.spec.toLocal.truncateToSatoshi()
         if (commitments.localParams.isFunder) {
            // we had to pay the closingFees
            val inputAmount = commitments.commitInput.txOut.amount
            val outputAmount = it.txOut.map { it.amount }.sum()
            val feesAmount = outputAmount - inputAmount
            expectedAmount -= feesAmount
         }
         claimed += expectedAmount
      }
   }
   localCommitPublished?.let {
      val confirmedTxids = it.irrevocablySpent.values.toSet()
      val allTxs = listOfNotNull(it.commitTx, it.claimMainDelayedOutputTx) + it.htlcSuccessTxs + it.htlcTimeoutTxs
      val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
      txids += confirmedTxs.map { it.txid }
      claimed += confirmedTxs.map { it.txOut.map { it.amount }.sum() }.sum()
   }
   listOfNotNull(
      remoteCommitPublished,
      nextRemoteCommitPublished,
      futureRemoteCommitPublished
   ).forEach {
      val confirmedTxids = it.irrevocablySpent.values.toSet()
      val allTxs = listOfNotNull(it.commitTx, it.claimMainOutputTx) + it.claimHtlcSuccessTxs + it.claimHtlcTimeoutTxs
      val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
      txids += confirmedTxs.map { it.txid }
      claimed += confirmedTxs.map { it.txOut.map { it.amount }.sum() }.sum()
   }
   revokedCommitPublished.forEach {
      val confirmedTxids = it.irrevocablySpent.values.toSet()
      val allTxs = listOfNotNull(it.commitTx, it.claimMainOutputTx, it.mainPenaltyTx) + it.htlcPenaltyTxs + it.claimHtlcDelayedPenaltyTxs
      val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
      txids += confirmedTxs.map { it.txid }
      claimed += confirmedTxs.map { it.txOut.map { it.amount }.sum() }.sum()
   }
   return ChannelAction.Storage.CompleteChannelClosing(txids, claimed)
}
```

</p></details>

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
@@ -41,7 +41,8 @@ sealed class ChannelEvent {
         val localParams: LocalParams,
         val remoteInit: Init,
         val channelFlags: Byte,
-        val channelVersion: ChannelVersion
+        val channelVersion: ChannelVersion,
+        val channelOrigin: ChannelOrigin? = null
     ) : ChannelEvent() {
         init {
             require(channelVersion.hasStaticRemotekey) { "channel version $channelVersion is invalid (static_remote_key is not set)" }
@@ -91,6 +92,8 @@ sealed class ChannelAction {
         data class HtlcInfo(val channelId: ByteVector32, val commitmentNumber: Long, val paymentHash: ByteVector32, val cltvExpiry: CltvExpiry)
         data class StoreHtlcInfos(val htlcs: List<HtlcInfo>) : Storage()
         data class GetHtlcInfos(val revokedCommitTxId: ByteVector32, val commitmentNumber: Long) : Storage()
+        data class PaymentReceived(val amount: MilliSatoshi, val origin: ChannelOrigin?) : Storage()
+        data class PaymentSent(val amount: MilliSatoshi) : Storage()
     }
 
     data class ProcessIncomingHtlc(val add: UpdateAddHtlc) : ChannelAction()
@@ -152,7 +155,20 @@ sealed class ChannelState {
     fun process(event: ChannelEvent): Pair<ChannelState, List<ChannelAction>> {
         return try {
             val (newState, actions) = processInternal(event)
-            Pair(newState, newState.updateActions(actions))
+            val actions1 = when {
+                this is WaitForFundingCreated && newState is WaitForFundingConfirmed -> {
+                    actions + ChannelAction.Storage.PaymentReceived(pushAmount, channelOrigin)
+                }
+                // we only want to fire the PaymentSent event when we transition to Closing for the first time
+                this is WaitForInit && newState is Closing -> actions
+                this is Closing && newState is Closing -> actions
+                this is ChannelStateWithCommitments && newState is Closing -> {
+                    actions + ChannelAction.Storage.PaymentSent(this.commitments.localCommit.spec.toLocal)
+                }
+                else -> actions
+            }
+            val actions2 = newState.updateActions(actions1)
+            Pair(newState, actions2)
         } catch (t: Throwable) {
             handleLocalError(event, t)
         }
@@ -174,6 +190,7 @@ sealed class ChannelState {
                 it is ChannelAction.Message.Send && it.message is RevokeAndAck -> it.copy(message = it.message.copy(channelData = Serialization.encrypt(privateKey.value, this)))
                 it is ChannelAction.Message.Send && it.message is ClosingSigned -> it.copy(message = it.message.copy(channelData = Serialization.encrypt(privateKey.value, this)))
                 else -> it
+
             }
         }
         else -> actions
@@ -544,11 +561,9 @@ data class WaitForInit(override val staticParams: StaticParams, override val cur
                     // In order to allow TLV extensions and keep backwards-compatibility, we include an empty upfront_shutdown_script.
                     // See https://github.com/lightningnetwork/lightning-rfc/pull/714.
                     tlvStream = TlvStream(
-                        if (event.channelVersion.isSet(ChannelVersion.ZERO_RESERVE_BIT)) {
-                            listOf(ChannelTlv.UpfrontShutdownScript(ByteVector.empty), ChannelTlv.ChannelVersionTlv(event.channelVersion))
-                        } else {
-                            listOf(ChannelTlv.UpfrontShutdownScript(ByteVector.empty))
-                        }
+                        listOf(ChannelTlv.UpfrontShutdownScript(ByteVector.empty)) +
+                                if (event.channelVersion.isSet(ChannelVersion.ZERO_RESERVE_BIT)) listOf(ChannelTlv.ChannelVersionTlv(event.channelVersion)) else listOf<ChannelTlv>() +
+                                        if (event.channelOrigin != null) listOf(ChannelTlv.ChannelOriginTlv(event.channelOrigin)) else listOf()
                     )
                 )
                 val nextState = WaitForAcceptChannel(staticParams, currentTip, currentOnChainFeerates, event, open)
@@ -1091,6 +1106,7 @@ data class WaitForOpenChannel(
                                 return Pair(Aborted(staticParams, currentTip, currentOnChainFeerates), listOf(ChannelAction.Message.Send(Error(temporaryChannelId, err.value.message))))
                             }
                         }
+                        val channelOrigin = event.message.tlvStream.records.filterIsInstance<ChannelTlv.ChannelOriginTlv>().firstOrNull()?.channelOrigin
 
                         val fundingPubkey = keyManager.fundingPublicKey(localParams.fundingKeyPath).publicKey
                         val channelKeyPath = keyManager.channelKeyPath(localParams, channelVersion)
@@ -1143,6 +1159,7 @@ data class WaitForOpenChannel(
                             event.message.firstPerCommitmentPoint,
                             event.message.channelFlags,
                             channelVersion,
+                            channelOrigin,
                             accept
                         )
                         Pair(nextState, listOf(ChannelAction.Message.Send(accept)))
@@ -1181,6 +1198,7 @@ data class WaitForFundingCreated(
     val remoteFirstPerCommitmentPoint: PublicKey,
     val channelFlags: Byte,
     val channelVersion: ChannelVersion,
+    val channelOrigin: ChannelOrigin?,
     val lastSent: AcceptChannel
 ) : ChannelState() {
     override fun processInternal(event: ChannelEvent): Pair<ChannelState, List<ChannelAction>> {
@@ -1254,7 +1272,16 @@ data class WaitForFundingCreated(
                                         logger.info { "c:$channelId will wait for $fundingMinDepth confirmations" }
                                         val watchSpent = WatchSpent(channelId, commitInput.outPoint.txid, commitInput.outPoint.index.toInt(), commitments.commitInput.txOut.publicKeyScript, BITCOIN_FUNDING_SPENT)
                                         val watchConfirmed = WatchConfirmed(channelId, commitInput.outPoint.txid, commitments.commitInput.txOut.publicKeyScript, fundingMinDepth.toLong(), BITCOIN_FUNDING_DEPTHOK)
-                                        val nextState = WaitForFundingConfirmed(staticParams, currentTip, currentOnChainFeerates, commitments, null, currentBlockHeight.toLong(), null, Either.Right(fundingSigned))
+                                        val nextState = WaitForFundingConfirmed(
+                                            staticParams,
+                                            currentTip,
+                                            currentOnChainFeerates,
+                                            commitments,
+                                            null,
+                                            currentBlockHeight.toLong(),
+                                            null,
+                                            Either.Right(fundingSigned)
+                                        )
                                         val actions = listOf(
                                             ChannelAction.Blockchain.SendWatch(watchSpent),
                                             ChannelAction.Blockchain.SendWatch(watchConfirmed),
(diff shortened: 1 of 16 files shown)


src/commonMain/kotlin/fr/acinq/eclair/channel/Helpers.kt
```diff
@@ -357,6 +357,46 @@ object Helpers {
 
         fun isValidFinalScriptPubkey(scriptPubKey: ByteVector): Boolean = isValidFinalScriptPubkey(scriptPubKey.toByteArray())
 
+        fun btcAddressFromScriptPubKey(scriptPubKey: ByteVector, chainHash: ByteVector32): String? {
```
t-bast, line 360: It will be slightly more painful because it's a different lib, but I think this should be done in https://github.com/acinq/bitcoin-kmp

It's generally useful to have for any bitcoin application IMHO.

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
```diff
@@ -152,7 +157,54 @@ sealed class ChannelState {
...
+                            is Negotiating -> this.localShutdown
+                            is ShuttingDown -> this.localShutdown
+                            else -> null
+                        }
+                        if (localShutdown != null && localShutdown.scriptPubKey != defaultScriptPubKey) {
+                            // Non-default output address
+                            val btcAddr = Helpers.Closing.btcAddressFromScriptPubKey(
+                                scriptPubKey = localShutdown.scriptPubKey,
+                                chainHash = staticParams.nodeParams.chainHash
+                            ) ?: "unknown"
```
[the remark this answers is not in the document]
t-bast: You will be able to use `ClosingTx.toLocalOutput` to get the closing address once #229 lands

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
```diff
@@ -179,17 +206,26 @@ data class OutgoingPayment(val id: UUID, val recipientAmount: MilliSatoshi, val
...
     sealed class Status {
         object Pending : Status()
-        data class Succeeded(val preimage: ByteVector32, val completedAt: Long = currentTimestampMillis()) : Status()
-        data class Failed(val reason: FinalFailure, val completedAt: Long = currentTimestampMillis()) : Status()
+        sealed class Completed : Status() {
+            abstract val completedAt: Long
+            data class Failed(val reason: FinalFailure, override val completedAt: Long = currentTimestampMillis()) : Completed()
+            sealed class Succeeded : Completed() {
+                data class OffChain(val preimage: ByteVector32, override val completedAt: Long = currentTimestampMillis()) : Completed()
+                data class OnChain(val txids: List<ByteVector32>, val claimed: Satoshi, override val completedAt: Long = currentTimestampMillis()) : Completed()
```
t-bast, line 226: Can you explain how you reconcile between `claimed` here and `recipientAmount` in `OutgoingPayment`?

I'm also thinking that we could maybe better integrate with the existing `OutgoingPayment` by leveraging the `parts` field: each tx made in the channel closing could be inside a `part`.

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
```diff
@@ -91,6 +93,9 @@ sealed class ChannelAction {
         data class HtlcInfo(val channelId: ByteVector32, val commitmentNumber: Long, val paymentHash: ByteVector32, val cltvExpiry: CltvExpiry)
         data class StoreHtlcInfos(val htlcs: List<HtlcInfo>) : Storage()
         data class GetHtlcInfos(val revokedCommitTxId: ByteVector32, val commitmentNumber: Long) : Storage()
+        data class StoreIncomingAmount(val amount: MilliSatoshi, val origin: ChannelOrigin?) : Storage()
+        data class StoreChannelClosing(val amount: MilliSatoshi, val closingAddress: String, val isLocalWallet: Boolean) : Storage()
+        data class CompleteChannelClosing(val txids: List<ByteVector32>, val claimed: Satoshi) : Storage()
```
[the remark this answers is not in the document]
robbiehanson: Done in 9ae3492

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
```diff
@@ -40,11 +41,22 @@ interface OutgoingPaymentsDb {
...
+            completeOutgoingPayment(id, OutgoingPayment.Status.Completed.Failed(finalFailure, completedAt))
+        } else {
+            completeOutgoingPayment(id, OutgoingPayment.Status.Completed.Failed(finalFailure))
+        }
+
+    suspend fun completeOutgoingPayment(id: UUID, preimage: ByteVector32, completedAt: Long? = null) =
+        if (completedAt != null) {
+            completeOutgoingPayment(id, OutgoingPayment.Status.Completed.Succeeded.OffChain(preimage, completedAt))
+        } else {
+            completeOutgoingPayment(id, OutgoingPayment.Status.Completed.Succeeded.OffChain(preimage))
```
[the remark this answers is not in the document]
robbiehanson: Done in 5cd71d4

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
```diff
@@ -179,17 +206,26 @@ data class OutgoingPayment(val id: UUID, val recipientAmount: MilliSatoshi, val
         /** Swaps out send a lightning payment to a swap server, which will send an on-chain transaction to a given address. */
         data class SwapOut(val address: String, override val paymentHash: ByteVector32) : Details()
 
+        data class ChannelClosing(val closingAddress: String, val isLocalWallet: Boolean, override val paymentHash: ByteVector32) : Details()
```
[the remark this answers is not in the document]
robbiehanson: Great idea !

When `isLocalWallet` is true, this means the bitcoin `closingAddress` is derived from the user's seed. In the UI, we want to be able to tell the user "This address belongs to you, and is derived from your seed ..." I suppose there are other ways we could implement it, but this seemed like the simplest & most straight-forward.

src/commonMain/kotlin/fr/acinq/eclair/channel/Helpers.kt
robbiehanson: The only argument I have against this is that, this code is lightning-specific. There's a related function `isValidFinalScriptPubkey`, which checks the script against the restrictions defined in Bolt 02. (Only 4 script formats allowed.) The new function `btcAddressFromScriptPubKey` is basically a repurpose of that function, which takes advantage of the limited script formats.

Also, as you mentioned below:

> You will be able to use `ClosingTx.toLocalOutput` to get the closing address once [#229](https://github.com/ACINQ/eclair-kmp/pull/229) lands

In which case this function would be removed.

t-bast: > The only argument I have against this is that, this code is lightning-specific.

I don't think it is, it covers the most used bitcoin script types. If some application using this translation from bitcoin-kmp needs another script type covered at some point, then it can be added to bitcoin-kmp and it benefits everyone. Since this PR will not land soon (IIUC we're freezing eclair-kmp soon for a release) I think it's worth taking this time to put it in bitcoin-kmp?

> In which case this function would be removed.

It won't remove this function, the `ClosingTx.toLocalOutput` will still give you a `scriptPubKey` that you'll need to translate to a bitcoin address. But it will make it a bit easier for you to get the info than matching the output against a stored `finalScriptPubKey`.

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
robbiehanson: > each tx made in the channel closing could be inside a `part`

It could be super helpful to have this information. And we don't have to force it into the `part` structure. Currently the database is implemented such that there are 2 separate tables:

- outgoing_payments
- outgoing_payment_parts

When the UI fetches the list of payments (_a frequent operation_), it gets a light-weight version of an OutgoingPayment without all the parts. If the UI needs the full OutgoingPayment, it requests it specifically from the database. So we could add a new table:

- outgoing_bitcoin_transactions

This wouldn't slow down the list-payments operation.

robbiehanson: > Can you explain how you reconcile between `claimed` here and `recipientAmount` in `OutgoingPayment`?

`recipientAmount` = User's channel balance at time of channel closing.

`claimed` = Sum total of confirmed bitcoin transaction outputs going to the user.



Within the UI we've been increasingly adopting separate modes. We might call them "basic" & "advanced". When viewing a `ClosingChannel` transaction in advanced mode, it would be ideal to list all of the corresponding bitcoin transactions, a link to them on a blockchain explorer, along with a description, and fees information. Not only would this be cool (at least, I think so), but it would be super helpful for debugging & troubleshooting issues with users.



But we also need a basic mode. From the average user's pespective, they had a balance of X in the channel. The channel was closed, and they received Y in their bitcoin wallet. The difference between X & Y is ... fees.



And "fees" is what I'm trying to capture in this first pass. If we can add the `outgoing_bitcoin_transactions`, some of this primitive work can be removed.

t-bast: Thanks, that makes sense. I think it's worth adding a comment to explain how `recipientAmount` and `claimed` are used.

> When the UI fetches the list of payments (a frequent operation), it gets a light-weight version of an OutgoingPayment without all the parts.

Great, then it really does make sense to store transactions in a separate table (either `outgoing_payment_parts` or `outgoing_bitcoin_transactions`, whichever you think makes more sense).

And I think it makes sense to extend the `OutgoingPayment.Part` for that new use-case, it will make these closing cases consistent with other payments.

robbiehanson:
Superseded by PR #232

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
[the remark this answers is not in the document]
robbiehanson: Done in 0560763

PR #230, opened 2021-03-18
Handle swap-in storage

dpad85:
This PR stores new channels as incoming payment in the payments database. It does not implement the actual database query.

As a side effect, the `IncomingPayment.Origin.SwapIn` object has been simplified (only the address field was needed).

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/eclair/db/InMemoryPaymentsDb.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/db/InMemoryPaymentsDb.kt
@@ -32,7 +32,7 @@ class InMemoryPaymentsDb : PaymentsDb {
     }
 
     override suspend fun addAndReceivePayment(preimage: ByteVector32, origin: IncomingPayment.Origin, amount: MilliSatoshi, receivedWith: IncomingPayment.ReceivedWith, createdAt: Long, receivedAt: Long) {
-        val paymentHash = Crypto.sha256(preimage).toByteVector32()
+        val paymentHash = preimage.sha256()
         incoming[paymentHash] = IncomingPayment(preimage, origin, IncomingPayment.Received(amount, receivedWith, receivedAt), createdAt)
     }
 
--- a/src/commonMain/kotlin/fr/acinq/eclair/db/InMemoryPaymentsDb.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/db/InMemoryPaymentsDb.kt
@@ -31,6 +31,11 @@ class InMemoryPaymentsDb : PaymentsDb {
         }
     }
 
+    override suspend fun addAndReceivePayment(preimage: ByteVector32, origin: IncomingPayment.Origin, amount: MilliSatoshi, receivedWith: IncomingPayment.ReceivedWith, createdAt: Long, receivedAt: Long) {
+        val paymentHash = Crypto.sha256(preimage).toByteVector32()
+        incoming[paymentHash] = IncomingPayment(preimage, origin, IncomingPayment.Received(amount, receivedWith, receivedAt), createdAt)
+    }
+
     override suspend fun listReceivedPayments(count: Int, skip: Int, filters: Set<PaymentTypeFilter>): List<IncomingPayment> =
         incoming.values
             .asSequence()
--- a/src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
@@ -29,6 +29,9 @@ interface IncomingPaymentsDb {
      */
     suspend fun receivePayment(paymentHash: ByteVector32, amount: MilliSatoshi, receivedWith: IncomingPayment.ReceivedWith, receivedAt: Long = currentTimestampMillis())
 
+    /** Add and receive a payment. Use this method when receiving a spontaneous payment, for example a swap-in payment. */
+    suspend fun addAndReceivePayment(preimage: ByteVector32, origin: IncomingPayment.Origin, amount: MilliSatoshi, receivedWith: IncomingPayment.ReceivedWith, createdAt: Long = currentTimestampMillis(), receivedAt: Long = currentTimestampMillis())
+
     /** List received payments (with most recent payments first). */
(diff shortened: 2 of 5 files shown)


dpad85:
Note: the preimage for the incoming payment stored in the database is `Crypto.sha256(actualChannelId).toByteVector32()`. On Phoenix, the primary key for that table is the payment hash, so the identifier will be 2 times sha256 of the channel id.

If Phoenix receives 2 swap-in on the same address, then two different payment lines will be added to the database, one for each new channel.

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
```diff
@@ -128,15 +131,15 @@ data class IncomingPayment(val preimage: ByteVector32, val origin: Origin, val r
     data class Received(val amount: MilliSatoshi, val receivedWith: ReceivedWith, val receivedAt: Long = currentTimestampMillis())
 
     sealed class ReceivedWith {
-        abstract val fees: MilliSatoshi
+        abstract val fees: MilliSatoshi?
 
         /** Payment was received via existing lightning channels. */
         object LightningPayment : ReceivedWith() {
-            override val fees: MilliSatoshi = 0.msat
+            override val fees: MilliSatoshi? = null // with Lightning, the fee is paid by the sender
```
t-bast, line 138: I'd rather make it a fee of 0 sat than a null, what do you think?

dpad85: I wanted to make it explicit that this fee is optional, not just = 0 sometimes. This is mostly useful for swap-ins where the peer may not send the `ChannelOrigin` object containing the fee, hence the `ReceivedWith` fee attribute becomes nullable.

And so I also updated the fee for incoming lightning payments where having a fee does not make sense.

t-bast: The issue I have with nullable here is that we don't know if it means:

1. There was no fee for this operation (it's free)
2. We're missing the information about the fee (but there could be one)

Whereas explicitly setting it to 0 sat makes it obvious where we say "this operation doesn't cost any fee" and we can easily tell whether the code is correct or if we're in fact not transmitting enough information to compute the fee correctly.

But if you still think nullable makes more sense it's your call!

dpad85: True, we would need an "unknown" option and a "do not apply" option to be really explicit. So let's keep it simple and use 0 msat instead.

dpad85: done in 419e904

src/commonMain/kotlin/fr/acinq/eclair/db/InMemoryPaymentsDb.kt
```diff
@@ -31,6 +31,11 @@ class InMemoryPaymentsDb : PaymentsDb {
         }
     }
 
+    override suspend fun addAndReceivePayment(preimage: ByteVector32, origin: IncomingPayment.Origin, amount: MilliSatoshi, receivedWith: IncomingPayment.ReceivedWith, createdAt: Long, receivedAt: Long) {
+        val paymentHash = Crypto.sha256(preimage).toByteVector32()
```
t-bast, line 35: nit:
```suggestion
        val paymentHash = preimage.sha256()
```

src/commonMain/kotlin/fr/acinq/eclair/io/Peer.kt
```diff
@@ -367,6 +367,40 @@ class Peer(
                 action is ChannelAction.Storage.StoreHtlcInfos -> {
                     action.htlcs.forEach { db.channels.addHtlcInfo(actualChannelId, it.commitmentNumber, it.paymentHash, it.cltvExpiry) }
                 }
+                action is ChannelAction.Storage.StoreIncomingAmount -> {
+                    logger.info { "storing incoming amount=${action.amount} with origin=${action.origin}" }
+                    when (action.origin) {
+                        null -> {
+                            logger.warning { "n:$remoteNodeId c:$actualChannelId incoming amount with empty origin, store minimal information" }
+                            val fakePreimage = Crypto.sha256(actualChannelId).toByteVector32()
```
t-bast, line 375: nit:
```suggestion
                            val fakePreimage = actualChannelId.sha256()
```

src/commonMain/kotlin/fr/acinq/eclair/io/Peer.kt
```diff
@@ -367,6 +367,40 @@ class Peer(
...
+                                db.payments.receivePayment(paymentHash = action.origin.paymentHash, amount = action.amount, receivedWith = IncomingPayment.ReceivedWith.NewChannel(
+                                    fees = action.origin.fee.toMilliSatoshi(),
+                                    channelId = actualChannelId
+                                ))
+                            } else {
+                                logger.warning { "n:$remoteNodeId c:$actualChannelId ignored pay-to-open storage, no payments in db for hash=${action.origin.paymentHash}" }
+                            }
+                        }
+                        is ChannelOrigin.SwapInOrigin -> {
+                            val fakePreimage = Crypto.sha256(actualChannelId).toByteVector32()
```
t-bast, line 394: nit:
```suggestion
                            val fakePreimage = actualChannelId.sha256()
```

t-bast (commented):
LGTM, just nits

src/commonMain/kotlin/fr/acinq/eclair/db/InMemoryPaymentsDb.kt
dpad85: done in ab8b728

src/commonMain/kotlin/fr/acinq/eclair/io/Peer.kt
dpad85: done in ab8b728

src/commonMain/kotlin/fr/acinq/eclair/io/Peer.kt
dpad85: done in ab8b728

t-bast (approved):

PR #231, opened 2021-03-18
kotlin 1.4.31

romainbsl:
- eclair-kmp 0.5.0
- bitcoin-kmp 0.7.0-SN
- Ktor 1.5.2
- Kx Serialization 1.0.1
- Kx-Coroutines 1.4.3-mt
- kodein-memory-files:0.7.0-kx.ser-1.0.1-55-SN
- Kodein-Log 0.10.1

PR #232, opened 2021-03-18
Phoenix-kmm issue #123: Missing transactions when closing channels

robbiehanson:
Phoenix-kmm [issue #123](https://github.com/ACINQ/phoenix-kmm/issues/123):

> When the user chooses to close (or force-close) a channel, there should be a corresponding transaction. From the user's perspective, any operation that alters the ledger should have a corresponding transaction.

Logic overview:

- When a channel transitions to `Closing`, and the channel has a non-zero balance, we inject an OutgoingTransaction into the ledger. The transaction is marked as `Pending`
- When the corresponding channel transitions to `Closed`, we mark the OutgoingTransaction as `Completed.Succeeded.OnChain`, and we record the corresponding txid's & fees.

In terms of code changes, we needed to add a new class to `OutgoingPayment.Details`:

```kotlin
data class ChannelClosing(
   val closingAddress: String,
   val isLocalWallet: Boolean, /*...*/
) : Details()
```

And we refactored `OutgoingPayment.Status` to include support for on-chain payments:

```kotlin
sealed class Status {
   object Pending : Status()
   sealed class Completed : Status() {
      abstract val completedAt: Long
      data class Failed(/*...*/) : Completed()
         data class OffChain(/*...*/) : Completed()
         data class OnChain(
            val txids: List<ByteVector32>,
            val claimed: Satoshi,
            override val completedAt: Long
         ) : Completed()
      }
   }
}
```

Here's the code that runs when a channel transitions to the `Closing` state:

<details>
<summary>Code changes in Channel.kt</summary><p>

```kotlin
this is ChannelStateWithCommitments && newState is Closing -> {
   val channelBalance = commitments.localCommit.spec.toLocal
   if (channelBalance > 0.msat) {
      val defaultScriptPubKey = commitments.localParams.defaultFinalScriptPubKey
      val localShutdown = when (this) {
         is Normal -> this.localShutdown
         is Negotiating -> this.localShutdown
         is ShuttingDown -> this.localShutdown
         else -> null
      }
      if (localShutdown != null && localShutdown.scriptPubKey != defaultScriptPubKey) {
         // Non-default output address
         val btcAddr = Helpers.Closing.btcAddressFromScriptPubKey(
            scriptPubKey = localShutdown.scriptPubKey,
            chainHash = staticParams.nodeParams.chainHash
         ) ?: "unknown"
         actions + ChannelAction.Storage.StoreChannelClosing(
            amount = channelBalance,
            closingAddress = btcAddr,
            isLocalWallet = false
         )
      } else {
         // Default output address
         val btcAddr = Helpers.Closing.btcAddressFromScriptPubKey(
            scriptPubKey = defaultScriptPubKey,
            chainHash = staticParams.nodeParams.chainHash
         ) ?: "unknown"
         actions + ChannelAction.Storage.StoreChannelClosing(
            amount = channelBalance,
            closingAddress = btcAddr,
            isLocalWallet = true
         )
      }
   } else /* channelBalance <= 0.msat */ {
      actions
   }
}
```

</p></details>

From the user's perspective:

- they had a balance of X on the channel
- the channel was closed (one way or another)
- they see an OutgoingTransaction with amount -X
- this balances the ledger

Of course, the user may not actually receive X on-chain:

- pending HTLC's could go either direction
- millisatoshi's get truncated
- dust gets ignored

I'm taking a simplistic approach right now, and simply recording all on-chain amounts that we successfully "claimed". Any difference between the original amount and the claimed amount is regarded as "fees".

<details>
<summary>Code changes in Channel.kt</summary><p>

```kotlin
private fun completeChannelClosing(additionalConfirmedTx: Transaction?): ChannelAction.Storage.CompleteChannelClosing {
   // We want to give the user the list of btc transactions for their outputs
   val txids = mutableListOf<ByteVector32>()
   var claimed = 0.sat
   additionalConfirmedTx?.let { confirmedTx ->
      mutualClosePublished.firstOrNull { it == confirmedTx }?.let {
         txids += it.txid
         // NB: this code could be much simpler if we knew the localScriptPubKey
         var expectedAmount = commitments.localCommit.spec.toLocal.truncateToSatoshi()
         if (commitments.localParams.isFunder) {
            // we had to pay the closingFees
            val inputAmount = commitments.commitInput.txOut.amount
            val outputAmount = it.txOut.map { it.amount }.sum()
            val feesAmount = outputAmount - inputAmount
            expectedAmount -= feesAmount
         }
         claimed += expectedAmount
      }
   }
   localCommitPublished?.let {
      val confirmedTxids = it.irrevocablySpent.values.toSet()
      val allTxs = listOfNotNull(it.commitTx, it.claimMainDelayedOutputTx) + it.htlcSuccessTxs + it.htlcTimeoutTxs
      val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
      txids += confirmedTxs.map { it.txid }
      claimed += confirmedTxs.map { it.txOut.map { it.amount }.sum() }.sum()
   }
   listOfNotNull(
      remoteCommitPublished,
      nextRemoteCommitPublished,
      futureRemoteCommitPublished
   ).forEach {
      val confirmedTxids = it.irrevocablySpent.values.toSet()
      val allTxs = listOfNotNull(it.commitTx, it.claimMainOutputTx) + it.claimHtlcSuccessTxs + it.claimHtlcTimeoutTxs
      val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
      txids += confirmedTxs.map { it.txid }
      claimed += confirmedTxs.map { it.txOut.map { it.amount }.sum() }.sum()
   }
   revokedCommitPublished.forEach {
      val confirmedTxids = it.irrevocablySpent.values.toSet()
      val allTxs = listOfNotNull(it.commitTx, it.claimMainOutputTx, it.mainPenaltyTx) + it.htlcPenaltyTxs + it.claimHtlcDelayedPenaltyTxs
      val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
      txids += confirmedTxs.map { it.txid }
      claimed += confirmedTxs.map { it.txOut.map { it.amount }.sum() }.sum()
   }
   return ChannelAction.Storage.CompleteChannelClosing(txids, claimed)
}
```

</p></details>

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
@@ -2928,7 +2928,7 @@ data class Closing(
             val confirmedTxids = it.irrevocablySpent.values.toSet()
             val allTxs = listOfNotNull(it.commitTx, it.claimMainOutputTx) + it.claimHtlcSuccessTxs + it.claimHtlcTimeoutTxs
             val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
-            if (confirmedTxs.size > 0) {
+            if (confirmedTxs.isNotEmpty()) {
                 txids += confirmedTxs.map { it.txid }
                 claimed += confirmedTxs.map { it.txOut.map { it.amount }.sum() }.sum()
                 type = ChannelClosingType.Remote
--- a/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
@@ -2938,7 +2938,7 @@ data class Closing(
             val confirmedTxids = it.irrevocablySpent.values.toSet()
             val allTxs = listOfNotNull(it.commitTx, it.claimMainOutputTx, it.mainPenaltyTx) + it.htlcPenaltyTxs + it.claimHtlcDelayedPenaltyTxs
             val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
-            if (confirmedTxs.size > 0) {
+            if (confirmedTxs.isNotEmpty()) {
                 txids += confirmedTxs.map { it.txid }
                 claimed += confirmedTxs.map { it.txOut.map { it.amount }.sum() }.sum()
                 type = ChannelClosingType.Other
--- a/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
@@ -2902,23 +2902,15 @@ data class Closing(
             else -> ChannelClosingType.Other // includes revoked commit published case
         }
         additionalConfirmedTx?.let { confirmedTx ->
-            mutualClosePublished.firstOrNull { it == confirmedTx }?.let {
-                txids += it.txid
-                // NB: this code could be much simpler if we knew the localScriptPubKey
-                var expectedAmount = commitments.localCommit.spec.toLocal.truncateToSatoshi()
-                if (commitments.localParams.isFunder) {
-                    // we had to pay the closingFees
-                    val inputAmount = commitments.commitInput.txOut.amount
-                    val outputAmount = it.txOut.map { it.amount }.sum()
-                    val feesAmount = outputAmount - inputAmount
-                    expectedAmount -= feesAmount
-                }
-                claimed += expectedAmount
+            mutualClosePublished.firstOrNull { it.tx == confirmedTx }?.let {
+                txids += it.tx.txid
+                claimed += it.toLocalOutput?.amount ?: 0.sat
             }
         }
         localCommitPublished?.let {
-            val confirmedTxids = it.irrevocablySpent.values.toSet()
-            val allTxs = listOfNotNull(it.commitTx, it.claimMainDelayedOutputTx) + it.htlcSuccessTxs + it.htlcTimeoutTxs
+            val confirmedTxids = it.irrevocablySpent.values.map { it.txid }.toSet()
+            val allTxs = listOfNotNull(it.claimMainDelayedOutputTx?.tx) +
+                    it.claimHtlcDelayedTxs.map { it.tx }
             val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
             if (confirmedTxs.isNotEmpty()) {
                 txids += confirmedTxs.map { it.txid }
@@ -2930,8 +2922,9 @@ data class Closing(
             nextRemoteCommitPublished,
             futureRemoteCommitPublished
         ).forEach {
-            val confirmedTxids = it.irrevocablySpent.values.toSet()
-            val allTxs = listOfNotNull(it.commitTx, it.claimMainOutputTx) + it.claimHtlcSuccessTxs + it.claimHtlcTimeoutTxs
+            val confirmedTxids = it.irrevocablySpent.values.map { it.txid }.toSet()
+            val allTxs = listOfNotNull(it.claimMainOutputTx?.tx) +
+                    it.claimHtlcTxs.mapNotNull { it.value?.tx }
             val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
             if (confirmedTxs.isNotEmpty()) {
                 txids += confirmedTxs.map { it.txid }
@@ -2939,8 +2932,10 @@ data class Closing(
             }
         }
         revokedCommitPublished.forEach {
-            val confirmedTxids = it.irrevocablySpent.values.toSet()
-            val allTxs = listOfNotNull(it.commitTx, it.claimMainOutputTx, it.mainPenaltyTx) + it.htlcPenaltyTxs + it.claimHtlcDelayedPenaltyTxs
+            val confirmedTxids = it.irrevocablySpent.values.map { it.txid }.toSet()
+            val allTxs = listOfNotNull(it.claimMainOutputTx?.tx, it.mainPenaltyTx?.tx) +
+                    it.htlcPenaltyTxs.map { it.tx } +
+                    it.claimHtlcDelayedPenaltyTxs.map { it.tx }
             val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
             if (confirmedTxs.isNotEmpty()) {
                 txids += confirmedTxs.map { it.txid }
--- a/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
@@ -759,10 +759,10 @@ data class Offline(val state: ChannelStateWithCommitments) : ChannelState() {
                                 currentTip,
                                 currentOnChainFeerates,
                                 state.commitments,
-                                null,
-                                currentBlockHeight.toLong(),
-                                state.closingTxProposed.flatten().map { it.unsignedTx },
-                                listOf(closingTx)
+                                fundingTx = null,
+                                waitingSinceBlock = currentBlockHeight.toLong(),
+                                mutualCloseProposed = state.closingTxProposed.flatten().map { it.unsignedTx },
+                                mutualClosePublished = listOf(closingTx)
                             )
                             val actions = listOf(
                                 ChannelAction.Storage.StoreState(nextState),
@@ -1030,10 +1030,10 @@ data class Syncing(val state: ChannelStateWithCommitments, val waitForTheirReest
                                 currentTip,
                                 currentOnChainFeerates,
                                 state.commitments,
-                                null,
-                                currentBlockHeight.toLong(),
-                                state.closingTxProposed.flatten().map { it.unsignedTx },
-                                listOf(closingTx)
+                                fundingTx = null,
+                                waitingSinceBlock = currentBlockHeight.toLong(),
+                                mutualCloseProposed = state.closingTxProposed.flatten().map { it.unsignedTx },
+                                mutualClosePublished = listOf(closingTx)
                             )
                             val actions = listOf(
                                 ChannelAction.Storage.StoreState(nextState),
@@ -2367,10 +2367,10 @@ data class Negotiating(
                                 currentTip,
                                 currentOnChainFeerates,
                                 commitments,
-                                null,
-                                currentBlockHeight.toLong(),
-                                this.closingTxProposed.flatten().map { it.unsignedTx },
-                                listOf(signedClosingTx)
+                                fundingTx = null,
+                                waitingSinceBlock = currentBlockHeight.toLong(),
+                                mutualCloseProposed = this.closingTxProposed.flatten().map { it.unsignedTx },
+                                mutualClosePublished = listOf(signedClosingTx)
                             )
                             val actions = listOf(
                                 ChannelAction.Storage.StoreState(nextState),
@@ -2388,10 +2388,10 @@ data class Negotiating(
                                 currentTip,
                                 currentOnChainFeerates,
                                 commitments,
-                                null,
-                                currentBlockHeight.toLong(),
-                                this.closingTxProposed.flatten().map { it.unsignedTx } + listOf(signedClosingTx),
-                                listOf(signedClosingTx)
+                                fundingTx = null,
+                                waitingSinceBlock = currentBlockHeight.toLong(),
+                                mutualCloseProposed = this.closingTxProposed.flatten().map { it.unsignedTx } + listOf(signedClosingTx),
+                                mutualClosePublished = listOf(signedClosingTx)
                             )
                             val actions = listOf(
                                 ChannelAction.Storage.StoreState(nextState),
@@ -2435,10 +2435,10 @@ data class Negotiating(
                             currentTip,
                             currentOnChainFeerates,
                             commitments,
-                            null,
-                            currentBlockHeight.toLong(),
-                            this.closingTxProposed.flatten().map { it.unsignedTx },
-                            listOf(closingTx)
+                            fundingTx = null,
+                            waitingSinceBlock = currentBlockHeight.toLong(),
+                            mutualCloseProposed = this.closingTxProposed.flatten().map { it.unsignedTx },
+                            mutualClosePublished = listOf(closingTx)
                         )
                         val actions = listOf(
                             ChannelAction.Storage.StoreState(nextState),
@@ -2482,10 +2482,10 @@ data class Negotiating(
                     currentTip,
                     currentOnChainFeerates,
                     commitments,
-                    null,
-                    currentBlockHeight.toLong(),
-                    this.closingTxProposed.flatten().map { it.unsignedTx } + listOf(bestUnpublishedClosingTx),
-                    listOf(bestUnpublishedClosingTx)
+                    fundingTx = null,
+                    waitingSinceBlock = currentBlockHeight.toLong(),
+                    mutualCloseProposed = this.closingTxProposed.flatten().map { it.unsignedTx } + listOf(bestUnpublishedClosingTx),
+                    mutualClosePublished = listOf(bestUnpublishedClosingTx)
                 )
                 val actions = listOf(
                     ChannelAction.Storage.StoreState(nextState),
--- a/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
@@ -2899,7 +2899,8 @@ data class Closing(
             remoteCommitPublished != null -> ChannelClosingType.Remote
             nextRemoteCommitPublished != null -> ChannelClosingType.Remote
             futureRemoteCommitPublished != null -> ChannelClosingType.Remote
-            else -> ChannelClosingType.Other // includes revoked commit published case
+            revokedCommitPublished.isNotEmpty() -> ChannelClosingType.Revoked
+            else -> ChannelClosingType.Other
         }
         additionalConfirmedTx?.let { confirmedTx ->
             mutualClosePublished.firstOrNull { it.tx == confirmedTx }?.let {
--- a/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
@@ -2914,7 +2914,7 @@ data class Closing(
             val confirmedTxids = it.irrevocablySpent.values.toSet()
             val allTxs = listOfNotNull(it.commitTx, it.claimMainDelayedOutputTx) + it.htlcSuccessTxs + it.htlcTimeoutTxs
             val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
-            if (confirmedTxs.size > 0) {
+            if (confirmedTxs.isNotEmpty()) {
                 txids += confirmedTxs.map { it.txid }
                 claimed += confirmedTxs.map { it.txOut.map { it.amount }.sum() }.sum()
                 type = ChannelClosingType.Local
--- a/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
@@ -96,7 +96,7 @@ sealed class ChannelAction {
         data class GetHtlcInfos(val revokedCommitTxId: ByteVector32, val commitmentNumber: Long) : Storage()
         data class StoreIncomingAmount(val amount: MilliSatoshi, val origin: ChannelOrigin?) : Storage()
         data class StoreChannelClosing(val amount: MilliSatoshi, val closingAddress: String, val isLocalWallet: Boolean) : Storage()
-        data class CompleteChannelClosing(val txids: List<ByteVector32>, val claimed: Satoshi) : Storage()
+        data class StoreChannelClosed(val txids: List<ByteVector32>, val claimed: Satoshi) : Storage()
     }
 
     data class ProcessIncomingHtlc(val add: UpdateAddHtlc) : ChannelAction()
@@ -2678,7 +2678,7 @@ data class Closing(
                                 if (closingType !is MutualClose) {
                                     logger.debug { "c:$channelId last known remoteChannelData=${commitments.remoteChannelData}" }
                                 }
-                                Pair(Closed(closing1), listOf(closing1.completeChannelClosing(watch.tx)))
+                                Pair(Closed(closing1), listOf(closing1.storeChannelClosed(watch.tx)))
                             }
                         }
                         val actions = buildList {
@@ -2888,7 +2888,7 @@ data class Closing(
         }
     }
 
-    private fun completeChannelClosing(additionalConfirmedTx: Transaction?): ChannelAction.Storage.CompleteChannelClosing {
+    private fun storeChannelClosed(additionalConfirmedTx: Transaction?): ChannelAction.Storage.StoreChannelClosed {
         // We want to give the user the list of btc transactions for their outputs
         val txids = mutableListOf<ByteVector32>()
         var claimed = 0.sat
@@ -2932,7 +2932,7 @@ data class Closing(
             txids += confirmedTxs.map { it.txid }
             claimed += confirmedTxs.map { it.txOut.map { it.amount }.sum() }.sum()
         }
-        return ChannelAction.Storage.CompleteChannelClosing(txids, claimed)
+        return ChannelAction.Storage.StoreChannelClosed(txids, claimed)
     }
 }
 
--- a/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
@@ -2893,7 +2893,14 @@ data class Closing(
         // We want to give the user the list of btc transactions for their outputs
         val txids = mutableListOf<ByteVector32>()
         var claimed = 0.sat
-        var type = ChannelClosingType.Other
+        val type = when {
+            mutualClosePublished.isNotEmpty() -> ChannelClosingType.Mutual
+            localCommitPublished != null -> ChannelClosingType.Local
+            remoteCommitPublished != null -> ChannelClosingType.Remote
+            nextRemoteCommitPublished != null -> ChannelClosingType.Remote
+            futureRemoteCommitPublished != null -> ChannelClosingType.Remote
+            else -> ChannelClosingType.Other // includes revoked commit published case
+        }
         additionalConfirmedTx?.let { confirmedTx ->
             mutualClosePublished.firstOrNull { it == confirmedTx }?.let {
                 txids += it.txid
@@ -2907,7 +2914,6 @@ data class Closing(
                     expectedAmount -= feesAmount
                 }
                 claimed += expectedAmount
-                type = ChannelClosingType.Mutual
             }
         }
         localCommitPublished?.let {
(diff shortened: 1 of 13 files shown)


src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
```diff
@@ -2819,6 +2861,64 @@ data class Closing(
 
         return txs[tx.txid]?.let { (_, desc) -> fee(tx)?.let { it to desc } }
     }
+
+    private fun storeChannelClosed(additionalConfirmedTx: Transaction?): ChannelAction.Storage.StoreChannelClosed {
+        // We want to give the user the list of btc transactions for their outputs
+        val txids = mutableListOf<ByteVector32>()
+        var claimed = 0.sat
+        var type = ChannelClosingType.Other
```
dpad85, line 2869: The closing type depends on the channel's xxxCommitPublish types, but it's not very clear what types takes precedence.

How about something like:

```kotlin
val type = when {
    mutualClosePublished.isNotEmpty() -> ChannelClosingType.Mutual
    localCommitPublished != null -> ChannelClosingType.Local
    remoteCommitPublished != null -> ChannelClosingType.Remote
    else -> ChannelClosingType.Other // includes revoked commit published case
}
```

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
```diff
@@ -163,7 +166,41 @@ sealed class ChannelState {
...
+                        if (localShutdown != null && localShutdown.scriptPubKey != defaultScriptPubKey) {
+                            // Non-default output address
+                            val btcAddr = Helpers.Closing.btcAddressFromScriptPubKey(
+                                scriptPubKey = localShutdown.scriptPubKey,
+                                chainHash = staticParams.nodeParams.chainHash
+                            ) ?: "unknown"
+                            actions + ChannelAction.Storage.StoreChannelClosing(
+                                amount = channelBalance,
+                                closingAddress = btcAddr,
+                                isLocalWallet = false
```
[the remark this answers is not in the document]
robbiehanson: Done in 986acc7

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
robbiehanson: Done in 3f65583

I think that if `nextRemoteCommitPublished` is non-null, then the type would also be `ChannelClosingType.Remote`. However I'm not 100% sure about `futureRemoteCommitPublished`. Please check.

dpad85: Yes, 3f65583 is correct to me

src/commonMain/kotlin/fr/acinq/eclair/db/InMemoryPaymentsDb.kt
```diff
@@ -51,22 +51,25 @@ class InMemoryPaymentsDb : PaymentsDb {
         return outgoing[id]?.let { payment ->
             val parts = outgoingParts.values.filter { it.first == payment.id }.map { it.second }
             return when (payment.status) {
-                is OutgoingPayment.Status.Succeeded -> payment.copy(parts = parts.filter { it.status is OutgoingPayment.Part.Status.Succeeded })
+                // This doesn't work. Is there something like "instanceof" to check inheritance ???
+            //  is OutgoingPayment.Status.Completed.Succeeded -> {
+            //      payment.copy(parts = parts.filter { it.status is OutgoingPayment.Part.Status.Succeeded })
+            //  }
+                is OutgoingPayment.Status.Completed.Succeeded.OffChain -> {
```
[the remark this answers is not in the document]
robbiehanson: I thought it should work too. But it doesn't ...

As the code is now, it will pass all unit tests. If you change the code to what it was before, it will fail multiple unit tests:
```kotlin
return when (payment.status) {
    // This doesn't work ...
    is OutgoingPayment.Status.Completed.Succeeded -> {
        payment.copy(parts = parts.filter { it.status is OutgoingPayment.Part.Status.Succeeded })
    }
    else -> payment.copy(parts = parts)
}
```

For example, this test will fail:
OutgoingPaymentHandlerTestsCommon: 689: partial failure then fulfill (spec violation)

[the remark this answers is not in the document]
robbiehanson: Done in cc9e868 :hand_over_mouth:

robbiehanson:
@t-bast Can you check the changes I made in df2be1f. It incorporates the changes from your recent PR, but I'm not sure my logic is correct.

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
```diff
@@ -182,17 +201,44 @@ data class OutgoingPayment(val id: UUID, val recipientAmount: MilliSatoshi, val
         /** Swaps out send a lightning payment to a swap server, which will send an on-chain transaction to a given address. */
         data class SwapOut(val address: String, override val paymentHash: ByteVector32) : Details()
 
+        /** Corresponds to the on-chain payments made when closing a channel. */
+        data class ChannelClosing(
+            val channelId: ByteVector32,
+            val closingAddress: String,
+            val isSentToMyWallet: Boolean,
```
t-bast, line 208: It's quite unclear what `isSentToMyWallet` means, it should be documented.

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
```diff
@@ -182,17 +201,44 @@ data class OutgoingPayment(val id: UUID, val recipientAmount: MilliSatoshi, val
...
+                    override val completedAt: Long = currentTimestampMillis()
+                ) : Succeeded()
+                data class OnChain(
+                    val txids: List<ByteVector32>,
+                    val claimed: Satoshi,
+                    val type: ChannelClosingType,
+                    override val completedAt: Long = currentTimestampMillis()
+                ) : Succeeded() {
+                    enum class ChannelClosingType {
+                        Mutual, Local, Remote, Other
```
t-bast, line 237: Why don't you support `Revoked` here? It looks like it would be trivial to do, wouldn't it?

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
```diff
@@ -182,17 +201,44 @@ data class OutgoingPayment(val id: UUID, val recipientAmount: MilliSatoshi, val
...
+            abstract val completedAt: Long
+            data class Failed(val reason: FinalFailure, override val completedAt: Long = currentTimestampMillis()) : Completed()
+            sealed class Succeeded : Completed() {
+                data class OffChain(
+                    val preimage: ByteVector32,
+                    override val completedAt: Long = currentTimestampMillis()
+                ) : Succeeded()
+                data class OnChain(
+                    val txids: List<ByteVector32>,
+                    val claimed: Satoshi,
```
t-bast, line 232: You should document this field.
It looks like for this `OnChain` case, the fee we could display to the user would actually be `recipientAmount - claimed`, is that correct?

I still think the type architecture is incorrect here though. I think it should be:

```kotlin
data class OnChain(val type: ChannelClosingType, override val completedAt: Long = currentTimestampMillis()) : Succeeded()
```

And each claimed tx should be a `Part`.
It does require a bit more work (to make `Part` more abstract), but I'm afraid we'll have to pay that toll later anyway *and* a data migration...

src/commonMain/kotlin/fr/acinq/eclair/io/Peer.kt
```diff
@@ -405,6 +407,30 @@ class Peer(
                     val htlcInfos = db.channels.listHtlcInfos(actualChannelId, action.commitmentNumber).map { ChannelAction.Storage.HtlcInfo(actualChannelId, action.commitmentNumber, it.first, it.second) }
                     input.send(WrappedChannelEvent(actualChannelId, ChannelEvent.GetHtlcInfosResponse(action.revokedCommitTxId, htlcInfos)))
                 }
+                action is ChannelAction.Storage.StoreChannelClosing -> {
+                    val dbId = UUID.fromBytes(channelId.take(16).toByteArray())
+                    val payment = OutgoingPayment(
+                        id = dbId,
+                        recipientAmount = action.amount,
+                        recipient = PublicKey.Generator,
```
t-bast, line 415: The recipient should be us, shouldn't it? So `nodeParams.nodeId`?

src/commonMain/kotlin/fr/acinq/eclair/io/Peer.kt
```diff
@@ -405,6 +407,30 @@ class Peer(
...
+                        details = OutgoingPayment.Details.ChannelClosing(
+                            channelId = channelId,
+                            closingAddress = action.closingAddress,
+                            isSentToMyWallet = action.isSentToMyWallet,
+                            paymentHash = randomBytes32()
```
t-bast, line 420: We could do like we did for swaps, `channelId.sha256()` so that it's deterministic

src/commonTest/kotlin/fr/acinq/eclair/channel/states/ClosingTestsCommon.kt
```diff
@@ -1724,7 +1739,13 @@ class ClosingTestsCommon : EclairTestSuite() {
                 val (alice1, actions1) = alice.process(ChannelEvent.ExecuteCommand(CMD_FORCECLOSE))
                 assertTrue(alice1 is Closing)
                 assertEquals(7, actions1.size)
-                assertTrue(actions1.contains(ChannelAction.Storage.StoreOutgoingAmount(alice1.commitments.localCommit.spec.toLocal)))
+
+                val channelBalance = alice1.commitments.localCommit.spec.toLocal
+                if (channelBalance > 0.msat) {
+                    val ledgerPayment = actions1.filterIsInstance<ChannelAction.Storage.StoreChannelClosing>().firstOrNull()
```
t-bast, line 1745: We don't use `ledger`, maybe `onChainPayment` is more consistent? Same in other tests where it's used.

t-bast (commented):
We didn't need #229 that much in fact since we don't really need to check individual transaction fees (we can do an approximate aggregate by just comparing how much we had off-chain to how much we claimed on-chain).

I think the PR is missing tests though.

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
```diff
@@ -164,7 +167,41 @@ sealed class ChannelState {
                 this is WaitForInit && newState is Closing -> actions
                 this is Closing && newState is Closing -> actions
                 this is ChannelStateWithCommitments && newState is Closing -> {
-                    actions + ChannelAction.Storage.StoreOutgoingAmount(this.commitments.localCommit.spec.toLocal)
+                    val channelBalance = commitments.localCommit.spec.toLocal
```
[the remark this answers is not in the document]
t-bast: Note that it can quickly become tricky though if we want to do it *perfectly*, but we can use a simple approximation:

- we can consider that all incoming HTLCs will be claimed by us (this ignores only the case where people send us payments for invoices that aren't ours, which shouldn't happen often)
- we can consider that all outgoing HTLCs will work and correctly reach the recipient who will fulfill them, so we ignore them in our balance calculation

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
```diff
@@ -164,7 +167,41 @@ sealed class ChannelState {
...
+                        if (localShutdown != null && localShutdown.scriptPubKey != defaultScriptPubKey) {
+                            // Non-default output address
+                            val btcAddr = Helpers.Closing.btcAddressFromScriptPubKey(
+                                scriptPubKey = localShutdown.scriptPubKey,
+                                chainHash = staticParams.nodeParams.chainHash
+                            ) ?: "unknown"
+                            actions + ChannelAction.Storage.StoreChannelClosing(
+                                amount = channelBalance,
+                                closingAddress = btcAddr,
+                                isSentToMyWallet = false
```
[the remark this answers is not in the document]
dpad85: `isSentToMyWallet` indicates that the funds are sent to the default Phoenix address derived from the wallet seed. It lets the UI provide additional info and explain that the funds are still theirs, and how they can recover them. 

Typically this would be useful when the closing is a force close and the user may be confused, and would need help.

This field is also useful if the closing is mutual and Phoenix did not provide an address.

How about renaming it to `isSentToMyAddress` ?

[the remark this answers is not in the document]
t-bast: Gotcha, what about `isSentToDefaultAddress` then?

src/commonMain/kotlin/fr/acinq/eclair/channel/Helpers.kt
```diff
@@ -363,6 +363,46 @@ object Helpers {
 
         fun isValidFinalScriptPubkey(scriptPubKey: ByteVector): Boolean = isValidFinalScriptPubkey(scriptPubKey.toByteArray())
 
+        fun btcAddressFromScriptPubKey(scriptPubKey: ByteVector, chainHash: ByteVector32): String? {
```
sstone, line 366: nit: I would just call this `btcAddressFromScriptPubKey`
We need unit tests for this, I'll create some from Bitcoin Core's reference tests. 
This conversion should also be provided by `bitcoin-kmp`, I'll add it there.

src/commonMain/kotlin/fr/acinq/eclair/channel/Helpers.kt
```diff
@@ -363,6 +363,46 @@ object Helpers {
...
+            return runTrying {
+                val script = Script.parse(scriptPubKey)
+                when {
+                    Script.isPay2pkh(script) -> {
+                        // OP_DUP OP_HASH160 OP_PUSHDATA(20) OP_EQUALVERIFY OP_CHECKSIG
+                        val opPushData = script[2] as OP_PUSHDATA
+                        val prefix = when (chainHash) {
+                            Block.LivenetGenesisBlock.hash -> Base58.Prefix.PubkeyAddress
+                            Block.TestnetGenesisBlock.hash -> Base58.Prefix.PubkeyAddressTestnet
+                            else -> Base58.Prefix.PubkeyAddressSegnet
```
sstone, line 376: `segnet` is not used anymore, regtest addresses also use the  `PubkeyAddressTestnet` prefix, and we should explicitly match against `Block.RegtestGenesisBlock`

src/commonMain/kotlin/fr/acinq/eclair/channel/Helpers.kt
```diff
@@ -363,6 +363,46 @@ object Helpers {
...
+                        }
+                        Base58Check.encode(prefix, opPushData.data)
+                    }
+                    Script.isPay2sh(script) -> {
+                        // OP_HASH160 OP_PUSHDATA(20) OP_EQUAL
+                        val opPushData = script[1] as OP_PUSHDATA
+                        val prefix = when (chainHash) {
+                            Block.LivenetGenesisBlock.hash -> Base58.Prefix.ScriptAddress
+                            Block.TestnetGenesisBlock.hash -> Base58.Prefix.ScriptAddressTestnet
+                            else -> Base58.Prefix.ScriptAddressSegnet
```
sstone, line 386: same as above

src/commonMain/kotlin/fr/acinq/eclair/channel/Helpers.kt
```diff
@@ -363,6 +363,46 @@ object Helpers {
...
+                        Base58Check.encode(prefix, opPushData.data)
+                    }
+                    Script.isPay2wpkh(script) || Script.isPay2wsh(script) -> {
+                        // isPay2wpkh : OP_0 OP_PUSHDATA(20)
+                        // isPay2wsh  : OP_0 OP_PUSHDATA(32)
+                        val opPushData = script[1] as OP_PUSHDATA
+                        val hrp = when (chainHash) {
+                            Block.LivenetGenesisBlock.hash -> "bc"
+                            Block.TestnetGenesisBlock.hash -> "tb"
+                            else -> "bcrt"
```
sstone, line 397: We should also explicitly match against `Block.RegtestGenesisBlock` here

src/commonMain/kotlin/fr/acinq/eclair/channel/Helpers.kt
sstone: You can use test data from https://github.com/ACINQ/bitcoin-kmp/blob/3b64f865daf3abc34208ce428600f8d7256ae851/src/commonTest/kotlin/fr/acinq/bitcoin/BitcoinTestsCommon.kt

robbiehanson: This function was ported from phoenix-kmm. I forgot to port the unit tests, but they're [here](https://github.com/ACINQ/phoenix-kmm/blob/fix/issue-123/phoenix-shared/src/commonTest/kotlin/fr/acinq/phoenix/app/UtilitiesTest.kt).

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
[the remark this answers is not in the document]
robbiehanson: Done in 443e204

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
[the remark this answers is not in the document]
robbiehanson: > it can quickly become tricky though if we want to do it _perfectly_

The goal is to "balance the ledger" from the user's point-of-view. The user had a balance of X in the channel. And then the channel was closed. So the transaction should have an amount of -X

I used `commitments.localCommit.spec.toLocal` for the simple reason that, it's what we use in phoenix-kmm to report the user's balance (per channel). [link](https://github.com/ACINQ/phoenix-kmm/blob/d555efe981cd37149578bf2091d5fedf21d7bb98/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/app/ctrl/AppHomeController.kt#L48)

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
```diff
@@ -2631,19 +2672,20 @@ data class Closing(
                             commitments.payments[add.id]?.let { paymentId -> logger.info { "c:$channelId paymentId=$paymentId will settle on-chain (htlc #${add.id} sending ${add.amountMsat})" } }
                         }
 
-                        val nextState = when (val closingType = closing1.isClosed(watch.tx)) {
-                            null -> closing1
+                        val (nextState, closedActions) = when (val closingType = closing1.isClosed(watch.tx)) {
+                            null -> Pair(closing1, listOf<ChannelAction>())
```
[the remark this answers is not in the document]
robbiehanson: > IntelliJ should be showing a warning

It's not :frowning: IntelliJ has been horrible lately. It can't properly type-check the iOS specific files. It always struggles with SQLDelight files. And for several weeks type-checking of Gradle files was completely broken. And then it couldn't find `kolinx` for several weeks, which resulted in red highlights everywhere.

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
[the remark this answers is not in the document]
t-bast: You're right, it's probably good enough to only look at the main output, either way could be confusing in some scenarios so let's keep the simplest one.

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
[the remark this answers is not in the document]
t-bast: Damn that must be frustrating...good luck :sweat_smile:

src/commonMain/kotlin/fr/acinq/eclair/channel/Channel.kt
```diff
@@ -2846,6 +2888,62 @@ data class Closing(
...
+            }
+        }
+        localCommitPublished?.let {
+            val confirmedTxids = it.irrevocablySpent.values.map { it.txid }.toSet()
+            val allTxs = listOfNotNull(it.claimMainDelayedOutputTx?.tx) +
+                    it.claimHtlcDelayedTxs.map { it.tx }
+            val confirmedTxs = allTxs.filter { confirmedTxids.contains(it.txid) }
+            if (confirmedTxs.isNotEmpty()) {
+                txids += confirmedTxs.map { it.txid }
+                claimed += confirmedTxs.map { it.txOut.map { it.amount }.sum() }.sum()
```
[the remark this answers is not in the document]
robbiehanson: Done in 35b217f

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
```diff
@@ -157,14 +160,30 @@ data class IncomingPayment(val preimage: ByteVector32, val origin: Origin, val r
...
+                        0.msat
+                    } else {
+                        sum - recipientAmount
+                    }
+                }
+            }
+
+        } else {
+            0.msat
+        }
```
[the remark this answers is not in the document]
robbiehanson: Done in f63b2ef

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
t-bast: Actually the data migration shouldn't be too bad, so feel free to ignore this comment for now for the sake of shipping sooner ;)

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
robbiehanson: Done in b181149

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
robbiehanson: Indeed. Done in 202dfc2

src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
robbiehanson: Done in 551cf9f

> I think it should be: each claimed tx should be a `Part`.

Agree 100%

> we'll have to pay that toll later anyway and a data migration...

I believe the data migration will actually be easy.

`Part` will probably become a sealed class:
```kotlin
sealed class Part {
    data class LightningPart(/* ... */) : Part()
    data class ClosingTransaction(/* ... */): Part()
}
```

But on-disk, the `ClosingTransaction` it will have it's own separate database table.

The structure of `OnChain` will change. But luckily we've already planned for this, and the database architecture is now designed to handle these upgrades:

```kotlin
// The database table has these columns:
// - status_type INTEGER DEFAULT 0
// - status BLOB DEFAULT NULL
//
// The `status_type` column is populated with a value from this enum.
//
enum class OutgoingStatusType(val type: Long) {
    // Versioning: Each type supports up to 100 versions.
    // For example, Succeeded_v1(type = 101)
    //
    Succeeded_v0(100),
    Failed_v0(200),
    Mined_v0(300),
    Mined_v1(301); // <= we just add a new version

    fun version(): Long = type % 100L
}
```

If there are existing `Mined_v0` blobs in the database, we can upgrade them to the new structure on-the-fly. In particular, we will create a fake `Part.ClosingTransaction` with values from the old `OnChain` JSON.

t-bast: Good, let's wait to do that later then!

Do note though that with the current code, we may end up showing users a negative fee in the following scenario:

- The channel has a big pending outgoing htlc (to be extreme, all the funds that were on the user side)
- We manage to get the htlc funds back via an htlc-timeout tx on-chain
- `recipientAmount` was `0` because it didn't take this htlc into account
- `claimed` will be `N sats` (the htlc amount)

To correctly address that we'll need to do what I mentioned in another comment: when going to `Closed`, we should check the final state, update `recipientAmount` and calculate accurate fees (but let's do it later).

src/commonMain/kotlin/fr/acinq/eclair/io/Peer.kt
robbiehanson: Yes (if actions.isSentToDefaultAddress)

Done in d105365

src/commonMain/kotlin/fr/acinq/eclair/io/Peer.kt
robbiehanson: Done in d105365

src/commonTest/kotlin/fr/acinq/eclair/channel/states/ClosingTestsCommon.kt
robbiehanson: Done in b155d3d

src/commonMain/kotlin/fr/acinq/eclair/channel/Helpers.kt
robbiehanson: Added to bitcoin-kmp: https://github.com/ACINQ/bitcoin-kmp/pull/17

robbiehanson: I believe we need to merge PR #231 so we can get access to the new function in bitcoin-kmp ?

robbiehanson: Until we're ready to use the latest bitcoin-kmp from this project, I've updated the existing function, and copied the unit tests.

Done in cdeb6db

src/commonMain/kotlin/fr/acinq/eclair/channel/Helpers.kt
robbiehanson: Done in cdeb6db

src/commonMain/kotlin/fr/acinq/eclair/channel/Helpers.kt
robbiehanson: Done in cdeb6db

src/commonMain/kotlin/fr/acinq/eclair/channel/Helpers.kt
robbiehanson: Done in cdeb6db

src/commonTest/kotlin/fr/acinq/eclair/channel/states/ClosingTestsCommon.kt
```diff
@@ -1724,7 +1762,13 @@ class ClosingTestsCommon : EclairTestSuite() {
                 val (alice1, actions1) = alice.process(ChannelEvent.ExecuteCommand(CMD_FORCECLOSE))
                 assertTrue(alice1 is Closing)
                 assertEquals(7, actions1.size)
-                assertTrue(actions1.contains(ChannelAction.Storage.StoreOutgoingAmount(alice1.commitments.localCommit.spec.toLocal)))
+
+                val channelBalance = alice1.commitments.localCommit.spec.toLocal
+                if (channelBalance > 0.msat) {
+                    val onChainPayment = actions1.filterIsInstance<ChannelAction.Storage.StoreChannelClosing>().firstOrNull()
+                    assertNotNull(onChainPayment)
+                    assertTrue(onChainPayment.amount == channelBalance)
```
t-bast, line 1770: You should also validate the other fields of `StoreChannelClosing`.
There should be one test where we send to a non-default address as well.
And there should be a test validating the contents of `StoreChannelClosed` too.

robbiehanson: Done in 9353876

t-bast (approved):

dpad85 (approved):

PR #236, opened 2021-03-23
Handle aggregated anchor outputs htlc txs

t-bast:
An interesting side-effect of anchor outputs is that htlc txs can be merged when they have the same lockTime (thanks to sighash flags).

We're not currently doing that, but our peers may do it, so we need to handle it in the revoked commit tx case and correctly claim multiple outputs if necessary.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Channel.kt
@@ -2586,8 +2586,8 @@ data class Closing(
 
                         val revokedCommitPublishActions = mutableListOf<ChannelAction>()
                         val revokedCommitPublished1 = revokedCommitPublished.map { rev ->
-                            val (newRevokedCommitPublished, tx) = Helpers.Closing.claimRevokedHtlcTxOutputs(keyManager, commitments, rev, watch.tx, currentOnChainFeerates)
-                            tx?.let {
+                            val (newRevokedCommitPublished, penaltyTxs) = Helpers.Closing.claimRevokedHtlcTxOutputs(keyManager, commitments, rev, watch.tx, currentOnChainFeerates)
+                            penaltyTxs.forEach {
                                 revokedCommitPublishActions += ChannelAction.Blockchain.PublishTx(it.tx)
                                 revokedCommitPublishActions += ChannelAction.Blockchain.SendWatch(WatchSpent(channelId, watch.tx, it.input.outPoint.index.toInt(), BITCOIN_OUTPUT_SPENT))
                             }
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
@@ -795,6 +795,9 @@ object Helpers {
          * - by spending the delayed output of [[HtlcSuccessTx]] and [[HtlcTimeoutTx]] if those get confirmed; because the output of these txs is protected by
          * an OP_CSV delay, we will have time to spend them with a revocation key. In that case, we generate the spending transactions "on demand",
          * this is the purpose of this method.
+         *
+         * NB: when anchor outputs is used, htlc transactions can be aggregated in a single transaction if they share the same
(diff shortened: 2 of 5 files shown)


src/commonMain/kotlin/fr/acinq/eclair/transactions/Transactions.kt
```diff
@@ -791,17 +790,28 @@ object Transactions {
         return TransactionWithInputInfo.ClosingTx(commitTxInput, tx, toLocalOutput)
     }
 
+    private fun matchPubKeyScript(txOut: TxOut, pubkeyScript: ByteArray, amount_opt: Satoshi?): Boolean {
```
sstone, line 793: What is the use case for  non-null `amount_opt`  ?

t-bast: It looks like a remnant from previous code, it's not used anywhere with a non-null value (not even in `eclair`).
I can remove it (it's quite brittle to rely on the amount to match anyway).

src/commonTest/kotlin/fr/acinq/eclair/channel/states/ClosingTestsCommon.kt
```diff
@@ -1413,6 +1414,65 @@ class ClosingTestsCommon : EclairTestSuite() {
...
+            ChannelAction.Storage.HtlcInfo(alice0.channelId, 4, htlcsBob[1].paymentHash, htlcsBob[1].cltvExpiry),
+        )
+        val (alice2, _) = alice1.processEx(ChannelEvent.GetHtlcInfosResponse(bobRevokedTx.commitTx.tx.txid, htlcInfos))
+        assertTrue(alice2 is Closing)
+
+        // bob claims multiple htlc-timeout in a single transaction (this is possible with anchor outputs because signatures
+        // use sighash_single | sighash_anyonecanpay)
+        val bobHtlcTimeoutTxs = bobRevokedTx.htlcTxsAndSigs.filter { it.txinfo is Transactions.TransactionWithInputInfo.HtlcTx.HtlcTimeoutTx }
+        assertEquals(2, bobHtlcTimeoutTxs.size)
+        val bobHtlcTx = Transaction(
```
sstone, line 1442: Here Bob could also publish a tx that spends both his htlc-success txs ? Shouldn't we test for that too ?

t-bast: Yes he could, I can add a test but I feel it's redundant, we've tested elsewhere that we're able to claim revoked htlc-success, and we have a test that verifies that we scan all outputs (here and in `TransactionTestsCommon`).
Testing the htlc-success case here is a bit painful because we need to fetch the preimage to create valid htlc-success witnesses, it can be done but since it's already tested elsewhere I'm not sure it's worth it.

t-bast: I'll add one, it doesn't hurt to have more tests and not worth being lazy :)

sstone: I saw the new transaction test  in TransactionTestsCommon where all htlc txs are aggregated but missed the one we have here ?

src/commonMain/kotlin/fr/acinq/eclair/transactions/Transactions.kt
t-bast: Done in https://github.com/ACINQ/eclair-kmp/pull/236/commits/09e5ca8b4e489450631920395d40ec0c5561a2dd

src/commonTest/kotlin/fr/acinq/eclair/channel/states/ClosingTestsCommon.kt
t-bast: Done in https://github.com/ACINQ/eclair-kmp/pull/236/commits/915c275886b640c645a51952ae5c28959480f315
It's even simpler in `eclair-kmp` than `eclair`, no need to ensure the transaction is completely valid and we don't do script introspection.

sstone (approved):

sstone (approved):

PR #237, opened 2021-03-24
Add feature bits for experimental wallet features

t-bast:
Wallets should explicitly advertise support for experimental features (e.g. pay-to-open, swaps, wake-up notifications, etc) in their `init` message, to allow the LSPs they connect to to know whether they should use these experimental features (e.g. open a channel on-the-fly when liquidity is missing to receive a payment).

Otherwise the LSP must use heuristics that may be incorrect and initiate experimental protocols with nodes/wallets that don't support them (this is happening with electrum).

I used two feature bits for asymmetric features, as explored here: https://github.com/lightningnetwork/lightning-rfc/issues/885

A later PR should introduce support for actually disabling these features (and turning off the corresponding code branches).

NB: this can be merged after the initial release, no rush.


dpad85:
Seems like a good idea 👍 we could add the swap-in bit now as well

Also why not use a even higher feature bit? 60 does not seem very high to me.

t-bast:
> Also why not use a even higher feature bit? 60 does not seem very high to me.

True, we could. At the rate at which features get adopted in the spec, I think 60 and above still let us enjoy these for at least 10 years, but we can be more prudent :laughing:

dpad85 (dismissed):

sstone (approved):

PR #240, opened 2021-03-29
Make ChannelClosingType serializable

dpad85:
This PR makes the channel closing type enum explicitly serializable. 

This should not be needed. However on Phoenix, this object is serialized and saved to DB. And when building the app on Mac,  kotlinx.serialization raises a compilation error (`AssertionError: Enum class does not have .values() function`). This does not happen on Android. Adding the annotation fixes the kotlinx.serialization compilation issue.

This commit also sanitizes the `ChannelClosingType` hierarchy which was a bit long.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
@@ -11,6 +11,7 @@ import fr.acinq.eclair.payment.FinalFailure
 import fr.acinq.eclair.payment.PaymentRequest
 import fr.acinq.eclair.utils.*
 import fr.acinq.eclair.wire.FailureMessage
+import kotlinx.serialization.Serializable
 
 interface PaymentsDb : IncomingPaymentsDb, OutgoingPaymentsDb {
     /** List sent and received payments (with most recent payments first). */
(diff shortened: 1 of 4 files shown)


src/commonMain/kotlin/fr/acinq/eclair/db/PaymentsDb.kt
```diff
@@ -262,6 +259,11 @@ data class OutgoingPayment(val id: UUID, val recipientAmount: MilliSatoshi, val
     }
 }
 
+@Serializable
+enum class ChannelClosingType {
+    Mutual, Local, Remote, Revoked, Other;
```
t-bast, line 264: nit: what does the trailing `;` do?

dpad85: It just marks the end of the enum values, it is not needed in our case since the compiler is smart enough to know that from the `}` closing the enum. But if we were to add a method below the enum values, we would need a trailing `;` as a separator

t-bast (approved):

dpad85:
Closing this pull request after discussion with @sstone offline. Phoenix needs should not leak into eclair-kmp. If Phoenix needs to serialize some types it should be handled in Phoenix.

PR #242, opened 2021-03-30
Do not override amount in db for incoming amount of pay-to-open origin

dpad85:
This PR changes how the `StoreIncomingAmount` action is handled when the origin is `ChannelOrigin.PayToOpenOrigin`. The action's amount is now ignored, and only the payment's received-with field is updated.

### The issue

For pay-to-open (a channel is created but the origin is a LN invoice), a payment line has already been added and received in the database. That part is done by the `IncomingPaymentHandler` main processing method : https://github.com/ACINQ/eclair-kmp/blob/7d29237a54258cd67b1a32bcaf9599d6b66931ec/src/commonMain/kotlin/fr/acinq/eclair/payment/IncomingPaymentHandler.kt#L212-L219

However, the `db.receiveIncomingPayment` method will **ADD** the given amount to the already received amount for this payment.

The reason why is that we may receive several payments for a given payment hash, and as such `db.receiveIncomingPayment` must be additive. For reference, see this issue on phoenix: https://github.com/ACINQ/phoenix-kmm/issues/100

So saving the `StoreIncomingAmount` action amount causes the amount to be incorrect (actually doubled) in the database.

### Proposed solution

The proposed solution is to save a 0.msat amount for `StoreIncomingAmount` where origin is `PayToOpenOrigin`.

This PR also add some tests and fixes the `InMemoryDb.receiveIncomingPayment` method.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/eclair/payment/IncomingPaymentHandler.kt
+++ b/src/commonMain/kotlin/fr/acinq/eclair/payment/IncomingPaymentHandler.kt
@@ -94,6 +94,54 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
     }
 
     /**
+     * Save the "received-with" details of an incoming amount.
+     *
+     * - for a pay-to-open origin, we only save the id of the channel that was created for this payment.
+     * - for a swap-in origin, a new incoming payment must be created. We use the channel id to generate the payment's preimage.
+     * - for unknown origin, the amount is handled as a swap-in coming from an unknown address.
+     */
+    suspend fun process(channelId: ByteVector32, action: ChannelAction.Storage.StoreIncomingAmount) {
+        when (action.origin) {
+            null -> {
+                logger.warning { "incoming amount with empty origin, we store only minimal information" }
+                val fakePreimage = channelId.sha256()
+                db.addAndReceivePayment(
+                    preimage = fakePreimage,
+                    origin = IncomingPayment.Origin.SwapIn(address = ""),
+                    amount = action.amount,
+                    receivedWith = IncomingPayment.ReceivedWith.NewChannel(fees = 0.msat, channelId = channelId)
+                )
+            }
+            is ChannelOrigin.PayToOpenOrigin -> {
+                if (db.getIncomingPayment(action.origin.paymentHash) != null) {
+                    db.receivePayment(
+                        paymentHash = action.origin.paymentHash,
+                        amount = 0.msat, // do not update the amount, it's already set by the main payment handler.
+                        receivedWith = IncomingPayment.ReceivedWith.NewChannel(
+                            fees = action.origin.fee.toMilliSatoshi(),
+                            channelId = channelId
+                        )
+                    )
+                } else {
+                    logger.warning { "c:$channelId ignored pay-to-open origin action, there are no payments in db for payment_hash=${action.origin.paymentHash}" }
+                }
+            }
+            is ChannelOrigin.SwapInOrigin -> {
+                val fakePreimage = channelId.sha256()
+                db.addAndReceivePayment(
+                    preimage = fakePreimage,
+                    origin = IncomingPayment.Origin.SwapIn(address = action.origin.bitcoinAddress),
+                    amount = action.amount,
+                    receivedWith = IncomingPayment.ReceivedWith.NewChannel(
+                        fees = action.origin.fee.toMilliSatoshi(),
+                        channelId = channelId
+                    )
+                )
+            }
+        }
+    }
+
+    /**
      * Process an incoming htlc.
      * Before calling this, the htlc must be committed and ack-ed by both sides.
      *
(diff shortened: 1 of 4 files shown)


src/commonMain/kotlin/fr/acinq/eclair/payment/IncomingPaymentHandler.kt
```diff
@@ -93,6 +93,54 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
...
+     * Save the "received-with" details of an incoming amount.
+     *
+     * - for a pay-to-open origin, we only save the id of the channel that was created for this payment.
+     * - for a swap-in origin, a new incoming payment must be created. We use the channel id to generate the payment's preimage.
+     * - for unknown origin, the amount is handled as a swap-in coming from an unknown address.
+     */
+    suspend fun process(channelId: ByteVector32, action: ChannelAction.Storage.StoreIncomingAmount) {
+        when (action.origin) {
+            null -> {
+                logger.warning { "incoming amount with empty origin, we store only minimal information" }
```
pm47, line 106: It's unrelated to this PR, but there should be a comment explaining what the scenario is here. Why use a `SwapIn` origin with an empty addres?

src/commonMain/kotlin/fr/acinq/eclair/payment/IncomingPaymentHandler.kt
```diff
@@ -93,6 +93,54 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
...
+                    db.receivePayment(
+                        paymentHash = action.origin.paymentHash,
+                        amount = 0.msat, // do not update the amount, it's already set by the main payment handler.
+                        receivedWith = IncomingPayment.ReceivedWith.NewChannel(
+                            fees = action.origin.fee.toMilliSatoshi(),
+                            channelId = channelId
+                        )
+                    )
+                } else {
+                    logger.warning { "c:$channelId ignored pay-to-open origin action, there are no payments in db for payment_hash=${action.origin.paymentHash}" }
```
pm47, line 126: It seems that the node/channel context is handled somewhere else, so maybe we can get rid of it here?
```suggestion
                    logger.warning { "ignored pay-to-open origin action, there are no payments in db for payment_hash=${action.origin.paymentHash}" }
```

src/commonMain/kotlin/fr/acinq/eclair/payment/IncomingPaymentHandler.kt
```diff
@@ -93,6 +93,54 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
...
+                }
+            }
+            is ChannelOrigin.SwapInOrigin -> {
+                val fakePreimage = channelId.sha256()
```
pm47, line 130:
```suggestion
                // swap-ins are push payments made with an on-chain tx, there is no related preimage so we make up one so it fits in our model
                val fakePreimage = channelId.sha256()
```

pm47 (commented):
Looks fine to me, just nits.

Have you considered the opposite? That is: storing the amount in the `ChannelAction` and not in the main payment handler? The rationale would be that whatever is in the `ChannelAction` is the _actual_ amount received by the user. With the change you proposed, if the invoice is overpaid, we will update the db, but will not open a new channel, so there will be a discrepancy. But it doesn't change anything for the user, and this is a corner case, so I'm okay with both.

dpad85:
> Have you considered the opposite? That is: storing the amount in the ChannelAction and not in the main payment handler?

I did, but it would bring additional complexity/special cases to the main payment handler, something that we may want to avoid.

> With the change you proposed, if the invoice is overpaid, we will update the db, but will not open a new channel, so there will be a discrepancy.

What discrepancy do you have in mind? Regarding the amount, if the invoice is overpaid, whatever the solution we use, the amount in the database for this payment will always be greater than the amount initially pushed to the channel when it was opened.

src/commonMain/kotlin/fr/acinq/eclair/payment/IncomingPaymentHandler.kt
dpad85: I use a `SwapIn` origin because in that case we are basically receiving an unsolicited amount. We could use a `KeySend` origin but since we don't really support keysend yet, it does not seems right.

pm47: In the keysend scenario there is a preimage (it is provided by the sender), but here we are generating it. Shouldn't we remove that case altogether? We can deal with that later of course and just add this:
```suggestion
                // TODO: hacky, needs clean-up
                logger.warning { "incoming amount with empty origin, we store only minimal information" }
```

pm47:
> I'm not sure anymore whether db.receivePayment should be additive or should instead override the amount received.

When in doubt, let's fall back to the same behavior as eclair, which is additive.

src/commonMain/kotlin/fr/acinq/eclair/payment/IncomingPaymentHandler.kt
dpad85: So we would ignore this action and not store anything, right?

pm47: I guess so but I haven't looked into it in detail. Let's deal with that in a separate PR? Doesn't seem very urgent?

pm47 (approved):

PR #244, opened 2021-03-31
Release v1.0-beta8

dpad85:


robbiehanson (approved):
Compiles fine on my machine. And does create "1.0-beta8" directories in ~/.m2

PR #246, opened 2021-04-01
Sanitize logs

t-bast:
Don't print seed. Ever. Just don't.
No need to print Ping/Pong, it's too noisy in the logs.


dpad85 (approved):

dpad85:
Test ok 👍

PR #252, opened 2021-04-07
Rename package eclair to lightning

sstone:


dpad85 (approved):
Test on android/ios

Not backward compatible because it breaks channels serialization (package name has changed and is used for serialization).

PR #255, opened 2021-04-20
Precompute channel keys

sstone:
Key manager will compute all channel keys and will pass generated private keys back to the channel where they will be kept and used. We use the same scheme as before and this is not a breaking change, we just return private keys as well as public keys now.
This is much more efficient than re-computing them from the channel path every time as we do now.
Another option would be to cache private keys in the key manager (as we do on eclair) but it's not as useful for mobile apps as we are not likely to delegate signing to an external module.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
@@ -459,13 +459,12 @@ data class Commitments(
 
         // remote commitment will include all local changes + remote acked changes
         val spec = CommitmentSpec.reduce(remoteCommit.spec, remoteChanges.acked, localChanges.proposed)
-        val (remoteCommitTx, htlcTxs) = makeRemoteTxs(keyManager, channelVersion, remoteCommit.index + 1, localParams, remoteParams, commitInput, remoteNextPerCommitmentPoint, spec)
-        val sig = keyManager.sign(remoteCommitTx, keyManager.fundingPublicKey(localParams.fundingKeyPath))
+        val (remoteCommitTx, htlcTxs) = makeRemoteTxs( remoteCommit.index + 1, localParams, remoteParams, commitInput, remoteNextPerCommitmentPoint, spec)
+        val sig = keyManager.sign(remoteCommitTx, localParams.channelKeys.fundingPrivateKey)
 
         val sortedHtlcTxs: List<HtlcTx> = htlcTxs.sortedBy { it.input.outPoint.index }
-        val channelKeyPath = keyManager.channelKeyPath(localParams, channelVersion)
         // we sign our peer's HTLC txs with SIGHASH_SINGLE || SIGHASH_ANYONECANPAY
-        val htlcSigs = sortedHtlcTxs.map { keyManager.sign(it, keyManager.htlcPoint(channelKeyPath), remoteNextPerCommitmentPoint, SigHash.SIGHASH_SINGLE or SigHash.SIGHASH_ANYONECANPAY) }
+        val htlcSigs = sortedHtlcTxs.map { keyManager.sign(it, localParams.channelKeys.htlcKey, remoteNextPerCommitmentPoint, SigHash.SIGHASH_SINGLE or SigHash.SIGHASH_ANYONECANPAY) }
 
         // NB: IN/OUT htlcs are inverted because this is the remote commit
         log.info {
@@ -504,10 +503,9 @@ data class Commitments(
         // receiving money i.e its commit tx has one output for them
 
         val spec = CommitmentSpec.reduce(localCommit.spec, localChanges.acked, remoteChanges.proposed)
-        val channelKeyPath = keyManager.channelKeyPath(localParams, channelVersion)
-        val localPerCommitmentPoint = keyManager.commitmentPoint(channelKeyPath, localCommit.index + 1)
-        val (localCommitTx, htlcTxs) = makeLocalTxs(keyManager, channelVersion, localCommit.index + 1, localParams, remoteParams, commitInput, localPerCommitmentPoint, spec)
-        val sig = keyManager.sign(localCommitTx, keyManager.fundingPublicKey(localParams.fundingKeyPath))
+        val localPerCommitmentPoint = keyManager.commitmentPoint(localParams.channelKeys.shaSeed, localCommit.index + 1)
+        val (localCommitTx, htlcTxs) = makeLocalTxs( localCommit.index + 1, localParams, remoteParams, commitInput, localPerCommitmentPoint, spec)
+        val sig = Transactions.sign(localCommitTx, localParams.channelKeys.fundingPrivateKey)
 
         log.info {
             val htlcsIn = spec.htlcs.incomings().map { it.id }.joinToString(",")
@@ -516,7 +514,7 @@ data class Commitments(
         }
 
         // no need to compute htlc sigs if commit sig doesn't check out
-        val signedCommitTx = Transactions.addSigs(localCommitTx, keyManager.fundingPublicKey(localParams.fundingKeyPath).publicKey, remoteParams.fundingPubKey, sig, commit.signature)
+        val signedCommitTx = Transactions.addSigs(localCommitTx, localParams.channelKeys.fundingPubKey, remoteParams.fundingPubKey, sig, commit.signature)
         when (val check = Transactions.checkSpendable(signedCommitTx)) {
             is Try.Failure -> {
                 log.error(check.error) { "c:$channelId remote signature $commit is invalid" }
@@ -528,7 +526,7 @@ data class Commitments(
         if (commit.htlcSignatures.size != sortedHtlcTxs.size) {
             return Either.Left(HtlcSigCountMismatch(channelId, sortedHtlcTxs.size, commit.htlcSignatures.size))
         }
-        val htlcSigs = sortedHtlcTxs.map { keyManager.sign(it, keyManager.htlcPoint(channelKeyPath), localPerCommitmentPoint, SigHash.SIGHASH_ALL) }
+        val htlcSigs = sortedHtlcTxs.map { keyManager.sign(it, localParams.channelKeys.htlcKey, localPerCommitmentPoint, SigHash.SIGHASH_ALL) }
         val remoteHtlcPubkey = Generators.derivePubKey(remoteParams.htlcBasepoint, localPerCommitmentPoint)
         // combine the sigs to make signed txs
         val htlcTxsAndSigs = Triple(sortedHtlcTxs, htlcSigs, commit.htlcSignatures).zipped().map { (htlcTx, localSig, remoteSig) ->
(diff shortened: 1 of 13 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
```diff
@@ -339,12 +338,11 @@ object Helpers {
...
-            val channelKeyPath = keyManager.channelKeyPath(localParams, channelVersion)
-            val commitmentInput = makeFundingInputInfo(fundingTxHash, fundingTxOutputIndex, fundingAmount, fundingPubKey.publicKey, remoteParams.fundingPubKey)
-            val localPerCommitmentPoint = keyManager.commitmentPoint(channelKeyPath, 0)
-            val localCommitTx = Commitments.makeLocalTxs(keyManager, channelVersion, 0, localParams, remoteParams, commitmentInput, localPerCommitmentPoint, localSpec).first
-            val remoteCommitTx = Commitments.makeRemoteTxs(keyManager, channelVersion, 0, localParams, remoteParams, commitmentInput, remoteFirstPerCommitmentPoint, remoteSpec).first
+            val fundingPubKey = localParams.channelKeys.fundingPubKey
+            val commitmentInput = makeFundingInputInfo(fundingTxHash, fundingTxOutputIndex, fundingAmount, fundingPubKey, remoteParams.fundingPubKey)
+            val localPerCommitmentPoint = keyManager.commitmentPoint(localParams.channelKeys.shaSeed, 0)
+            val localCommitTx = Commitments.makeLocalTxs( 0, localParams, remoteParams, commitmentInput, localPerCommitmentPoint, localSpec).first
```
t-bast, line 344: nit: reformat this file (this unnecessary space is in multiple places):
```suggestion
            val localCommitTx = Commitments.makeLocalTxs(0, localParams, remoteParams, commitmentInput, localPerCommitmentPoint, localSpec).first
```

src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
```diff
@@ -29,15 +30,21 @@ interface KeyManager {
...
             // deterministic mode: use the funding pubkey to compute the channel key path
-            channelKeyPath(fundingPublicKey(localParams.fundingKeyPath))
+            channelKeyPath(fundingPublicKey(fundingKeyPath))
         } else {
             // legacy mode:  we reuse the funding key path as our channel key path
-            localParams.fundingKeyPath
+            fundingKeyPath
         }
 
+    fun channelKeyPath(localParams: LocalParams, channelVersion: ChannelVersion): KeyPath = channelKeyPath(localParams.channelKeys.fundingKeyPath, channelVersion)
```
t-bast, line 46: Do we want to keep that function exposed?
I think we can make the interface of the `KeyManager` is bit smaller to avoid mis-using it.

src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
```diff
@@ -47,14 +54,16 @@ interface KeyManager {
      */
     fun newFundingKeyPath(isFunder: Boolean): KeyPath
 
+    fun channelKeys(fundingKeyPath: KeyPath) : ChannelKeys
+
     /**
      *
      * @param tx        input transaction
      * @param publicKey extended public key
```
t-bast, line 62: The documentation for all these functions should be updated.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v1/ChannelState.kt
```diff
@@ -169,7 +169,7 @@ data class LocalParams constructor(
 ) {
     constructor(from: fr.acinq.lightning.channel.LocalParams) : this(
         from.nodeId,
-        from.fundingKeyPath,
+        from.channelKeys.fundingKeyPath,
```
t-bast, line 172: So we keep serializing with the old format and re-compute the channel keys when we load.
The main reason is to avoid storing private keys directly in the channels DB, right?
It may be worth a comment to ensure we don't "fix this" by mistake in the future.

src/commonTest/kotlin/fr/acinq/lightning/crypto/LocalKeyManagerTestsCommon.kt
```diff
@@ -36,6 +37,29 @@ class LocalKeyManagerTestsCommon : LightningTestSuite() {
...
+    @Test
+    fun `generate channel keys`() {
+        val seed = ByteVector("aeb3e9b5642cd4523e9e09164047f60adb413633549c3c6189192921311894d501")
+        val keyManager = LocalKeyManager(seed, Block.TestnetGenesisBlock.hash)
+        val fundingKeyPath = makefundingKeyPath(ByteVector("06535806c1aa73971ec4877a5e2e684fa636136c073810f190b63eefc58ca488"), isFunder = false)
+        val channelKeys = keyManager.channelKeys(fundingKeyPath)
+
+        // README !
+        // test data generated with v1.0-beta11, but they should never change
+        // if this test fails it means that we cannot restore channels created with older phoenix of lightning-kmp without
```
t-bast, line 49:
```suggestion
        // if this test fails it means that we cannot restore channels created with older versions of lightning-kmp without
```

src/commonTest/kotlin/fr/acinq/lightning/transactions/AnchorOutputsTestsCommon.kt
```diff
@@ -90,7 +88,8 @@ class AnchorOutputsTestsCommon {
     private fun runHighLevelTest(testCase: TestCase) {
         val localParams = LocalParams(
             TestConstants.Alice.nodeParams.nodeId,
-            KeyPath.empty, 546.sat, 1000000000L, 0.sat, 0.msat, CltvExpiryDelta(144), 1000, true,
+            ChannelKeys( KeyPath.empty, local_funding_privkey, local_payment_basepoint_secret, local_delayed_payment_basepoint_secret, local_payment_basepoint_secret, local_payment_basepoint_secret, randomBytes32()),
```
t-bast, line 91:
```suggestion
            ChannelKeys(KeyPath.empty, local_funding_privkey, local_payment_basepoint_secret, local_delayed_payment_basepoint_secret, local_payment_basepoint_secret, local_payment_basepoint_secret, randomBytes32()),
```

t-bast (commented):
Nice change!

src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
sstone: I'd like to keep it for now, it feels a bit cleaner to use `localParams`

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -459,13 +459,12 @@ data class Commitments(
 
         // remote commitment will include all local changes + remote acked changes
         val spec = CommitmentSpec.reduce(remoteCommit.spec, remoteChanges.acked, localChanges.proposed)
-        val (remoteCommitTx, htlcTxs) = makeRemoteTxs(keyManager, channelVersion, remoteCommit.index + 1, localParams, remoteParams, commitInput, remoteNextPerCommitmentPoint, spec)
-        val sig = keyManager.sign(remoteCommitTx, keyManager.fundingPublicKey(localParams.fundingKeyPath))
+        val (remoteCommitTx, htlcTxs) = makeRemoteTxs( remoteCommit.index + 1, localParams, remoteParams, commitInput, remoteNextPerCommitmentPoint, spec)
```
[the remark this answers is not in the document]
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/255/commits/3b01af57bbb8bc95af625ce1d11ae54e433214de

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -504,10 +503,9 @@ data class Commitments(
         // receiving money i.e its commit tx has one output for them
 
         val spec = CommitmentSpec.reduce(localCommit.spec, localChanges.acked, remoteChanges.proposed)
-        val channelKeyPath = keyManager.channelKeyPath(localParams, channelVersion)
-        val localPerCommitmentPoint = keyManager.commitmentPoint(channelKeyPath, localCommit.index + 1)
-        val (localCommitTx, htlcTxs) = makeLocalTxs(keyManager, channelVersion, localCommit.index + 1, localParams, remoteParams, commitInput, localPerCommitmentPoint, spec)
-        val sig = keyManager.sign(localCommitTx, keyManager.fundingPublicKey(localParams.fundingKeyPath))
+        val localPerCommitmentPoint = keyManager.commitmentPoint(localParams.channelKeys.shaSeed, localCommit.index + 1)
+        val (localCommitTx, htlcTxs) = makeLocalTxs( localCommit.index + 1, localParams, remoteParams, commitInput, localPerCommitmentPoint, spec)
```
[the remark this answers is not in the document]
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/255/commits/3b01af57bbb8bc95af625ce1d11ae54e433214de

src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
sstone: yes I upgraded/downgraded idea many times and lost the "save actions" configurations :(
done in https://github.com/ACINQ/lightning-kmp/pull/255/commits/3b01af57bbb8bc95af625ce1d11ae54e433214de

src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/255/commits/3b01af57bbb8bc95af625ce1d11ae54e433214de

src/commonMain/kotlin/fr/acinq/lightning/serialization/v1/ChannelState.kt
sstone: Exactly, I added an explanation.

src/commonTest/kotlin/fr/acinq/lightning/crypto/LocalKeyManagerTestsCommon.kt
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/255/commits/3b01af57bbb8bc95af625ce1d11ae54e433214de

src/commonTest/kotlin/fr/acinq/lightning/transactions/AnchorOutputsTestsCommon.kt
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/255/commits/3b01af57bbb8bc95af625ce1d11ae54e433214de

t-bast (approved):

PR #256, opened 2021-04-21
Use current timestamp in invoice preimage

t-bast:
In case our random number generator completely fails, we may end up generating the same payment preimage for unrelated payments.

To mitigate that, we use the current time at the end of the preimage.
This still leaves at least 192 bits of randomness which is more than enough.

NB: wallets who want to leverage this (e.g. Phoenix) will need to call this new function instead of `randomBytes32()` when generating an invoice.


t-bast:
Closing in favor of #258

We don't want our preimages to look different from other wallet's preimages, because it would give intermediate node information about the recipient.

PR #257, opened 2021-04-21
Don't pay already paid invoices

### LINKED ISSUE #238: Disallow paying an invoice multiple times
We should not pay an invoice if we already know the preimage.
We won't be able to prove we paid, and the recipient may not receive our payment if an intermediate node remembers the preimage and fulfills without relaying.

t-bast:
If we already have the preimage for a given invoice, we should not attempt to pay it again. We will introduce static invoices in the future to allow paying multiple times the same recipient, but it shouldn't be done with standard invoices.

Fixes #238


dpad85 (approved):
LGTM

This check will have to be made in phoenix-kmm too, right after the invoice has been scanned/pasted. It's better to fail the payment early and not even try to send it.

PR #258, opened 2021-04-21
Add secondary pseudo-random number generator

t-bast:
In case of catastrophic failures of the OS `SecureRandom` instance, we add a secondary randomness source that we mix into the random stream.

This is a somewhat weak random source and should not be used on its own, but it doesn't hurt to xor it with the output of `SecureRandom`.

The `runtimeEntropy` should be customized for iOS and Android by fetching all kind of random noise available to the application (available memory, process ID, sensor data, network statistics, etc).

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
@@ -3,7 +3,7 @@ package fr.acinq.lightning.crypto
 import fr.acinq.bitcoin.*
 import fr.acinq.bitcoin.DeterministicWallet.derivePrivateKey
 import fr.acinq.bitcoin.DeterministicWallet.hardened
-import fr.acinq.lightning.Lightning.secureRandom
+import fr.acinq.lightning.Lightning.randomLong
 import fr.acinq.lightning.transactions.Transactions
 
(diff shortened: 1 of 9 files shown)


PhoenixCrypto/PhoenixCrypto/Classes/NativeWeakRandom.swift
```diff
@@ -0,0 +1,70 @@
...
+@objc
+public class NativeWeakRandom: NSObject {
+
+    @objc
+    private class func toByteArr(i: UInt64) -> [UInt8] {
+        let count = MemoryLayout<UInt64>.size
+        var _i = i
+        let bytePtr = withUnsafePointer(to: &_i) {
+            $0.withMemoryRebound(to: UInt8.self, capacity: count) {
+                UnsafeBufferPointer(start: $0, count: count)
```
robbiehanson, line 19: I'd feel better if the `UnsafeBufferPointer` didn't escape the `withUnsafePointer` lambda:

```swift
return withUnsafePointer(to: &_i) {
  $0.withMemoryRebound(to: UInt8.self, capacity: count) {
    [UInt8](UnsafeBufferPointer(start: $0, count: count))
  }
}
```

robbiehanson (dismissed):

robbiehanson (dismissed):

PR #261, opened 2021-04-28
More flexible mutual close fees

t-bast:
Add support for https://github.com/lightningnetwork/lightning-rfc#847

With legacy nodes, we will keep the existing behavior (slowly converge on a fee).
But for nodes that support closing `fee_range`s, mutual close will take much less round-trips.
This has also been implemented in eclair: https://github.com/ACINQ/eclair/pull/1768

**Open question: should we drop support for the old negotiation and greatly simplify that code?**

NB: this builds on #277 to leverage the new `ChannelState` version, and should be included in the same release.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Channel.kt
@@ -1007,7 +1007,7 @@ data class Syncing(val state: ChannelStateWithCommitments, val waitForTheirReest
                             state.commitments,
                             state.localShutdown.scriptPubKey.toByteArray(),
                             state.remoteShutdown.scriptPubKey.toByteArray(),
-                            currentOnChainFeerates.mutualCloseFeerate
+                            state.closingFeerates ?: ClosingFeerates(currentOnChainFeerates.mutualCloseFeerate)
                         )
                         val closingTxProposed1 = state.closingTxProposed + listOf(listOf(ClosingTxProposed(closingTx, closingSigned)))
                         val nextState = state.copy(closingTxProposed = closingTxProposed1)
@@ -1778,6 +1778,7 @@ data class WaitForFundingLocked(
                         initialChannelUpdate,
                         null,
                         null,
+                        null,
                         null
                     )
                     val actions = listOf(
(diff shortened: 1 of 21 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/Channel.kt
```diff
@@ -2335,82 +2342,107 @@ data class Negotiating(
...
-                            val (closingTx, closingSigned) = Helpers.Closing.makeClosingTx(keyManager, commitments, localShutdown.scriptPubKey.toByteArray(), remoteShutdown.scriptPubKey.toByteArray(), nextClosingFee)
-                            logger.info { "c:$channelId proposing closingFeeSatoshis=${closingSigned.feeSatoshis}" }
-                            val closingProposed1 = closingTxProposed.updated(
-                                closingTxProposed.lastIndex,
-                                closingTxProposed.last() + listOf(ClosingTxProposed(closingTx, closingSigned))
-                            )
-                            val nextState = this.copy(
-                                commitments = commitments.copy(remoteChannelData = event.message.channelData),
-                                closingTxProposed = closingProposed1,
-                                bestUnpublishedClosingTx = closingTx
```
t-bast, line 2403: NB: there was a bug here: we must use `signedClosingTx` and not `closingTx`, otherwise we're forgetting our peer's signature :face_with_head_bandage:

src/commonMain/kotlin/fr/acinq/lightning/serialization/v1/ChannelState.kt
```diff
@@ -695,7 +695,8 @@ data class Normal(
         channelUpdate,
         remoteChannelUpdate,
         localShutdown,
-        remoteShutdown
+        remoteShutdown,
+        null
```
t-bast, line 699: Do you think we should introduce a new version to persist those?
I had to use `process` instead of `processEx` in some tests because the closing feerates are currently ignored when storing the state.

t-bast: Note to self: we should probably use #277 that introduces a new version to also include this new field (as it will probably be merged after this PR).

sstone (approved):

PR #267, opened 2021-05-19
Make payment secret mandatory

t-bast:
This is a security feature that has been introduced a long time ago and is widely supported across the network.
We can safely make it mandatory which closes probing attack vectors.


sstone (approved):
Tested with eclair-mobile.

PR #268, opened 2021-05-26
Check max "safe" number of pending htlcs

sstone:
We want to make sure that LN messages that include encrypted peer backups are still < 65 Kb, which means that we also need to enforce a max.value on the number of outgoing payments and not just incoming ones.
We test that we can safely have 6 incoming and 6 outgoing payments.


src/commonMain/kotlin/fr/acinq/lightning/channel/ChannelException.kt
```diff
@@ -57,6 +57,7 @@ data class HtlcValueTooSmall                       (override val channelId: Byte
 @OptIn(ExperimentalUnsignedTypes::class)
 data class HtlcValueTooHighInFlight                (override val channelId: ByteVector32, val maximum: ULong, val actual: MilliSatoshi) : ChannelException(channelId, "in-flight htlcs hold too much value: maximum=$maximum actual=$actual")
 data class TooManyAcceptedHtlcs                    (override val channelId: ByteVector32, val maximum: Long) : ChannelException(channelId, "too many accepted htlcs: maximum=$maximum")
+data class TooManyOffereddHtlcs                    (override val channelId: ByteVector32, val maximum: Long) : ChannelException(channelId, "too many offered htlcs: maximum=$maximum")
```
t-bast, line 60: nit:
```suggestion
data class TooManyOfferedHtlcs                    (override val channelId: ByteVector32, val maximum: Long) : ChannelException(channelId, "too many offered htlcs: maximum=$maximum")
```

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -270,6 +270,11 @@ data class Commitments(
             return Either.Left(TooManyAcceptedHtlcs(channelId, maximum = commitments1.remoteParams.maxAcceptedHtlcs.toLong()))
         }
 
+        // README: this is not part of the LN Bolts: we also check against our own limit, to avoid creating commit txs that have too many outputs
+        if (outgoingHtlcs.size > commitments1.localParams.maxAcceptedHtlcs) {
```
t-bast, line 274: That means the maximum number of HTLC we'll have in the commit tx is `2 * localParams.maxAcceptedHtlcs` (since we have at most `localParams.maxAcceptedHtlcs` incoming HTLCs + at most `localParams.maxAcceptedHtlcs` outgoing HTLCs), is that ok?

sstone: Yes. But maybe it would be cleaner to add a specific `max-offered-htlcs` parameter to`NodeParams`, I'll check how it impacts serialisation and peer backups.

t-bast: To be honest too many configuration hooks can become overkill...
I would rather make it clear that the total number of HTLCs in the commit tx can be twice `max-accepted-htlcs` and use just that for now.

sstone: Ok I've changed my mind and I think using `max-accepted-htlcs` for both received and offered payments makes more sense :)

t-bast: I think we also need to apply `max-htlc-value-in-flight` symmetrically, don't you think?

sstone: Yes it makes sense

t-bast (approved):
LGTM, that means we need to configure Phoenix to set `max-accepted-htlcs` to `6`.

PR #269, opened 2021-06-09
Watch tower support

robbiehanson:
This PR adds support for a simple client-side "Watch Tower" task, that can be run on either iOS or Android. For example, iOS may schedule a background operation to run every few days.

The idea is that the client will run a task that:

- connects to the electrum server (_doesn't need to connect to lightning server_)
- receives status of watched transactions
- reacts appropriately if commits have been published

This is all well and good. It just means we connect to the electrum server, and let the existing code do its thing. And we already have unit tests for these code paths.

But, of course, Apple puts restrictions on background tasks. So on iOS, a background task has a limited amount of time to execute. And it needs to explicitly tell the OS when it has finished.

Which creates a new problem: **How do we know when the task has finished ?**

Answering this question is the focus of this PR.

After investigating a few ideas, I realized there was a simple approach we could use to tell us when we've finished syncing with the electrum server:

- after we've connected to electrum server (post handshake)
- when there are no outstanding requests (excluding pings)
- and after a brief delay to ensure we've processed all incoming information

After that point we've finished syncing with the electrum server, and we consider ourselves up-to-date.

The implementation is composed of 2 primary changes:

### ElectrumClient

Previously, we were storing the outgoing requests via:

```kotlin
val requests: MutableMap<Int, ElectrumRequest>
```

So I've added another field to store the timestamps of the responses:

```kotlin
val responseTimestamps: MutableMap<Int, Long?>
```

If the value is null, then we know we have an outstanding request. If non-null, we know how long ago we received the response.

(Note that some requests, such as subscriptions, may receive multiple responses. So this always stores the timestamp of the most recent response.)

### ElectrumWatcher

After connecting (i.e. post handshake), we start a timer that fires every 2 seconds. When the timer fires, we check to see if:

- every request has received a response from the server
- the last response was at least a few seconds ago

Once the criteria is satisfied, we fire a `NotifyUpToDateEvent`, which the client can listen for.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
@@ -557,6 +557,7 @@ class ElectrumWatcher(
                 is GetScriptHashHistory -> true
                 is GetTransaction -> true
                 is GetMerkle -> true
+                is BroadcastTransaction -> true
                 else -> false
             }
         }.all {
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
@@ -12,13 +12,11 @@ import fr.acinq.lightning.blockchain.electrum.ElectrumWatcher.Companion.makeDumm
 import fr.acinq.lightning.blockchain.electrum.ElectrumWatcher.Companion.registerToScriptHash
 import fr.acinq.lightning.transactions.Scripts
 import fr.acinq.lightning.utils.Connection
+import fr.acinq.lightning.utils.currentTimestampMillis
 import fr.acinq.lightning.utils.lightningLogger
 import kotlinx.coroutines.*
-import kotlinx.coroutines.channels.BroadcastChannel
-import kotlinx.coroutines.channels.Channel
-import kotlinx.coroutines.channels.consumeEach
-import kotlinx.coroutines.channels.produce
-import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.channels.*
+import kotlinx.coroutines.flow.*
 import kotlin.math.absoluteValue
 import kotlin.math.max
 import kotlin.native.concurrent.ThreadLocal
@@ -30,6 +28,11 @@ class ReceiveWatch(val watch: Watch) : WatcherEvent()
 class ReceiveWatchEvent(val watchEvent: WatchEvent) : WatcherEvent()
 class ReceivedMessage(val message: ElectrumMessage) : WatcherEvent()
 class ClientStateUpdate(val connection: Connection) : WatcherEvent()
+
+sealed class NotifyEvent
+class NotifyWatchEvent(val watchEvent: WatchEvent): NotifyEvent()
+class NotifyTxEvent(val channelId: ByteVector32, val txWithMeta: GetTxWithMetaResponse): NotifyEvent()
+class NotifyUpToDateEvent(val millis: Long): NotifyEvent()
 
 internal sealed class WatcherAction
 private object AskForHeaderUpdate : WatcherAction()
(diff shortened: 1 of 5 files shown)


robbiehanson:
>  Do we really need to know when the task has finished?

In short, yes. Apple intelligently schedules background tasks according to several factors. One of which is the "budget", which is the amount of time that your app uses when running its background task(s). The less time you use, the more likely the system will run your background task in the future. Using more time means we drain our budgets, and the OS is more likely to delay running our background task, or simply not do it at all.



This was discussed in a WWDC 2020 video: [Background execution demystified](https://developer.apple.com/videos/play/wwdc2020/10063)



Of course, Apple doesn't tell us what our budgets are. But the video mentions that our budget is dependent upon how often the user uses the app. This presents a problem for us. Because if the user is using Phoenix often (for example, every few days), then we don't even need the Watch Tower. It's when the user is using the app infrequently that our Watch Tower task becomes important. And that also means we can safely assume our budget is lowest when our background task is most critical.



(_Note that one other budget is how much data we download when in the background. So it's helpful that we only connect to electrum, and not the peer._)



> Instead, [...] after 20 seconds [...] (we can save in some database that the watch tower job did not run properly this time).

This implies that we know whether or not the Watch Tower job finished. But we don't currently have that information. The easiest way to determine this yes/no value is:

- after we've connected to electrum server (post handshake)
- when there are no outstanding requests (excluding pings)
- and after a brief delay to ensure we've processed all incoming information



Which is what this PR provides :)

As long as we need this information, we should also give iOS what it needs. Which is some kind of completion notification.

dpad85:
Thanks for the explanation, I was confused. So in this design, the client job only starts the node in the background (without connecting to the peer), node that will in turn publish the closing tx if need be. But the client does not interact with electrum directly, yet it needs to know when the work is done, in order to optimize its "background budget" within iOS.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
```diff
@@ -429,16 +494,85 @@ class ElectrumWatcher(val client: ElectrumClient, val scope: CoroutineScope) : C
...
+        // We want to emit an "up-to-date" notification to the client.
+        // This is used during WatchTower operation, where the client connects to electrum
+        // in order to ensure the acinq server didn't publish an old commit (i.e. attempt to cheat).
+        //
+        // So we want to emit this message when we're fairly confident the electrum server
+        // has delivered all the pending information we need. Here's how we do it:
+        // - we track the last response time (timestamp in millis) per request
+        // - we filter the list to only include those operations related to the WatchTower
+        // - we wait until all such responses have arrived, and are older than 5 seconds
+
```
dpad85, line 552: Why are we waiting 5 seconds? Is it to let the node publish the closing transaction in case it is needed?

What happens if electrum detects an issue but we don't have time to retaliate? Will the user be notified anyway, or do we hope that the node will have time to close the channels next time the job runs?

robbiehanson: Good question. I meant to ask about this.



The ElectrumClient marks the request and completed as soon as the JSON response arrives. But then the response needs to be processed by the channel. Which may result in several actions being queued. Such as storing a new state in the database, and then broadcasting a couple transactions.

I assume that the timer can fire while another coroutine is busy. Such as saving the state to the database. So I'm also assuming that some kind of delay needs to be added here.

But I'm not 100% certain about this. And I'm also not sure what a good wait time should be. Probably 5 seconds is too long.

robbiehanson: > What happens if electrum detects an issue but we don't have time to retaliate?

I think we can break it down into 2 scenarios:
1. `NotifyUpToDate` fires
    In this scenario, the node reacted appropriately. So either nothing needed to be done, or a commit was discovered and one or more transactions were broadcast. We schedule another background task to execute in a few days.
2. `NotifyUpToDate` does NOT fire
    Probably because the connection was slow, or electrum server was slow. Or because the OS decided to terminate the background task for other reasons. (Low battery mode?) We schedule another background task to execute in a few hours.

Either way, there's some kind of cleanup code that runs when the background task is finished. It will know whether the task finished successfully (i.e. via `NotifyUpToDate` fired). It should also be able to detect whether any channels were closed, and why they were closed. So we should be able to switch on this, and display appropriate notifications.

(The logic for this will probably be in phoenix-shared and/or the native layer.)

dpad85 (approved):

PR #271, opened 2021-06-14
Correctly handle incoming htlcs on already paid payments

dpad85:
This pull request fixes an issue where the incoming payment handler would not be able to differentiate a local htlc replay from a new htlc on an already paid invoice. 

Note that local replay can happen when in some edge cases the channel would send to the incoming payment handler an htlc that has already been accepted. In which case the HTLC must be fulfilled, but the incoming amount must not be added to the database again.

The proposed solution is to store additional data in the payments database so that the incoming payment handler can know whether this htlc has already been paid (that must be accepted but not saved again in the db), or is a new, unknown htlc (that must be rejected).

This PR will fix https://github.com/ACINQ/phoenix-kmm/issues/100.

### Changes in `Received` and `ReceivedWith` objects

- Removed `amount` in `Received`. The amount received is now computed by summing the amounts in the payment's `receivedWith` parts
- `receivedWith` is now a set in `Received`
- `ReceivedWith.LightningPayment` now contains the htlc ID and the channel ID

### `IncomingPaymentHandler` changes

When receiving an htlc where the payment is set as paid, then the handler checks if this htlc is already known or not.

### `IncomingPaymentsDb` changes

A new method `updateNewChannelReceivedWithChannelId` has been added to be used when we need to save the channel ID for a pay-to-open payments part. Re-using the `receivePayment` method for that as we used to do was confusing. It's better to have a separate, explicitly named method.


robbiehanson (approved):
Looks good to me :thumbsup:

I think we just need to fix the failing linux tests. (The jvm & iOS64 are passing on my machine)

src/commonTest/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandlerTestsCommon.kt
```diff
@@ -925,37 +933,91 @@ class IncomingPaymentHandlerTestsCommon : LightningTestSuite() {
     }
 
     @Test
-    fun `mpp success then additional HTLC`() = runSuspendTest {
+    fun `mpp success then HTLC local replay`() = runSuspendTest {
```
t-bast, line 936: nit: I would use a more high-level name for this test:
```suggestion
    fun `fulfill locally replayed htlcs`() = runSuspendTest {
```

src/commonTest/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandlerTestsCommon.kt
```diff
@@ -925,37 +933,91 @@ class IncomingPaymentHandlerTestsCommon : LightningTestSuite() {
...
+        run {
+            val result = paymentHandler.process(htlc1, TestConstants.defaultBlockHeight)
+            assertTrue { result is IncomingPaymentHandler.ProcessAddResult.Accepted }
+            val expected = WrappedChannelEvent(channelId1, ChannelEvent.ExecuteCommand(CMD_FULFILL_HTLC(htlc1.id, incomingPayment.preimage, commit = true)))
+            assertEquals(setOf(expected), result.actions.toSet())
+        }
+    }
+
+    @Test
+    fun `mpp success then new HTLC`() = runSuspendTest {
```
t-bast, line 974: nit: same for this one:
```suggestion
    fun `reject htlcs for already paid invoices`() = runSuspendTest {
```

t-bast (commented):
LGTM :rocket: 

Conceptually it makes a lot of sense, we simply record more details about the payment (the list of its parts instead of the aggregated amount). It can be useful in the future as well if we need to record more data about each payment part for some reason.

Just a few nit comments.

t-bast (approved):

Issue #270, opened 2021-06-10
Crash when deserializing JsonRpcError from Electrum


robbiehanson:
I experienced a crash, and luckily I had added logging for the Electrum RPC messages:
```
[ElectrumClient] RECV: {"error":"sendrawtransaction RPC error: {\"code\":-27,\"message\":\"Transaction already in block chain\"}","id":4,"jsonrpc":"2.0"}
```

I think, technically, this is an illegal format for RPC. But if this is being generated by electrum servers all over the world...

The crash:
```
Caused by: kotlinx.serialization.json.internal.JsonDecodingException: Expected class kotlinx.serialization.json.JsonObject as the serialized body of fr.acinq.lightning.utils.JsonRPCError, but had class kotlinx.serialization.json.JsonLiteral
```

robbiehanson:
How I normally run into this problem:
- I close one or more channels
- I come back later (maybe the next day, or several hours later)
- Launch the app, and get a crash

PR #274, opened 2021-06-22
Handle non-standard JSON-RPC errors from electrum

dpad85:
Electrum servers may return a json-rpc error that is not formatted as
per the json-rpc specs (that is, an integer code field + a string
message field), and instead only contains a string, such as:

```json
{
  "jsonrpc": "2.0",
  "id": 4,
  "error": "sendrawtransaction RPC error: {\"code\":123, ...}"
}
```

Instead of throwing a serialization error, we now handle this as a
proper `JsonRPCError` with an id=0.

Fixes https://github.com/ACINQ/lightning-kmp/issues/270


robbiehanson (approved):
The new unit test looks good. And I tested the build on both iOS simulator and device, and everything appears to working normally.

PR #277, opened 2021-07-13
Require explicit channel type negotiation

t-bast:
This feature was introduced in the spec by https://github.com/lightningnetwork/lightning-rfc/pull/880

We take this opportunity to also clean up the channel state machine to leverage the feature bits added in #237 instead of the somewhat hackish reliance on `isZeroReserve` for a bunch of unrelated Phoenix features.

We should only merge this once the ACINQ node has deployed support for https://github.com/ACINQ/eclair/pull/1867 since we will now require explicit channel type negotiation. And we should probably also update our E2E test repository as well to verify that channel opening works correctly.


src/commonMain/kotlin/fr/acinq/lightning/channel/Channel.kt
```diff
@@ -574,43 +569,51 @@ data class WaitForInit(override val staticParams: StaticParams, override val cur
...
-                        buildList {
-                            add(ChannelTlv.UpfrontShutdownScript(ByteVector.empty))
-                            if (event.channelVersion.isSet(ChannelVersion.ZERO_RESERVE_BIT)) add(ChannelTlv.ChannelVersionTlv(event.channelVersion))
-                            if (event.channelOrigin != null) add(ChannelTlv.ChannelOriginTlv(event.channelOrigin))
-                        }
-                    )
-                )
-                val nextState = WaitForAcceptChannel(staticParams, currentTip, currentOnChainFeerates, event, open)
-                Pair(nextState, listOf(ChannelAction.Message.Send(open)))
+                when (event.channelType) {
```
sstone, line 576: nit: I find that having more detailed high-level branches is often easier to read than nesting when/ifs...
for example here I would write:
```scala
            event is ChannelEvent.InitFunder && (event.channelType == ChannelType.SupportedChannelType.AnchorOutputs || event.channelType == ChannelType.SupportedChannelType.AnchorOutputsZeroConfZeroReserve) -> ...
``` 
or even define a local `isValidChannelType()` method ?

t-bast: Agreed, it's easy to end up with a lot of nesting in kotlin which can become hard to read.
Done in https://github.com/ACINQ/lightning-kmp/pull/277/commits/914d672f81831a595ae6517dd104ccb31c9172d5

sstone (approved):
Tested against endurance (I tested the upgrade process with an existing wallet as well)

t-bast:
Thanks for the tests!

PR #283, opened 2021-07-19
Fix pay-to-open below minimum when using MPP

t-bast:
When using a mix of normal HTLC and pay-to-open in an MPP payment, we incorrectly compared the minimum funding amount to the payment total amount.

We must instead compare it to the amount offered via pay-to-open (excluding HTLCs from this check).

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
@@ -223,7 +223,7 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
                             pending[paymentPart.paymentHash] = payment
                             return ProcessAddResult.Pending(incomingPayment)
                         }
-                        payToOpenMinAmount != null && payToOpenAmount < payToOpenMinAmount -> {
+                        payToOpenMinAmount != null && (payToOpenAmount < payToOpenMinAmount || payToOpenAmount < nodeParams.minFundingSatoshis) -> {
                             // Because of the cost of opening a new channel, there is a minimum amount for incoming payments to trigger
                             // a pay-to-open. Given that the total amount of a payment is included in each payment part, we could have
                             // rejected pay-to-open parts as they arrived, but it would have caused two issues:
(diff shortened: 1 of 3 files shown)


src/commonTest/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandlerTestsCommon.kt
```diff
@@ -554,6 +554,79 @@ class IncomingPaymentHandlerTestsCommon : LightningTestSuite() {
...
+        run {
+            val payToOpenRequest2 = makePayToOpenRequest(incomingPayment, makeMppPayload(amount3, totalAmount, paymentSecret), payToOpenMinAmount = 75_000.msat)
+            val result = paymentHandler.process(payToOpenRequest2, TestConstants.defaultBlockHeight)
+            assertTrue { result is IncomingPaymentHandler.ProcessAddResult.Rejected }
+            val expected = setOf(
+                WrappedChannelEvent(
+                    channelId,
+                    ChannelEvent.ExecuteCommand(CMD_FAIL_HTLC(0, CMD_FAIL_HTLC.Reason.Failure(IncorrectOrUnknownPaymentDetails(totalAmount, TestConstants.defaultBlockHeight.toLong())), commit = true))
+                ),
+                PayToOpenResponseEvent(
```
t-bast, line 597: Shall we take this opportunity to return a single `PayToOpenResponse` instead of one per request?

pm47: I'm not sure, because this creates assumptions on how our node behaves? It could expect exactly one response per request.

t-bast: Sounds good, let's keep it like that then!

pm47 (dismissed):

t-bast:
Rebased and added an additional check for `nodeParams.minFundingSatoshis`.

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -222,7 +223,7 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
                             pending[paymentPart.paymentHash] = payment
                             return ProcessAddResult.Pending(incomingPayment)
                         }
-                        payToOpenMinAmount is MilliSatoshi && payment.amountReceived < payToOpenMinAmount -> {
+                        payToOpenMinAmount != null && (payToOpenAmount < payToOpenMinAmount || payToOpenAmount < nodeParams.minFundingSatoshis) -> {
```
sstone, line 226: nit: the warning that is logged below should also include `payToOpenAmount`

t-bast: Good catch, fixed in https://github.com/ACINQ/lightning-kmp/pull/283/commits/c2d8b93dfdb502629384fe3e1f0296e650671848

sstone (approved):

PR #290, opened 2021-08-03
Subtract fees from pay-to-open amounts in database

dpad85:
The `WalletPayment.amount` field now always reflects the change in the wallet's balance, that is after applying the fees required to complete the payment:
- for outgoing payments, this the total amount that left the wallet, including the fees.
- for incoming payments, this is the total amount that ended up in the wallet, after applying fees if any (e.g for creating a channel)

This changes the way the amount of a pay-to-open is stored in the database: the channel creation fee is now subtracted for the amount when processing the payment in the `IncomingPaymentHandler`. For swap-ins, there are no changes since the channel event's amount was already the actual amount pushed to the channel.

The `WalletPayment` companion object has been removed.

See https://github.com/ACINQ/phoenix/issues/170


robbiehanson (approved):
> "now always reflects the change in the wallet's balance"

That's a great way to think of it. Good idea.

PR #303, opened 2021-09-23
Workaround cinterop issue with PhoenixCrypto

dpad85:
There's an incompatibility between XCode 13 (recently released) and the clang shipped with
Kotlin/Native, causing issues when building `PhoenixCrypto`. The workaround
is to make `NS_FORMAT_ARGUMENT` a noop.

More on this here:
https://youtrack.jetbrains.com/issue/KT-48807#focus=Comments-27-5210791.0-0

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -223,3 +223,10 @@ if (currentOs.isLinux) {
         filter.excludeTestsMatching("*IntegrationTest")
     }
 }
+
+// Make NS_FORMAT_ARGUMENT(1) a no-op
+// This fixes an issue when building PhoenixCrypto using XCode 13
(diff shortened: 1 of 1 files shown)


robbiehanson (approved):
Verified the problem exists with Xcode 13, and this patch fixes the compilation issue.

sstone:
Additional info: The github CI environment  that we use to test and build ios/macos (and android) projects is https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md which includes XCode12.

It should be upgraded to https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md soon, which still seems to use XCode12

PR #304, opened 2021-09-24
Configure github CI to build and store snapshot and release binaries

sstone:
This is the first step towards publishing lightning-kmp on maven central.


dpad85 (approved):
👍

PR #312, opened 2021-12-09
Add support for option_shutdown_anysegwit

t-bast:
Opt-in to allow any future segwit script in shutdown as long as it complies with BIP 141 (see  lightning/bolts#672).
This is particularly useful to allow wallet users to close channels to a Taproot address.
We can probably make it mandatory in wallets such as Phoenix now that Taproot has been activated.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
@@ -384,14 +384,21 @@ object Helpers {
         // used only to compute tx weights and estimate fees
         private val dummyPublicKey by lazy { PrivateKey(ByteArray(32) { 1.toByte() }).publicKey() }
 
-        private fun isValidFinalScriptPubkey(scriptPubKey: ByteArray): Boolean {
+        private fun isValidFinalScriptPubkey(scriptPubKey: ByteArray, allowAnySegwit: Boolean): Boolean {
             return runTrying {
                 val script = Script.parse(scriptPubKey)
-                Script.isPay2pkh(script) || Script.isPay2sh(script) || Script.isPay2wpkh(script) || Script.isPay2wsh(script)
+                when {
+                    Script.isPay2pkh(script) -> true
+                    Script.isPay2sh(script) -> true
+                    Script.isPay2wpkh(script) -> true
+                    Script.isPay2wsh(script) -> true
+                    Script.isNativeWitnessScript(script) -> allowAnySegwit
+                    else -> false
+                }
             }.getOrElse { false }
         }
 
-        fun isValidFinalScriptPubkey(scriptPubKey: ByteVector): Boolean = isValidFinalScriptPubkey(scriptPubKey.toByteArray())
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
```diff
@@ -384,14 +384,21 @@ object Helpers {
...
+        private fun isValidFinalScriptPubkey(scriptPubKey: ByteArray, allowAnySegwit: Boolean): Boolean {
             return runTrying {
                 val script = Script.parse(scriptPubKey)
-                Script.isPay2pkh(script) || Script.isPay2sh(script) || Script.isPay2wpkh(script) || Script.isPay2wsh(script)
+                when {
+                    Script.isPay2pkh(script) -> true
+                    Script.isPay2sh(script) -> true
+                    Script.isPay2wpkh(script) -> true
+                    Script.isPay2wsh(script) -> true
+                    Script.isNativeWitnessScript(script) -> allowAnySegwit
```
sstone, line 395: https://github.com/lightning/bolts/blob/master/02-peer-protocol.md#closing-initiation-shutdown specifies V1 to V16 witness scripts but `Script.isNativeWitnessScript()` would also match V0 witness scripts (see https://github.com/ACINQ/bitcoin-kmp/blob/master/src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt#L257). 
I'm not sure what the best fix is, maybe for now we could as something like this here:
```scala
        private fun pushSize(op: ScriptElt): Int? = when (op) {
            is OP_PUSHDATA -> op.data.size()
            else -> null
        }

        fun getWitnessVersion(script:  List<ScriptElt>): Int? = when {
            script.size != 2 -> null
            script[0] == OP_0 && (script[1].isPush(20) || script[1].isPush(32)) -> 0
            Script.simpleValue(script[0]) in 1..16 && pushSize(script[1]) in 2..40 -> Script.simpleValue(script[0]).toInt()
            else -> null
        }
```
and also add these to bitcoin-kmp ?

sstone (changes requested):

src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
t-bast: Good catch. I think we should do something much simpler though.
The condition here should be `isNativeWitnessScript with version > 0`, so I'll just update it to:

```kotlin
// option_shutdown_anysegwit doesn't cover segwit v0
Script.isNativeWitnessScript(script) && script[0] != OP_0 -> allowAnySegwit
```

Which should perfectly match the spec.
I'll independently add these functions to `bitcoin-kmp`, they seem useful as general tools.

t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/312/commits/83c061715d235d1107b41834f62853f71eec85c6

sstone (approved):

PR #313, opened 2021-12-09
Add support for `option_payment_metadata`

t-bast:
Add support for `option_payment_metadata` as specified in https://github.com/lightning/bolts/pull/912
This feature lets wallets generate "light" invoices without storing them in the local DB until they are paid.

The trick is to simply put the important data (mainly `payment_secret` and `total_amount`) in an encrypted blob inside the invoice, and verify its presence and integrity when receiving a payment that contains a `payment_metadata` field (and insert it in the DB on-the-fly).

This PR only implements support for the sender side, as recipients will only be able to leverage this features once all senders support it.

We do include a `payment_metadata` in all invoices we generate, which lets us know whether our payers support this feature: this is important data to have before we decide to make it mandatory and actually implement storage-less invoices.

We also take this opportunity to have our first invoice-only feature to implement feature bit filtering, as required by the spec (`IN9` flags as specified in [bolt 9](https://github.com/lightning/bolts/blob/master/09-features.md)): see the first commit of the PR.

The second commit does not contain any logic and simply moves data classes around and rename them, it's best reviewed independently.

The third commit contains the actual logic for `option_payment_metadata`.


src/commonMain/kotlin/fr/acinq/lightning/Features.kt
```diff
@@ -21,6 +21,9 @@ sealed class Feature {
     abstract val rfcName: String
     abstract val mandatory: Int
     val optional: Int get() = mandatory + 1
+    abstract val init: Boolean
+    abstract val nodeAnnouncement: Boolean
+    abstract val invoice: Boolean
```
sstone, line 26: there is no `channelAnnouncement` flag, is it because we assume that channels will always be private ?

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -84,8 +87,10 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
             nodeParams.nodePrivateKey,
             description,
             PaymentRequest.DEFAULT_MIN_FINAL_EXPIRY_DELTA,
-            invoiceFeatures,
+            nodeParams.features,
             randomBytes32(),
+            // We always include a payment metadata in our invoices, which lets us test whether senders support it
+            randomBytes(64).toByteVector(),
```
sstone, line 93: nit:  use `randomBytes64()` ?

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -84,8 +87,10 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
             nodeParams.nodePrivateKey,
             description,
             PaymentRequest.DEFAULT_MIN_FINAL_EXPIRY_DELTA,
-            invoiceFeatures,
+            nodeParams.features,
```
sstone, line 90: why not `nodeParams.features.initFeatures()` ?

src/commonMain/kotlin/fr/acinq/lightning/Features.kt
t-bast: It could be a good reason, but it's just because there's simply no channel feature yet in the spec, and it's unclear how it would even be used.

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
t-bast: Updated in https://github.com/ACINQ/lightning-kmp/pull/313/commits/aa81e7d069416777250c9bc709955cf6b8898f60 to use a single byte to avoid making invoices too big

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/313/commits/abd2f4c6e67fa4967a1a67a21eca93a61c3ca39a, the feature bits were already filtered inside `PaymentRequest.create` but it still makes sense to do it here as well.

sstone (approved):

PR #317, opened 2022-01-20
Allow for user-configurable fee options

robbiehanson:
Phoenix allows the user to configure the maximum allowable fee ([Issue 191](https://github.com/ACINQ/phoenix/issues/191))

However, we need more flexibility in lightning-kmp to allow for this.

In the current version, the fee schedule (`List<TrampolineFee>`) is configured when the Peer is created, and isn't configurable. This PR allows the default fee schedule to be overriden when making a payment.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -42,7 +42,7 @@ sealed class PaymentEvent : PeerEvent()
 data class ReceivePayment(val paymentPreimage: ByteVector32, val amount: MilliSatoshi?, val description: String, val expirySeconds: Long? = null, val result: CompletableDeferred<PaymentRequest>) : PaymentEvent()
 object CheckPaymentsTimeout : PaymentEvent()
 data class PayToOpenResponseEvent(val payToOpenResponse: PayToOpenResponse) : PeerEvent()
-data class SendPayment(val paymentId: UUID, val amount: MilliSatoshi, val recipient: PublicKey, val details: OutgoingPayment.Details.Normal, val trampolineFees: List<TrampolineFees>? = null) : PaymentEvent() {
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -42,7 +42,7 @@ sealed class PaymentEvent : PeerEvent()
 data class ReceivePayment(val paymentPreimage: ByteVector32, val amount: MilliSatoshi?, val description: String, val expirySeconds: Long? = null, val result: CompletableDeferred<PaymentRequest>) : PaymentEvent()
 object CheckPaymentsTimeout : PaymentEvent()
 data class PayToOpenResponseEvent(val payToOpenResponse: PayToOpenResponse) : PeerEvent()
-data class SendPayment(val paymentId: UUID, val amount: MilliSatoshi, val recipient: PublicKey, val details: OutgoingPayment.Details.Normal) : PaymentEvent() {
+data class SendPayment(val paymentId: UUID, val amount: MilliSatoshi, val recipient: PublicKey, val details: OutgoingPayment.Details.Normal, val trampolineFees: List<TrampolineFees>? = null) : PaymentEvent() {
```
t-bast, line 45: nit: I'd make it explicit that it's an override (and there's a default value coming from elsewhere) so that applications don't feel compelled to always provide it (and better understand why it's optional).

```suggestion
data class SendPayment(val paymentId: UUID, val amount: MilliSatoshi, val recipient: PublicKey, val details: OutgoingPayment.Details.Normal, val trampolineFeesOverride: List<TrampolineFees>? = null) : PaymentEvent() {
```

t-bast (commented):
That sounds perfectly reasonable :+1:

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
robbiehanson: Good idea

robbiehanson: Done in ef0c044

t-bast (approved):
LGTM :+1:

dpad85 (approved):

PR #318, opened 2022-02-02
Add simple LN node application

sstone:
Add a simple JVM node application, with an sqlite database, and a docker file. A docker image will be automatically built from the master branch and pushed to docker hub. 
It can be used for local tests/debug, and also for integration tests.


sstone:
This change includes a lot of code that was copied from https://github.com/ACINQ/phoenix/ (db implementation for example) and will be tedious to update, a better approach would be to add an JVM app and docker image to the https://github.com/ACINQ/phoenix/.

dpad85:
Adding a `phoenix-jvm` module in the phoenix repository would make sense. Code (like db implementation) would be shared. Also, lightning-kmp is a library, not a node, so I think it would be cleaner to move an application layer to the another repository like the phoenix one.

PR #319, opened 2022-02-02
Add a tlv to facilitate phoenix migration

pm47:
When connecting to our node, phoenix-kmp attaches a special tlv to its
`init` message with the `nodeId` of the legacy phoenix-eclair. Our node
later sends a `warning` message indicating whether the legacy `nodeId`
has any channels attached to it.


src/commonMain/kotlin/fr/acinq/lightning/wire/InitTlv.kt
```diff
@@ -31,4 +32,22 @@ sealed class InitTlv : Tlv {
             }
         }
     }
+
+    @Serializable
+    data class PhoenixAndroidLegacyNodeId(@Contextual val legacyNodeId: PublicKey) : InitTlv() {
+        override val tag: Long get() = Networks.tag
```
t-bast, line 38:
```suggestion
        override val tag: Long get() = PhoenixAndroidLegacyNodeId.tag
```

t-bast: How is the unit test passing...??

src/commonTest/kotlin/fr/acinq/lightning/wire/InitTlvTestsCommon.kt
```diff
@@ -0,0 +1,33 @@
...
+        val readers = mapOf(
+            InitTlv.PhoenixAndroidLegacyNodeId.tag to InitTlv.PhoenixAndroidLegacyNodeId.Companion as TlvValueReader<InitTlv>
+        )
+        val tlvStreamSerializer = TlvStreamSerializer(false, readers)
+
+        testCases.forEach {
+            val decoded = tlvStreamSerializer.read(it.second)
+            val legacyNodeId = decoded.records.mapNotNull { record ->
+                when (record) {
+                    is InitTlv.PhoenixAndroidLegacyNodeId -> record.legacyNodeId
```
sstone, line 26: adding a check on the record's tag here (for example `assertEquals(InitTlv.PhoenixAndroidLegacyNodeId.tag, record.tag`) would fail (see @t-bast's comment above)

t-bast: Oh right, the test passes because when we register the `reader` we correctly use the object tag, it's only on the instances that the tag is incorrect

sstone: This could be detected automatically if we modify `TlvStreamSerializer.readTlvs()` and check that the tag of the record that was deserialized matches the tag that was read from the input stream and used to select a reader in the internal reader map ?

pm47: Done with 6f2bee6, I'll let the modification to `TlvStreamSerializer.readTlvs()` for another PR.

src/commonMain/kotlin/fr/acinq/lightning/wire/InitTlv.kt
pm47: See https://github.com/ACINQ/lightning-kmp/pull/319#discussion_r798327722

t-bast (approved):

sstone (dismissed):

dpad85 (approved):

PR #322, opened 2022-02-21
Delete expired payments

remyers:
This is the first step to bring automatic purging of expired incoming payments to projects built on lightning-kmp. This ability was requested in issue [#2078](https://github.com/ACINQ/eclair/issues/2078)  in Eclair.

This PR adds functions to find and remove expired incoming payments similar to Eclair PR [#2174](https://github.com/ACINQ/eclair/pull/2174). This PR does not launch an automatic thread for purging expired invoices like in Eclair because mobile implementations may not always be running and may want to purge invoices using a platform specific trigger.

Additional tasks are needed to bring this to Phoenix, including:
[  ] add  platform specific database support for new functions added to `InMemoryPaymentsDb` eg. [SqlitePaymentsDb.kt](https://github.com/ACINQ/phoenix/blob/master/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/db/SqlitePaymentsDb.kt)
[  ] add platform specific tests
[  ] add platform specific triggers to call PurgeExpiredPayments

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
@@ -54,7 +54,7 @@ interface IncomingPaymentsDb {
     suspend fun listExpiredPayments(fromCreatedAt: Long = 0, toCreatedAt: Long = currentTimestampMillis()): List<IncomingPayment>
 
     /** Remove a pending incoming payment.*/
-    suspend fun removeIncomingPayment(paymentHash: ByteVector32)
+    suspend fun removeIncomingPayment(paymentHash: ByteVector32): Boolean
 }
 
 interface OutgoingPaymentsDb {
--- a/src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
@@ -46,6 +46,12 @@ interface IncomingPaymentsDb {
 
     /** List received payments (with most recent payments first). */
     suspend fun listReceivedPayments(count: Int, skip: Int, filters: Set<PaymentTypeFilter> = setOf()): List<IncomingPayment>
+
+    /** List expired unpaid normal payments created within specified time range (with the most recent payments first). */
+    suspend fun listExpiredPayments(fromCreatedAt: Long = 0, toCreatedAt: Long = currentTimestampMillis()): List<IncomingPayment>
+
+    /** Remove a pending incoming payment.*/
+    suspend fun removeIncomingPayment(paymentHash: ByteVector32)
 }
 
 interface OutgoingPaymentsDb {
--- a/src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
@@ -47,6 +47,9 @@ interface IncomingPaymentsDb {
     /** List received payments (with most recent payments first). */
     suspend fun listReceivedPayments(count: Int, skip: Int, filters: Set<PaymentTypeFilter> = setOf()): List<IncomingPayment>
 
+    /** List incoming payments (with the most recent payments first). */
+    suspend fun listIncomingPayments(count: Int, skip: Int, filters: Set<PaymentTypeFilter> = setOf()): List<IncomingPayment>
+
     /** List expired unpaid normal payments created within specified time range (with the most recent payments first). */
     suspend fun listExpiredPayments(fromCreatedAt: Long = 0, toCreatedAt: Long = currentTimestampMillis()): List<IncomingPayment>
 
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
@@ -391,11 +391,9 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
      * @return number of invoices purged
      */
     suspend fun purgeExpiredPayments(fromCreatedAt: Long = 0, toCreatedAt: Long = currentTimestampMillis()): Int {
-        val expiredInvoices = db.listExpiredPayments(fromCreatedAt, toCreatedAt)
-        expiredInvoices.forEach { payment ->
-            db.removeIncomingPayment(payment.paymentHash)
-        }
-        return expiredInvoices.size
+        return db.listExpiredPayments(fromCreatedAt, toCreatedAt).map {
+            db.removeIncomingPayment(it.paymentHash)
+        }.count()
     }
 
     companion object {
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
@@ -5,7 +5,6 @@ import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.bitcoin.Crypto
 import fr.acinq.bitcoin.PrivateKey
 import fr.acinq.lightning.CltvExpiry
-import fr.acinq.lightning.Lightning.randomBytes
 import fr.acinq.lightning.Lightning.randomBytes32
 import fr.acinq.lightning.MilliSatoshi
 import fr.acinq.lightning.NodeParams
@@ -382,6 +381,21 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
 
         pending.minusAssign(keysToRemove)
         return actions
+    }
+
+    /**
+     * Purge all expired unpaid normal payments with creation times in the given time range.
+     *
+     * @param fromCreatedAt from absolute time in milliseconds since UNIX epoch when the payment request was generated.
+     * @param toCreatedAt to absolute time in milliseconds since UNIX epoch when the payment request was generated.
+     * @return number of invoices purged
+     */
+    suspend fun purgeExpiredPayments(fromCreatedAt: Long = 0, toCreatedAt: Long = currentTimestampMillis()): Int {
+        val expiredInvoices = db.listExpiredPayments(fromCreatedAt, toCreatedAt)
+        expiredInvoices.forEach { payment ->
+            db.removeIncomingPayment(payment.paymentHash)
+        }
+        return expiredInvoices.size
     }
 
     companion object {
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
@@ -391,9 +391,9 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
      * @return number of invoices purged
      */
     suspend fun purgeExpiredPayments(fromCreatedAt: Long = 0, toCreatedAt: Long = currentTimestampMillis()): Int {
-        return db.listExpiredPayments(fromCreatedAt, toCreatedAt).map {
+        return db.listExpiredPayments(fromCreatedAt, toCreatedAt).count {
             db.removeIncomingPayment(it.paymentHash)
-        }.count()
+        }
     }
 
     companion object {
--- a/src/commonTest/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandlerTestsCommon.kt
+++ b/src/commonTest/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandlerTestsCommon.kt
@@ -11,6 +11,7 @@ import fr.acinq.lightning.crypto.sphinx.Sphinx
 import fr.acinq.lightning.db.InMemoryPaymentsDb
 import fr.acinq.lightning.db.IncomingPayment
 import fr.acinq.lightning.db.IncomingPaymentsDb
+import fr.acinq.lightning.db.PaymentTypeFilter
 import fr.acinq.lightning.io.PayToOpenResponseEvent
 import fr.acinq.lightning.io.WrappedChannelEvent
 import fr.acinq.lightning.router.ChannelHop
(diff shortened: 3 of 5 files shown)


remyers:
@t-bast - a few questions:
* is it the correct approach to defer automatic purging to platform specific app implementations instead of in the lightning-kmp library?
* OK to continue with next step as PRs to the Phoenix repo before this is merged?

t-bast:
> is it the correct approach to defer automatic purging to platform specific app implementations instead of in the lightning-kmp library?

Yes, I think it makes sense to let the application call us to purge expired invoices. Let's first add that capability and then we'll see with @dpad85 how Phoenix will use it.

> OK to continue with next steps as PRs to the Phoenix repo before this is merged?

I'm not sure I fully understand what you ask, but I think you should simply add the ability to purge expired invoices on-demand to lightning-kmp, and we'll let @dpad85 and @robbiehanson handle the Phoenix side and call it.

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
```diff
@@ -46,6 +46,15 @@ interface IncomingPaymentsDb {
 
     /** List received payments (with most recent payments first). */
     suspend fun listReceivedPayments(count: Int, skip: Int, filters: Set<PaymentTypeFilter> = setOf()): List<IncomingPayment>
+
+    /** List incoming payments (with the most recent payments first). */
+    suspend fun listIncomingPayments(count: Int, skip: Int, filters: Set<PaymentTypeFilter> = setOf()): List<IncomingPayment>
```
t-bast, line 51: I'm not sure we should add this function, I don't see it being used by wallets.
Users are only interested in listing the payments that they received.
@dpad85 let us know what you think.

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -384,6 +383,21 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
...
     }
 
+    /**
+     * Purge all expired unpaid normal payments with creation times in the given time range.
+     *
+     * @param fromCreatedAt from absolute time in milliseconds since UNIX epoch when the payment request was generated.
+     * @param toCreatedAt to absolute time in milliseconds since UNIX epoch when the payment request was generated.
+     * @return number of invoices purged
+     */
+    suspend fun purgeExpiredPayments(fromCreatedAt: Long = 0, toCreatedAt: Long = currentTimestampMillis()): Int {
```
t-bast, line 393: Applications building on top of `lightning-kmp` won't be able to call this function right now, as the payment handler is a private field in a `Peer` instance.

Can you create a new [payment event](https://github.com/ACINQ/lightning-kmp/blob/234e049f7ee93296cfa03caa0e9de7b1f2fe33ae/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt#L41), for example `data class PurgeExpiredPayments(val fromCreatedAt: Long, val toCreatedAt: Long) : PaymentEvent()` and add a handler for it in the `processEvent` function?
That will let Phoenix easily interact with this feature.

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
```diff
@@ -46,6 +46,15 @@ interface IncomingPaymentsDb {
...
+
+    /** List expired unpaid normal payments created within specified time range (with the most recent payments first). */
+    suspend fun listExpiredPayments(fromCreatedAt: Long = 0, toCreatedAt: Long = currentTimestampMillis()): List<IncomingPayment>
+
+    /** Remove a pending incoming payment.*/
+    suspend fun removeIncomingPayment(paymentHash: ByteVector32)
```
t-bast, line 57: What about returning a boolean here to indicate whether the payment was removed or not?
This would let us return `false` for paid (or partially paid) payments.

src/commonTest/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandlerTestsCommon.kt
```diff
@@ -1149,6 +1151,38 @@ class IncomingPaymentHandlerTestsCommon : LightningTestSuite() {
...
+
+        // create incoming payment that has expired and been paid
+        val paidInvoice = paymentHandler.createInvoice(defaultPreimage, defaultAmount, "paid", listOf(), expirySeconds = 3600,
+            timestampSeconds = currentTimestampSeconds() - 3600 + 2)
+
+        // pay incoming invoice before it expires
+        val add = makeUpdateAddHtlc(0, randomBytes32(), paymentHandler, paidInvoice.paymentHash, makeMppPayload(defaultAmount, defaultAmount, paidInvoice.paymentSecret))
+        val result = paymentHandler.process(add, TestConstants.defaultBlockHeight)
+        assertTrue { result is IncomingPaymentHandler.ProcessAddResult.Accepted }
+        delay(2_000) // wait for 2 seconds so paid invoice expires
```
t-bast, line 1170: I'd like to avoid introducing delays in unit tests, otherwise that tends to creep in many places and we end up with a very slow test runs. We really want our test suite to run as fast as it can.

We can achieve by creating the received payment directly in the DB instead of going through `paymentHandler.process`:

```kotlin
val db = InMemoryPaymentsDb()
val paymentHandler = IncomingPaymentHandler(TestConstants.Bob.nodeParams, TestConstants.Bob.walletParams, db)

// create incoming payment that has expired and not been paid
val expiredInvoice = paymentHandler.createInvoice(randomBytes32(), defaultAmount, "expired", listOf(), expirySeconds = 3600,
    timestampSeconds = currentTimestampSeconds() - 3600 - 60)

// create incoming payment that has expired, but has been paid before its expiry
val paidInvoice = paymentHandler.createInvoice(defaultPreimage, defaultAmount, "paid", listOf(), expirySeconds = 3600,
    timestampSeconds = currentTimestampSeconds() - 3600 - 60)
db.receivePayment(defaultPaymentHash, setOf(IncomingPayment.ReceivedWith.LightningPayment(defaultAmount, randomBytes32(), 3)), receivedAt = currentTimestampSeconds() - 1800)
```

t-bast (commented):
Looks mostly good, just a few improvements to consider.

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
remyers: Ah ha, I didn't think of that. Should PurgeExpiredPayments notify the calling application by sending a PeerListenerEvent like this then?
`object PurgeExpiredPaymentsCompleted : PeerListenerEvent()` 

(assuming I change to return a boolean and not number of purged invoices)

t-bast: Good question, I don't know if the application is really interested about knowing that it has completed.
I think they only want to trigger this as a fire-and-forget task, but @dpad85 or @robbiehanson will know better.
I'd say that for now we don't emit an event when it completes, and we can add it if they think it's useful?

remyers: Actually shouldn't be Boolean message, but can return the number of successfully purged expired invoices :
data class PurgeExpiredPaymentsCompleted(val count: Int) : PeerListenerEvent()

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
remyers: fixed in [be38964](https://github.com/ACINQ/lightning-kmp/pull/322/commits/be38964082f80475adeb7ff0f2bd9455988f5afb); I also only count successfully removed payments in the `IncomingPaymentsHandler`.

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
remyers: I added it to be extra careful in the test and check the exact state of the DB before and, more importantly, after purging invoices.

If it's only useful for testing so perhaps I should take it out. I can instead use the fact that only one incoming payment is purged, but not two, to ensure that partially paid payments aren't purged.

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
remyers: On second thought, `listExpiredPayments` will not return paid (or partially paid) incoming payments so it should be a rare race condition to have on that list expired payments that were [partially] paid. It is still worth keeping this change though for the case where `removeIncomingPayment` is used for something where an absent payment is more likely (or important).

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
remyers: Added `PurgeExpiredPayments` in [53a9360](https://github.com/ACINQ/lightning-kmp/pull/322/commits/53a9360cad36172ef6123ba8b592e62500a18da4) but I will hold off on a return message until hearing what @dpad85 and @robbiehanson think.

remyers: Btw, what is the policy towards removing warnings in a file I changed that weren't caused by my changes? They could just be caused by differences in linter settings, but there were a few in Peer.kt .

src/commonTest/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandlerTestsCommon.kt
remyers: Nice! It's easier to read this way too. Fixed in [abd5760](https://github.com/ACINQ/lightning-kmp/pull/322/commits/abd576080cd4b240c509346e83efbb90ac2b6a71).

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
t-bast: Let's wait for dpad85 to decide!

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
t-bast: There are a few about fields that could be private, but I think those shouldn't be fixed because Phoenix actually uses those fields so it needs to have access to it (maybe there's a kotlin tag to tell the compiler that these fields really need to stay public?). Apart from those, I think it's always good to fix a few warnings when you see them, even if they're unrelated to your change ;)

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
t-bast: Agreed, we will probably never see it return `false` because we're using it correctly, but it makes sense to keep that boolean for people who may use this function differently in the future.

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -384,6 +383,19 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
...
+        return db.listExpiredPayments(fromCreatedAt, toCreatedAt).map {
```
t-bast, line 394: I think you want to use `filter` instead of `map` to count only the ones that actually get removed?

remyers: Doh! yes, actually it seems `count` applied to `removeIncomingPayment` as a predicate is what I really wanted here. Fixed in [290d8df](https://github.com/ACINQ/lightning-kmp/pull/322/commits/290d8dfcd3afb500aee2bbea91e06a14a816ca91).

t-bast (approved):
LGTM, well done! Let's wait for @dpad85 or @robbiehanson to chime in on some of the comments, feel free to merge when they're satisfied.

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
dpad85: Regarding the return type of `purgeExpiredPayments`, we don't need that information on Phoenix's side. We do not even need to know it succeeded.

Also, since in Phoenix the expiration information we need is [actually stored as JSON in the `incoming_payments` database](https://github.com/ACINQ/phoenix/blob/19c5f252ef2be7e107f989d1ab034b3208a4ef79/phoenix-shared/src/commonMain/paymentsdb/fr.acinq.phoenix.db/IncomingPayments.sq#L11), we won't be able to make a single SQL query that would delete things in one go. Instead, we'll have to make one SQL query for each payments in the given range, go back to kotlin, deserizalize the json data, check the expiration status and then actually make a new SQL query to delete that payment. So we can expect this operation to be a quite slow! 

Btw, I note that in our implementation of the query for expired payments, we should first check that the `received` rows are null for better perfs.

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
dpad85: Indeed we don't need it in the app, but it's nice to have that check in the tests. So I think we should keep it, especially if `purgeExpiredPayments` does not return the affected rows count anymore (which affects the test).

robbiehanson:
I think this makes some sense from an interface perspective. That is, when building atop of lightning-kmp, the developer will be asked to implement functions to search for expired payments. Which is a good reminder that this task should be performed.





The actual implementation of this task will be highly dependent upon how they structure the database layer. In the case of Phoenix, we need to:

- check that `received_at` is null
- deserialize the `origin_blob` to see if the payment is expired



So what we will *probably* do is something like:

- add an index on `receieved_at`
- fetch a batch (10?) of rows where `received_at` is null, sorted by `created_at`
- delete the expired items
- stop if the batch was empty, or any were not expired
- repeat as needed, each iteration using a different database transaction



This means we might not implement `listExpiredPayments` in Phoenix, instead opting for an optimized garbageCollection function.

PR #325, opened 2022-03-02
Improve legacy channels TLV mechanism and response from Peer

dpad85:
This PR improves the logic for requesting information about legacy channels from the Peer.

- The `KeyManager` now exposes the legacy node id as computed on eclair-core.

- The `Peer` object has a new, optional TLV stream parameter which can be used to enrich the `Init` message with TLVs.

- The `PhoenixAndroidLegacyNodeId` TLV now contains a `signature` field signing the `legacyNodeId` to prevent information leak.
```kt
data class PhoenixAndroidLegacyNodeId(val legacyNodeId: PublicKey, val signature: ByteVector64) : InitTlv()
```

- A new `LightningMessage` has been added. This message is sent by the peer after receiving the aforementioned tlv, and signals the presence of channels for the provided legacy node id.
```kt
data class PhoenixAndroidLegacyInfo(val hasChannels: Boolean) : LightningMessage {
```

- a new `PeerListenerEvent` has been added containing the legacy channels information, in order to be consumed by Phoenix in its migration logic.


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -59,12 +59,24 @@ data class SwapInResponseEvent(val swapInResponse: SwapInResponse) : PeerListene
...
+ * @param socketBuilder Builds the TCP socket used to connect to the Peer.
+ */
 @ObsoleteCoroutinesApi
 @OptIn(ExperimentalStdlibApi::class, ExperimentalCoroutinesApi::class, ExperimentalTime::class)
 class Peer(
     val nodeParams: NodeParams,
     val walletParams: WalletParams,
-    val watcher: ElectrumWatcher,
+    private val initTlvStream: TlvStream<InitTlv> = TlvStream.empty(),
+    private val watcher: ElectrumWatcher,
```
robbiehanson, line 80: This change is incompatible with iOS, which needs access to the ElectrumWatcher. In particular, it uses `peer.watcher.openUpToDateFlow()`. It needs this information as part of the iOS run-in-the-background routine.

dpad85: Fixed in 2fedf27

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -1295,3 +1296,22 @@ data class SwapInConfirmed(
...
+
+@OptIn(ExperimentalUnsignedTypes::class)
+data class PhoenixAndroidLegacyInfo(
+    val hasChannels: Boolean
+) : LightningMessage {
+    override val type: Long get() = PhoenixAndroidLegacyInfo.type
+
+    override fun write(out: Output) {
+       LightningCodecs.writeByte(if (hasChannels) 1 else 0, out)
+    }
```
pm47, line 1308: Do you really need to implement the _writing_ part?

src/commonTest/kotlin/fr/acinq/lightning/wire/InitTlvTestsCommon.kt
```diff
@@ -1,16 +1,22 @@
...
 import kotlin.test.assertEquals
+import kotlin.test.assertTrue
 
 class InitTlvTestsCommon : LightningTestSuite() {
     @Test
     fun `legacy phoenix TLV`() {
         val testCases = listOf(
-            Pair(PublicKey(Hex.decode("03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f")), Hex.decode("fe47020001 21 03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f"))
+            Pair(
+                first = DeterministicWallet.generate(MnemonicCode.toSeed("ghost runway obscure", passphrase = "")).publicKey,
```
pm47, line 17: Why not use an actual real-life Phoenix seed?

pm47 (commented):
A few nits, otherwise LGTM

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
dpad85: No we don't really need it and can remove it, although we the test would have to be updated.

src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
```diff
@@ -11,6 +11,7 @@ import fr.acinq.lightning.transactions.Transactions
 data class LocalKeyManager(val seed: ByteVector, val chainHash: ByteVector32) : KeyManager {
 
     private val master = DeterministicWallet.generate(seed)
+    override val legacyNodeKey: DeterministicWallet.ExtendedPrivateKey = derivePrivateKey(master, eclairNodeKeyBasePath(chainHash))
```
sstone, line 14: We should have a test to check that this key is correct (with test data generated from the legacy app).

sstone (changes requested):

src/commonTest/kotlin/fr/acinq/lightning/wire/InitTlvTestsCommon.kt
dpad85: Done in 6362ac4.

pm47 (approved):

PR #326, opened 2022-03-04
Refactoring iOS socket code - moving it into Swift layer

robbiehanson:
This PR simplifies the native iOS socket code.

In the past, this code was written in Kotlin. As in, we were attempting to call various native libraries thru some goofy bridge layer. The resulting code required several workarounds due to bugs in the bridge layer (see `ios_network_framework.def`). Further, it was difficult to maintain because nearly **all** of the documentation & sample code on the internet is in Swift.

Luckily we've already figured out how to write native Swift code, and consume the resulting library from within Kotlin! We did this when we implemented native ChaChaPoly routines.

So this PR migrates the socket code to Swift, and the existing Kotlin class has been rewritten to simply call functions in the compiled Swift module.

Note: This is prep work for my next PR, which allows pinning publicKeys for TLS. Which we intend to use to increase security for Electrum connections (as many people use self-signed certificates).


robbiehanson:
Related to https://github.com/ACINQ/phoenix/issues/182

dpad85 (approved):
🧙

PR #327, opened 2022-03-04
TcpSocket now supports pinning publicKeys for TLS

robbiehanson:
The TcpSocket previously only supported 2 types of TLS connections:

```kotlin
enum class TLS {
   SAFE, UNSAFE_CERTIFICATES
}
```

This PR adds support for pinning. The motivation is to increase security for Electrum connections. Since many (most?) of them are using self-signed certificates.

```kotlin
sealed class TLS {
   object DISABLED: TLS()
   object TRUSTED_CERTIFICATES: TLS()
   object UNSAFE_CERTIFICATES: TLS()
   data class PINNED_PUBLIC_KEY(
      // DER-encoded publicKey as base64 string.
      // (I.e. same as PEM format, without BEGIN/END header/footer)
      val pubKey: String
   ): TLS()
}
```

One other minor change I made was to make the TLS parameter required.

```kotlin
suspend fun connect(host: String, port: Int, tls: TLS): TcpSocket
//                                           ^^^^^^^^
```

Seems safer - no longer possible to forget the parameter, and end up with a non-secure connection when a secure connection was expected.

#### Important
The initial commit only includes support for the iOS socket. We need another commit to add support for the Android socket.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/TcpSocket.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/TcpSocket.kt
@@ -23,12 +23,19 @@ interface TcpSocket {
 
     fun close()
 
-    enum class TLS {
-        SAFE, UNSAFE_CERTIFICATES
+    sealed class TLS {
+        object DISABLED: TLS()
+        object TRUSTED_CERTIFICATES: TLS()
+        object UNSAFE_CERTIFICATES: TLS()
+        data class PINNED_PUBLIC_KEY(
+            // DER-encoded publicKey as base64 string.
+            // (I.e. same as PEM format, without BEGIN/END header/footer)
+            val pubKey: String
+        ): TLS()
     }
 
     interface Builder {
-        suspend fun connect(host: String, port: Int, tls: TLS? = null): TcpSocket
+        suspend fun connect(host: String, port: Int, tls: TLS): TcpSocket
 
         companion object {
             operator fun invoke(): Builder = PlatformSocketBuilder
@@ -49,5 +56,5 @@ fun TcpSocket.linesFlow(): Flow<String> =
             emit(buffer.subArray(size))
         }
     }
-        .decodeToString()
-        .splitByLines()
+    .decodeToString()
+    .splitByLines()
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/TcpSocket.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/TcpSocket.kt
@@ -24,17 +24,21 @@ interface TcpSocket {
     fun close()
 
     sealed class TLS {
-        // used for Lightning connections
+        /** used for Lightning connections */
         object DISABLED: TLS()
-        // used for Electrum servers when expecting a valid certificate
+        /** used for Electrum servers when expecting a valid certificate */
         object TRUSTED_CERTIFICATES: TLS()
-        // only used in unit tests
+        /** only used in unit tests */
         object UNSAFE_CERTIFICATES: TLS()
-        // used for Electrum servers when expecting a specific public key
-        // (for example self-signed certificates)
+        /**
+         * used for Electrum servers when expecting a specific public key
+         * (for example self-signed certificates)
+         */
         data class PINNED_PUBLIC_KEY(
-            // DER-encoded publicKey as base64 string.
-            // (I.e. same as PEM format, without BEGIN/END header/footer)
+            /**
+             *  DER-encoded publicKey as base64 string.
+             * (I.e. same as PEM format, without BEGIN/END header/footer)
+             */
             val pubKey: String
         ): TLS()
     }
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/TcpSocket.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/TcpSocket.kt
@@ -24,23 +24,26 @@ interface TcpSocket {
     fun close()
 
     sealed class TLS {
-        /** used for Lightning connections */
-        object DISABLED: TLS()
-        /** used for Electrum servers when expecting a valid certificate */
-        object TRUSTED_CERTIFICATES: TLS()
-        /** only used in unit tests */
-        object UNSAFE_CERTIFICATES: TLS()
+        /** Used for Lightning connections */
+        object DISABLED : TLS()
+
+        /** Used for Electrum servers when expecting a valid certificate */
+        object TRUSTED_CERTIFICATES : TLS()
+
+        /** Only used in unit tests */
+        object UNSAFE_CERTIFICATES : TLS()
+
         /**
-         * used for Electrum servers when expecting a specific public key
+         * Used for Electrum servers when expecting a specific public key
          * (for example self-signed certificates)
          */
         data class PINNED_PUBLIC_KEY(
             /**
-             *  DER-encoded publicKey as base64 string.
+             * DER-encoded publicKey as base64 string.
              * (I.e. same as PEM format, without BEGIN/END header/footer)
              */
             val pubKey: String
-        ): TLS()
+        ) : TLS()
     }
 
     interface Builder {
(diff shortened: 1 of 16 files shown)


robbiehanson:
Related to https://github.com/ACINQ/phoenix/issues/182

src/commonMain/kotlin/fr/acinq/lightning/io/TcpSocket.kt
```diff
@@ -23,12 +23,19 @@ interface TcpSocket {
...
-        SAFE, UNSAFE_CERTIFICATES
+    sealed class TLS {
+        object DISABLED: TLS()
+        object TRUSTED_CERTIFICATES: TLS()
+        object UNSAFE_CERTIFICATES: TLS()
+        data class PINNED_PUBLIC_KEY(
+            // DER-encoded publicKey as base64 string.
+            // (I.e. same as PEM format, without BEGIN/END header/footer)
+            val pubKey: String
+        ): TLS()
```
dpad85, line 34: nit: I think it could be nice to add comments here, something like:
- `DISABLED`: used for Lightning connections
- `TRUSTED_CERTIFICATES`: used for Electrum servers when expecting a valid certificate
- `PINNED_PUBLIC_KEY`: used for Electrum servers when expecting a specific public key (for example self-signed certificates)
- `UNSAFE_CERTIFICATES`: only used in tests.

robbiehanson: Good idea :thumbsup:

robbiehanson: Done in [1e29485](https://github.com/ACINQ/lightning-kmp/pull/327/commits/1e294851f976629450e1f97a87f124a2a485144c)

dpad85:
Support for certificate pinning added in 9b019b16f7e6c3fab69a1ce122b2d0061e842d5e.

@sstone you may want to take a look. The idea is : 
- create a default `X509TrustManager` with default behaviour ;
- create a new `X509TrustManager` with a special `checkServerTrusted` method that whitelists the pinned pubkey ;
- if the server's pubkey does not match the pinned pubkey, fall back to default behaviour.

src/jvmMain/kotlin/fr/acinq/lightning/io/JvmTcpSocket.kt
```diff
@@ -58,34 +64,91 @@ class JvmTcpSocket(val socket: Socket) : TcpSocket {
...
+                                    } catch (e: Exception) {
+                                        logger.error(e) { "failed to read pinned pubkey=${tls.pubKey}" }
+                                        throw e
+                                    }
+
+                                    if (serverKey == pinnedKey) {
+                                        logger.info { "successfully checked server's certificate against pinned pubkey" }
+                                    } else {
+                                        logger.warning { "server's certificate does not match pinned pubkey, fallback to default check" }
+                                        defaultX509TrustManager.checkServerTrusted(chain, authType)
```
dpad85, line 142: @robbiehanson Not sure what we should do in that case. We could reject the connection attempt and dispatch an event containing the new pubkey of the server, and then Phoenix could display a message telling the user that the pubkey changed, and asking whether the app should trust this new key or keep the connection closed. Is that what you had in mind?

src/jvmMain/kotlin/fr/acinq/lightning/io/JvmTcpSocket.kt
```diff
@@ -58,34 +64,91 @@ class JvmTcpSocket(val socket: Socket) : TcpSocket {
         socket.close()
     }
 
+
+    companion object {
+        private val logger by lightningLogger<JvmTcpSocket>()
+        fun buildPublicKey(encodedKey: ByteArray): java.security.PublicKey {
+            val spec = X509EncodedKeySpec(encodedKey)
+            val algorithms = listOf("RSA", "EC")
```
dpad85, line 72: @sstone we only support `RSA` and `EC` to match what's done in Swift (see https://github.com/ACINQ/lightning-kmp/blob/9b019b16f7e6c3fab69a1ce122b2d0061e842d5e/PhoenixCrypto/PhoenixCrypto/Classes/NativeSocket.swift#L180-L191).

We could add more though (see https://docs.oracle.com/en/java/javase/11/docs/specs/security/standard-names.html#keyfactory-algorithms)

src/jvmMain/kotlin/fr/acinq/lightning/io/JvmTcpSocket.kt
robbiehanson: > We could reject the connection attempt

I think that's the safest option. Because otherwise a malicious actor could simply redirect to a server with a valid certificate.

> and dispatch an event [...], and then Phoenix could display a message

That's a great idea. As of now, the iOS code would simply stop connecting to Electrum. This idea is much better.

src/jvmMain/kotlin/fr/acinq/lightning/io/JvmTcpSocket.kt
robbiehanson: On iOS, the documentation is quite sparse. For the `kSecAttrKeyClass` value, those are the only 3 options I could find. It might be `kSecAttrKeyTypeEC` on iOS is equivalent to `[EC, XDH, X25519, X448]` on Android. :man_shrugging: I have no clue. But I think it's safe if you add all of them.

robbiehanson:
Rebased on master (to fix conflicts), and force-pushed.

robbiehanson:
Changes in commit [357b415](https://github.com/ACINQ/lightning-kmp/pull/327/commits/357b41536936c55961ebc2b1b82d3f43dfb028d6):

When a connection was closed, we previously didn't tell the client **why** the connection was closed. We only exposed the state:
```kotlin
enum class Connection { CLOSED, ESTABLISHING, ESTABLISHED }
```

We now expose the reason for the closure:
```kotlin
sealed class Connection {
    data class CLOSED(val reason: TcpSocket.IOException?): Connection()
    object ESTABLISHING: Connection()
    object ESTABLISHED: Connection()
}
```

This allows clients to react to certain situations. For example:
- if client has custom electrum server configured (with pinned public key)
- and electrum connection starts failing with "bad certificate" errors
- then the UI can display an error message with helpful information

src/jvmMain/kotlin/fr/acinq/lightning/io/JvmTcpSocket.kt
dpad85: Done with 357b41536936c55961ebc2b1b82d3f43dfb028d6 and f42bf230c242d639305c19bb03758b948f65e2c0

On the JVM, we throw a `TcpSocket.IOException.ConnectionClosed` exception where `cause` will be a `java.security.cert.CertificateException`.

src/jvmMain/kotlin/fr/acinq/lightning/io/JvmTcpSocket.kt
dpad85: Done in b5eebc2dacdd4ff37ca13b46c57693adccd72322

dpad85 (commented):
LGTM

A small issue is that the `reason` in `CLOSED(val reason: TcpSocket.IOException?)` is not the same between iOS and Android when the connection is closed because the server's key does not match the pinned key. On iOS we want to wrap the actual error and use a `NativeSocketException` specific to iOS. On Android we can't do that, and instead simply set the reason as a generic `CertificateException`. So on Phoenix's side we won't be able to handle that case in the `phoenix-shared` module. @robbiehanson Let me know if that's not correct!

robbiehanson:
> the reason [...] is not the same between iOS and Android when the connection is closed because the server's key does not match the pinned key

That's correct. These errors are specific to their respective networking stacks. And that's not even the end of the story.

It's expected that the exact error messages may differ by OS version too. For example, the iOS networking stack has undergone huge changes in the last few years, as they've transitioned from BSD sockets to NSURLSession to NetworkFramework. So I anticipate OS-version-specific checks will be required in the future. That's why I'm comfortable moving this logic out of the shared library and into the platform specific code.

dpad85 (approved):
🚀

robbiehanson:
Rebased on master

PR #330, opened 2022-03-30
Update trampoline feature bit

t-bast:
The `option_scid_alias` feature (https://github.com/lightning/bolts/pull/910) is going to clash with the feature bit we use for trampoline payments.

We are only using this bit to signal trampoline support in invoices, so we now set two bits for trampoline in invoices (the new one and the old one for backwards-compatibility).

We support reading old invoices that only set the previous bit to allow for a transition period. We also want to keep writing both feature bits to our invoices to allow older wallets to pay our invoices using trampoline. @dpad85 that means that Phoenix will need to activate both features in its configuration.

@SomberNight @ecdsa this change may impact Electrum as well, which will probably need to implement the same kind of transition


ecdsa:
Thanks for the update. what is the progress on merging the trampoline proposal?

t-bast:
> what is the progress on merging the trampoline proposal?

I've had some signalling from the c-lightning team that they would look into it once they'd successfully recruited a new team member, and I've got a few questions in private, but I'm still waiting for reviews on the spec PR...

dpad85 (approved):

ecdsa:
I suppose that we can use the same temporary bits as Eclair (148/149) this time. The reason why we did not use the same bits as Eclair last time (50/51) was to prevent ACINQ attempting pay-to-open with Electrum clients.

PR #333, opened 2022-04-15
Store closing txs as parts in outgoing payments db

dpad85:
When closing a channel, we used to aggregate the ids of the closing transactions and their amount, and store them in the status field for on-chain payments. With this PR, we instead store each of those closing transaction, their amount and their closing type as parts in the outgoing payments database.

- Status for on-chain payments do not contain any field related to closing channels anymore
```kotlin
// status for outgoing payments
sealed class Status {
    ...
        data class OffChain...
    
        data class OnChain(
            override val completedAt: Long = currentTimestampMillis()
        ) : Succeeded()
```

- Lightning parts now extend a generic `Part` sealed class:
```kotlin
data class LightningPart(
    override val id: UUID,
    val amount: MilliSatoshi,
    val route: List<HopDesc>,
    val status: Status,
    override val createdAt: Long = currentTimestampMillis()
) : Part() {
```

- A new Part for closing transactions has been added:
```kotlin
data class ClosingTxPart(
    override val id: UUID,
    val txId: ByteVector32,
    val claimed: Satoshi,
    val closingType: ChannelClosingType,
    override val createdAt: Long,
) : Part()
```

Also, some methods have been renamed to make the code a bit more readable:
- `completeOutgoingPayment` => `completeOutgoingPaymentOnchain`, `completeOutgoingPaymentOffchain`.
- `addOutgoingPart` => `completeOutgoingLightningPart`

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
@@ -8,12 +8,10 @@ import fr.acinq.lightning.channel.Helpers.Closing.onChainOutgoingHtlcs
 import fr.acinq.lightning.channel.Helpers.Closing.overriddenOutgoingHtlcs
 import fr.acinq.lightning.channel.Helpers.Closing.timedOutHtlcs
 import fr.acinq.lightning.db.ChannelClosingType
+import fr.acinq.lightning.db.OutgoingPayment
 import fr.acinq.lightning.transactions.Transactions
 import fr.acinq.lightning.transactions.Transactions.TransactionWithInputInfo.ClosingTx
-import fr.acinq.lightning.utils.Either
-import fr.acinq.lightning.utils.getValue
-import fr.acinq.lightning.utils.sat
-import fr.acinq.lightning.utils.sum
+import fr.acinq.lightning.utils.*
 import fr.acinq.lightning.wire.ChannelReestablish
 import fr.acinq.lightning.wire.Error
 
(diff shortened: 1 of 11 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
```diff
@@ -406,60 +404,60 @@ data class Closing(
...
+        fun extractClosingTx(
+            confirmedTxids: Set<ByteVector32>,
+            txs: List<Transaction>,
+            closingType: ChannelClosingType
+        ): List<OutgoingPayment.ClosingTxPart> {
+            return txs.filter { confirmedTxids.contains(it.txid) }.map {
+                OutgoingPayment.ClosingTxPart(
+                    id = UUID.randomUUID(),
+                    txId = it.txid,
+                    claimed = it.txOut.first().amount,
```
dpad85, line 426: @sstone Is it correct to only get the first output here? This is what was done previously so that part has not changed, but I'd like to be sure we are storing the correct amount.

robbiehanson (approved):
Looks good

src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
```diff
@@ -406,60 +404,60 @@ data class Closing(
...
+     * @param additionalConfirmedTx additional confirmed transaction for mutual close; we need this for the
+     *      mutual close scenario because we don't store the closing tx in the channel state.
+     */
+    private fun storeChannelClosed(additionalConfirmedTx: Transaction): ChannelAction.Storage.StoreChannelClosed {
+
+        /** Helper method to extract a [OutgoingPayment.ClosingTxPart] from a list of transactions, given a set of confirmed tx ids. */
+        fun extractClosingTx(
+            confirmedTxids: Set<ByteVector32>,
+            txs: List<Transaction>,
+            closingType: ChannelClosingType
```
sstone, line 420: LGTM, there are no functional changes and it is easier to read. 
One nit though: why not check that `closingType !=  ChannelClosingType.Mutual` since that would be the one instance where this method cannot be used ?
Another approach would be to add this method to `{Local/Remote/Revoked}CommitPublished`, not sure it would be better though

sstone, line 420: LGTM, there are no functional changes and it is easier to read. 
One nit though: why not check that `closingType !=  ChannelClosingType.Mutual` since that would be the one instance where this method cannot be used ?
Another approach would be to add this method to `{Local/Remote/Revoked}CommitPublished`, not sure it would be better though

robbiehanson (approved):
I re-ran the previous tests and everything looks good. The one bug I previously found has been fixed.

PR #335, opened 2022-04-21
Split Channel FSM into multiple files

sstone:
We use a language feature introduced with Kotlin 1.5 to split the channel state machine into multiple files (one per state) which makes it imho much easier to work with.
There are no functional changes.


sstone:
I was trying to split Channel.kt into multiple files while preserving git history for each new file, it does not seem to have worked I'll try again (but the end result should be the same).

dpad85 (approved):

t-bast (dismissed):
:+1:

sstone:
I've redone the splitting/cleaning of Channel.kt to preserve git history/blame info in the new channel states.

t-bast (approved):
Interesting approach, I didn't know that trick!

dpad85 (approved):

PR #338, opened 2022-04-28
Fixed rare purger test failure

remyers:
This was hard to reproduce until I reordered the `createInvoice` calls to be in the reverse of their timestamp order:

1. **unexpiredInvoice** with the default `timestampSeconds=currentTimestampSeconds`
2. **paidInvoice** with `timestampSeconds=100`
3. **expiredInvoice** with `timestampSeconds=1`) . 

This reordering caused an error because `listIncomingPayments` returns payments based on their time stamp, ordered as newest first. Before the fix to `createInvoice`, payments were listed in the order: expiredInvoice, paidInvoice, unexpiredInvoice instead of using the `timestampSeconds` parameter of `createInvoice` to produce the order: unexpiredInvoice, paidInvoice, expiredInvoice.

Now `createInvoice` properly passes the `timestampSeconds` parameter to the `addIncomingPayment` call instead of using the default `currentTimestampSeconds` every time.

I don't think this would have caused any problems in normal operation, but should now fix this test from failing on fast systems were all three invoices were marked as created with the same timestamp.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
@@ -96,7 +96,7 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
             timestampSeconds
         )
         logger.info { "h:$paymentHash generated payment request ${pr.write()}" }
-        db.addIncomingPayment(paymentPreimage, IncomingPayment.Origin.Invoice(pr))
+        db.addIncomingPayment(paymentPreimage, IncomingPayment.Origin.Invoice(pr), timestampSeconds)
         return pr
     }
 
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
@@ -96,7 +96,7 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
             timestampSeconds
         )
         logger.info { "h:$paymentHash generated payment request ${pr.write()}" }
-        db.addIncomingPayment(paymentPreimage, IncomingPayment.Origin.Invoice(pr), timestampSeconds)
+        db.addIncomingPayment(paymentPreimage, IncomingPayment.Origin.Invoice(pr))
         return pr
     }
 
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -96,7 +96,7 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
             timestampSeconds
         )
         logger.info { "h:$paymentHash generated payment request ${pr.write()}" }
-        db.addIncomingPayment(paymentPreimage, IncomingPayment.Origin.Invoice(pr))
+        db.addIncomingPayment(paymentPreimage, IncomingPayment.Origin.Invoice(pr), timestampSeconds)
```
sstone, line 99: Nice catch! It don't see why we would not want to use the  `timestampSeconds` parameters here, @dpad85 do you agree ?

sstone (approved):

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
dpad85: The `createdAt` field must be in millis, not seconds. 

Also we could simply add a `delay(100)` in-between the `createInvoice` calls in the test, to make sure the ordering is always the same. I think it's a better solution than modifying the payment handler logic, where we would now have a special rule for incoming payments using LN invoices where the invoice's timestamp somehow matters. The `createdAt` timestamp is not very useful in the end, so it's not really worth creating this rule.

remyers: Gotcha. I'll modify the test instead to ensure the ordering is correct. I see now why it makes sense that the expected behavior is for `addIncomingPayment` to use the default current time and not the `createAt` of the invoice.

remyers: fixed in [9fff056](https://github.com/ACINQ/lightning-kmp/pull/338/commits/9fff0563b30e009f255e4560c9e07be37d18a865)

dpad85: Thanks!

dpad85 (approved):
🏄

PR #339, opened 2022-05-02
 Split Channel FSM into multiple files

sstone:
We use a language feature introduced with Kotlin 1.5 to split the channel state machine into multiple files (one per state) which makes it imho much easier to work with.
There are no functional changes.
Git history/blame info has been preserved but for this to work we cannot squash-merge this PR otherwise we'll lose it again.


dpad85 (dismissed):

t-bast (dismissed):

dpad85 (approved):
The git history is lost in some classes but it is still better than losing everything.

PR #342, opened 2022-06-03
Add support for swap-out payments

dpad85:
This PR adds support for the custom swap-out Lightning messages that let us request a swap-out payment to the peer, and dispatch the response on the peer listener event stream.

### Swap-out details object

We change the `OutgoingPayment.Details.SwapOut` object that now contains a payment request parameter instead of a simple payment hash:

```kotlin
data class SwapOut(val address: String, val paymentRequest: PaymentRequest) : Details() {
    override val paymentHash: ByteVector32 = paymentRequest.paymentHash
}
```

This way, the `OutgoingPaymentHandler` is able to rebuild a `SendPayment` object (with a payment request) in `processPostRestartFailure` and `processPostRestartFulfill` even when the payment is a swap-out. This makes sense because a swap-out is actually a Lightning payment, but with some additional context.

### `SendPayment` event

As mentioned above, the `SendPayment` event object used to only work for `Normal` payments. It can now work for swap-outs :

```kotlin
interface SendPayment {
    val paymentRequest: PaymentRequest
    // other fields...
}

data class SendPaymentNormal(
    override val details: OutgoingPayment.Details.Normal,
    // other fields...
) : PaymentEvent(), SendPayment

data class SendPaymentSwapOut(
    override val details: OutgoingPayment.Details.SwapOut
    // other fields...
) : PaymentEvent(), SendPayment
```
This lets us support swap-outs with minimal changes to the payment handler.
See this commit for details: 90abfb1f27e7c5106eed39912c385ec521976c86

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
@@ -250,8 +250,10 @@ data class OutgoingPayment(
             override val paymentHash: ByteVector32 = Crypto.sha256(preimage).toByteVector32()
         }
 
-        /** Swaps out send a lightning payment to a swap server, which will send an on-chain transaction to a given address. */
-        data class SwapOut(val address: String, override val paymentHash: ByteVector32) : Details()
+        /** Swap-outs send a lightning payment to a swap server, which will send an on-chain transaction to a given address. */
+        data class SwapOut(val address: String, val paymentRequest: PaymentRequest) : Details() {
+            override val paymentHash: ByteVector32 = paymentRequest.paymentHash
+        }
 
         /** Corresponds to the on-chain payments made when closing a channel. */
         data class ChannelClosing(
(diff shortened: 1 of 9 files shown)


dpad85:
Note, this PR is built upon https://github.com/ACINQ/lightning-kmp/pull/333 and will be rebased for clarity once that PR is merged.

robbiehanson:
I've reviewed the code and everything looks good to me. But then I tried to test it using the corresponding `swap-out` on phoenix, and I ran into some issues. When trying to perform a swap out on testnet I consistently get an error:

> [OutgoingPaymentHandler] h:e008ac4e21c66f265b68b53dbfb1a254520f26ec870d464c808d45736e987074 p:115d869a-b43f-44cf-9587-def7f512927f i:7cceaa24-ff19-4b1a-a956-dd9ef3b1e688 HTLC failed: incorrect payment details or unknown payment hash

Note: the phoenix commit used was d7c5d462, and not the latest (74d7fa27), because I encountered compiler errors on the latest:

> Unresolved reference: toBigDecimal
> Unresolved reference: toBigDecimalOrNull

dpad85:
> I've reviewed the code and everything looks good to me. But then I tried to test it using the corresponding `swap-out` on phoenix, and I ran into some issues. When trying to perform a swap out on testnet I consistently get an error:

The actual error (amount received too small) is obfuscated by the receiving node. This is actually an issue on phoenix's side, see my comment here: https://github.com/ACINQ/phoenix/pull/286#issuecomment-1161972563

PR #347, opened 2022-08-17
StartTLS

robbiehanson:
This is a rebase of PR #190, which is now over 18 months old. So there were many merge issues with it.

Note that many of the changes from #190 were implemented via #196. The remaining changes were strictly related to StartTLS support. So that's what we see in this PR.

@dpad85 I took JvmTcpSocket from your `feature/socket-starttls-merge` branch. But I made a small change to function `receiveAvailable`

```kotlin
// Old:
connection.input.readAvailable(buffer, 0, maxLength)
// New:
connection.input.readAvailable(buffer, offset, length)
```

We forgot to take the `offset` parameter into account. And the compiler complained about the `maxLength` parameter name being different from the `length` parameter name in TcpSocket's interface declaration.

Also, in the original PR we were forced to decrease security during TLS verification for StartTLS. This is no longer necessary as I found a workaround for the problem. (However I'm not sure if this would have affected its use strictly for Tor...)

One other interesting change in this PR, not directly related to StartTLS:

I discovered that Apple's NWConnection doesn't "fail" if a standard TLS handshake fails. Instead of transitioning to a `NWConnection.state.failed` (as one would expect), it instead transitions to `NWConnection.state.waiting(err)` with a TLS error. This was unexpected, and would result in the connection never failing. (Or perhaps only after a 2 minute timeout.)

### DIFF
--- a/PhoenixCrypto/PhoenixCrypto/Classes/NativeSocket.swift
+++ b/PhoenixCrypto/PhoenixCrypto/Classes/NativeSocket.swift
@@ -19,10 +19,14 @@ enum DecodePubKeyError: Error {
 @objc
 public class NativeSocket: NSObject {
 	
+	private let host: String
 	private let connection: NWConnection
+	private let parentConnections: [NWConnection]
 	
-	private init(connection: NWConnection) {
+	private init(host: String, connection: NWConnection, parentConnections: [NWConnection] = []) {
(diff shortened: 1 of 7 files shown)


dpad85 (approved):
LGTM

dpad85 (approved):

PR #348, opened 2022-08-23
Remove `BroadcastChannel` and cleanup annotations

pm47:
No more `ObsoleteCoroutinesApi`!

Best reviewed commit by commit.

PR #350, opened 2022-08-24
Add a simple on-chain wallet

pm47:
Usage:
```kotlin
val client = ElectrumClient(...)
val wallet = ElectrumMiniWallet(Block.LivenetGenesisBlock.hash, client, this)

// watch-only
wallet.addAddress("16MmJT8VqW465GEyckWae547jKVfMB14P8")
// spendable p2wpkh
wallet.addPay2wpkhAddress(randomKey())

wallet.walletStateFlow
  .collect { println("balance=${it.balance} utxos=${it.utxos.size}") }
```

Example to watch funds on the final addresses (@dpad85):
```kotlin
val nodeParams: NodeParams = ...
val wallet: ElectrumMiniWallet = ...
val channels: List<ChannelStateWithCommitments> = ...

channels.forEach { data ->
  val bitcoinAddress = Bitcoin.addressFromPublicKeyScript(nodeParams.chainHash, data.commitments.localParams.defaultFinalScriptPubKey.toByteArray())!!
  wallet.addAddress(bitcoinAddress)
}
```

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
@@ -1,11 +1,9 @@
 package fr.acinq.lightning.blockchain.electrum
 
-import fr.acinq.bitcoin.Bitcoin
-import fr.acinq.bitcoin.ByteVector
-import fr.acinq.bitcoin.ByteVector32
-import fr.acinq.bitcoin.Script
+import fr.acinq.bitcoin.*
 import fr.acinq.lightning.utils.Connection
 import fr.acinq.lightning.utils.lightningLogger
+import fr.acinq.lightning.utils.sat
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
@@ -13,7 +11,9 @@ import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.launch
 import kotlin.native.concurrent.ThreadLocal
 
-data class Balance(val utxos: List<UnspentItem>)
+data class WalletState(val utxos: List<UnspentItem>) {
+    val balance: Satoshi = utxos.map { it.value }.sum().sat
+}
 
 /**
  * A very simple wallet that only watches one address and publishes its utxos.
@@ -26,8 +26,8 @@ class ElectrumMiniWallet(
 ) : CoroutineScope by scope {
 
     // state flow with the current balance
-    private val _balanceFlow = MutableStateFlow<Balance?>(null)
-    val balanceFlow get() = _balanceFlow.asStateFlow()
+    private val _walletStateFlow = MutableStateFlow<WalletState?>(null)
+    val walletStateFlow get() = _walletStateFlow.asStateFlow()
 
     private val pubkeyScript = ByteVector(Script.write(Bitcoin.addressToPublicKeyScript(chainHash, bitcoinAddress)))
     private val scriptHash = ElectrumClient.computeScriptHash(pubkeyScript)
@@ -60,12 +60,11 @@ class ElectrumMiniWallet(
                     }
                     is ScriptHashListUnspentResponse -> {
                         if (it.scriptHash == scriptHash) {
-                            logger.info { "${it.unspents.size} utxo(s) for address=$bitcoinAddress" }
-                            it.unspents.forEach {
-                                logger.info { "utxo=${it.outPoint.txid}:${it.outPoint.index} amount=${it.value} sat" }
-                            }
+                            val walletState = WalletState(it.unspents)
+                            logger.info { "${it.unspents.size} utxo(s) for address=$bitcoinAddress, balance=${walletState.balance}" }
+                            it.unspents.forEach { logger.debug { "utxo=${it.outPoint.txid}:${it.outPoint.index} amount=${it.value} sat" } }
                             // publish the updated balance
-                            _balanceFlow.value = Balance(it.unspents)
+                            _walletStateFlow.value = walletState
                         }
                     }
                     else -> {}
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
@@ -10,41 +10,48 @@ import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.launch
 import kotlin.native.concurrent.ThreadLocal
 
-data class WalletState(val addresses: Map<String, Pair<List<UnspentItem>, PrivateKey?>>) {
-    val utxos: List<UnspentItem> = addresses.flatMap { it.value.first }
+data class WalletState(val addresses: Map<String, List<UnspentItem>>, val privateKeys: Map<String, PrivateKey>) {
+    val utxos: List<UnspentItem> = addresses.flatMap { it.value }
     val balance: Satoshi = utxos.sumOf { it.value }.sat
 
     fun spendable(): List<UnspentItem> {
         return addresses
-            .filter { it.value.second != null }
-            .flatMap { it.value.first }
+            .filter { privateKeys.containsKey(it.key) }
+            .flatMap { it.value }
     }
+
+    private val outPoint2Address = addresses.entries.flatMap { entry -> entry.value.map { it.outPoint to entry.key } }.toMap()
 
     /** Sign the inputs owned by this wallet (only works for P2WPKH scripts) */
     fun sign(tx: Transaction): Transaction {
         return tx.txIn.foldIndexed(tx) { index, wipTx, txIn ->
-            when (val utxo = spendable().find { it.outPoint == txIn.outPoint }) {
-                is UnspentItem -> {
-                    // we know this utxo and have the private key
-                    val privateKey = addresses.values.find { it.first.contains(utxo) }!!.second!! // TODO improve that
-                    // mind this: the pubkey script used for signing is not the prevout pubscript (which is just a push
-                    // of the pubkey hash), but the actual script that is evaluated by the script engine, in this case a PAY2PKH script
-                    val publicKey = privateKey.publicKey()
-                    val pubKeyScript = Script.pay2pkh(publicKey)
-                    val sig = Transaction.signInput(
-                        tx,
-                        index,
-                        pubKeyScript,
-                        SigHash.SIGHASH_ALL,
-                        utxo.value.sat,
-                        SigVersion.SIGVERSION_WITNESS_V0,
-                        privateKey
-                    )
-                    // update the signature for the corresponding input
-                    val witness = ScriptWitness(listOf(sig.byteVector(), publicKey.value))
-                    wipTx.updateWitness(index, witness)
+            val witness = outPoint2Address[txIn.outPoint]?.let { address ->
+                addresses[address]?.find { it.outPoint == txIn.outPoint }?.let { utxo ->
+                    privateKeys[address]?.let { privateKey ->
+                        // mind this: the pubkey script used for signing is not the prevout pubscript (which is just a push
+                        // of the pubkey hash), but the actual script that is evaluated by the script engine, in this case a PAY2PKH script
+                        val publicKey = privateKey.publicKey()
+                        val pubKeyScript = Script.pay2pkh(publicKey)
+                        val sig = Transaction.signInput(
+                            tx,
+                            index,
+                            pubKeyScript,
+                            SigHash.SIGHASH_ALL,
+                            utxo.value.sat,
+                            SigVersion.SIGVERSION_WITNESS_V0,
+                            privateKey
+                        )
+                        ScriptWitness(listOf(sig.byteVector(), publicKey.value))
+                    }
                 }
-                else -> wipTx // we don't know how to sign this input
+            }
+            when (witness) {
+                is ScriptWitness ->
+                    // update the signature for the corresponding input
+                    wipTx.updateWitness(index, witness)
+                else ->
+                    // we don't know how to sign this input
+                    wipTx
             }
         }
     }
@@ -55,7 +62,6 @@ private sealed interface WalletCommand {
         object ElectrumConnected : WalletCommand
         data class ElectrumNotification(val msg: ElectrumResponse) : WalletCommand
         data class AddAddress(val bitcoinAddress: String, val privateKey: PrivateKey?) : WalletCommand
-        data class AddPubkeyScript(val pubkeyScript: ByteVector, val privateKey: PrivateKey?) : WalletCommand
     }
 }
 
@@ -70,7 +76,7 @@ class ElectrumMiniWallet(
 ) : CoroutineScope by scope {
 
     // state flow with the current balance
-    private val _walletStateFlow = MutableStateFlow(WalletState(emptyMap()))
+    private val _walletStateFlow = MutableStateFlow(WalletState(emptyMap(), emptyMap()))
     val walletStateFlow get() = _walletStateFlow.asStateFlow()
 
     // all currently watched script hashes and their corresponding bitcoin address
@@ -82,12 +88,6 @@ class ElectrumMiniWallet(
     fun addAddress(bitcoinAddress: String, privateKey: PrivateKey? = null) {
         launch {
             mailbox.send(WalletCommand.Companion.AddAddress(bitcoinAddress, privateKey))
-        }
-    }
-
-    fun addPubkeyScript(pubkeyScript: ByteVector, privateKey: PrivateKey? = null) {
-        launch {
-            mailbox.send(WalletCommand.Companion.AddPubkeyScript(pubkeyScript, privateKey))
         }
     }
 
@@ -125,10 +125,7 @@ class ElectrumMiniWallet(
                             }
                             is ScriptHashListUnspentResponse -> {
                                 scriptHashes[msg.scriptHash]?.let { address ->
-                                    val newValue = _walletStateFlow.value.addresses
-                                        .getOrElse(address, defaultValue = { emptyList<UnspentItem>() to null }) // default value
-                                        .let { v -> msg.unspents to v.second } // update the utxos, keep the private key
-                                    val walletState = WalletState(_walletStateFlow.value.addresses + (address to newValue))
+                                    val walletState = _walletStateFlow.value.copy(addresses = _walletStateFlow.value.addresses + (address to msg.unspents))
                                     logger.info { "${msg.unspents.size} utxo(s) for address=$address balance=${walletState.balance}" }
                                     msg.unspents.forEach { logger.debug { "utxo=${it.outPoint.txid}:${it.outPoint.index} amount=${it.value} sat" } }
                                     // publish the updated balance
@@ -141,10 +138,9 @@ class ElectrumMiniWallet(
                     is WalletCommand.Companion.AddAddress -> {
                         logger.info { "adding new address=${it.bitcoinAddress}" }
                         scriptHashes = scriptHashes + subscribe(it.bitcoinAddress)
-                    }
-                    is WalletCommand.Companion.AddPubkeyScript -> {
-                        logger.info { "adding new pubkeyScript=${it.pubkeyScript}" }
-                        scriptHashes = scriptHashes + subscribe(it.pubkeyScript)
+                        it.privateKey?.let { privateKey ->
+                            _walletStateFlow.value = _walletStateFlow.value.copy(privateKeys = _walletStateFlow.value.privateKeys + (it.bitcoinAddress to privateKey))
+                        }
                     }
                 }
             }
@@ -153,11 +149,6 @@ class ElectrumMiniWallet(
 
     private fun subscribe(bitcoinAddress: String): Pair<ByteVector32, String> {
         val pubkeyScript = ByteVector(Script.write(Bitcoin.addressToPublicKeyScript(chainHash, bitcoinAddress)))
-        return subscribe(pubkeyScript)
-    }
-
-    private fun subscribe(pubkeyScript: ByteVector): Pair<ByteVector32, String> {
-        val bitcoinAddress = Bitcoin.addressFromPublicKeyScript(chainHash, pubkeyScript.toByteArray())!!
         val scriptHash = ElectrumClient.computeScriptHash(pubkeyScript)
         logger.info { "subscribing to address=$bitcoinAddress pubkeyScript=$pubkeyScript scriptHash=$scriptHash" }
         client.sendElectrumRequest(ScriptHashSubscription(scriptHash))
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
@@ -58,6 +58,12 @@ class ElectrumMiniWallet(
         }
     }
 
+    fun addPubkeyScript(pubkeyScript: ByteVector, privateKey: PrivateKey? = null) {
+        launch {
+            mailbox.send(WalletCommand.Companion.AddPubkeyScript(pubkeyScript, privateKey))
+        }
+    }
+
     init {
         launch {
             // listen to connection events
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
@@ -23,8 +23,8 @@ data class WalletState(val addresses: Map<String, List<UnspentItem>>, val privat
         return addresses
             .filter { privateKeys.containsKey(it.key) }
             .flatMap { it.value }
-            .filter { parentTxs.containsKey(it.tx_hash) }
-            .map { Utxo(parentTxs[it.tx_hash]!!, it.tx_pos, it.height) }
+            .filter { parentTxs.containsKey(it.txid) }
+            .map { Utxo(parentTxs[it.txid]!!, it.outputIndex, it.blockHeight) }
     }
 
     private val outPoint2Address = addresses.entries.flatMap { entry -> entry.value.map { it.outPoint to entry.key } }.toMap()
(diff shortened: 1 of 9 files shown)


src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -0,0 +1,119 @@
...
+import fr.acinq.lightning.utils.Connection
+import fr.acinq.lightning.utils.lightningLogger
+import fr.acinq.lightning.utils.sat
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.launch
+import kotlin.native.concurrent.ThreadLocal
+
+data class WalletState(val addresses: Map<String, List<UnspentItem>>) {
```
pm47, line 13: The hierarchy is useful to to update the utxos (Electrum will return a set of utxos per script).

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -0,0 +1,119 @@
...
+    private val client: ElectrumClient,
+    private val scope: CoroutineScope
+) : CoroutineScope by scope {
+
+    // state flow with the current balance
+    private val _walletStateFlow = MutableStateFlow(WalletState(emptyMap()))
+    val walletStateFlow get() = _walletStateFlow.asStateFlow()
+
+    // all currently watched bitcoin addresses and their corresponding script hash
+    private var scriptHashes: Map<ByteVector32, String> = emptyMap()
```
pm47, line 41: Should we track addresses or pubkeyscripts internally? The former is more human friendly, the latter is more generic.

pm47:
Rebased on top of #348.

pm47:
Rebased on master.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -0,0 +1,139 @@
...
+    val utxos: List<UnspentItem> = pubKeyScripts.flatMap { it.value.first }
+    val balance: Satoshi = utxos.sumOf { it.value }.sat
+    val spendable: List<Triple<ByteVector, PrivateKey, UnspentItem>> by lazy {
+        pubKeyScripts
+            .entries
+            .filter { it.value.second != null }
+            .flatMap {
+                it.value.first.map { utxo -> Triple(it.key, it.value.second!!, utxo) }
+            }
+    }
```
pm47, line 23: @t-bast is this helper useful?

t-bast: Why do we need to have nullable private keys in the first place?
As a consumer of this, I don't want to be exposed to those, I'm only interested in utxos I can spend.
The ideal interface I'd like is:

1. A function that gives me a set of utxos I can spend (with an optional filter on a given address, if I only want the funds I swapped-in to a specific address)
2. A function that signs a transaction containing utxos I got from the first step (iterate through the transaction's inputs and sign all the inputs that belong to the wallet)

Does that work with the current model? In this model I won't need access to the private keys at all, so this should stay an internal implementation detail.

pm47: > Why do we need to have nullable private keys in the first place?

Because it may also be used as a watch-only wallet, e.g. to monitor final addresses from channels.

> Does that work with the current model?

Not well, because the wallet is as generic as possible, in order to also cover the case above. That's also why I moved from tracking addresses to scriptpubkeys. Problem is, being to generic prevents me from doing methods like the one you are asking for. 

Maybe I chose a wrong approach and should instead just build a `p2wpkh` wallet.

t-bast: > Maybe I chose a wrong approach and should instead just build a p2wpkh wallet.

I think it would make sense to start with only support for `p2wpkh` indeed, we're always going to generate our receiving addresses so we fully choose the script subset we want, don't we? We can then later add support for `p2tr` without any script, and maybe at some point for other types of script, but it's really not clear whether we'll need it.

pm47: Agreed, I'll try that.

pm47: Added your requested 2. in 25a6c32, didn't do 1. but it's should be straightforward.

I feel like next step is to wait until we actually use it and improve then.

pm47:
Since this is still unused code, I think we should merge to reduce rebase issues in wip branches.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -0,0 +1,162 @@
...
+import fr.acinq.lightning.utils.Connection
+import fr.acinq.lightning.utils.lightningLogger
+import fr.acinq.lightning.utils.sat
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.launch
+import kotlin.native.concurrent.ThreadLocal
+
+data class WalletState(val addresses: Map<String, List<UnspentItem>>, val privateKeys: Map<String, PrivateKey>) {
```
t-bast, line 13: We can't really work with `UnspentItem` for dual funding, we need the whole previous transaction...
Can't we define the following class and add more requests to Electrum to get this information?

```kotlin
data class Utxo(val previousTx: Transaction, val outputIndex: Int, val height: Long) {
    val outPoint = OutPoint(previousTx, outputIndex.toLong())
    val amount = previousTx.txOut[outputIndex].amount
}
```

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -0,0 +1,162 @@
...
+                        val sig = Transaction.signInput(
+                            tx,
+                            index,
+                            pubKeyScript,
+                            SigHash.SIGHASH_ALL,
+                            utxo.value.sat,
+                            SigVersion.SIGVERSION_WITNESS_V0,
+                            privateKey
+                        )
+                        ScriptWitness(listOf(sig.byteVector(), publicKey.value))
```
t-bast, line 44:
```suggestion
                        Script.witnessPay2wpkh(publicKey, sig.byteVector())
```

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -0,0 +1,162 @@
...
+    private val _walletStateFlow = MutableStateFlow(WalletState(emptyMap(), emptyMap()))
+    val walletStateFlow get() = _walletStateFlow.asStateFlow()
+
+    // all currently watched script hashes and their corresponding bitcoin address
+    private var scriptHashes: Map<ByteVector32, String> = emptyMap()
+
+    // the mailbox of this "actor"
+    private val mailbox: Channel<WalletCommand> = Channel(Channel.BUFFERED)
+
+    fun addAddress(bitcoinAddress: String, privateKey: PrivateKey? = null) {
```
t-bast, line 88: If we want to make sure that this is correctly used, I think we should split it into two distinct functions:

```
fun addPay2wpkhAddress(privateKey: PrivateKey)
fun addWatchOnlyAddress(bitcoinAddress: String)
```

In the first one we would compute the address corresponding to the p2wpkh of the given private key.
This way we're sure the privateKey is always consistent with the address.

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/WalletStateUnitTest.kt
```diff
@@ -0,0 +1,53 @@
...
+        assertEquals(address1, "tb1qtkl49wxh4a8mt7dhtwqg7z5zs3yn2vdn6qcz7p")
+        val scriptPubKey1 = ByteVector(Script.write(Script.pay2wpkh(pub1)))
+        assertEquals(scriptPubKey1, ByteVector.fromHex("00145dbf52b8d7af4fb5f9b75b808f0a8284493531b3"))
+
+
+        val walletState = WalletState(
+            addresses = mapOf(
+                address1 to listOf(utxo),
+                Bitcoin.computeP2WpkhAddress(randomKey().publicKey(), Block.TestnetGenesisBlock.hash) to emptyList<UnspentItem>(),
+                Bitcoin.computeP2WpkhAddress(randomKey().publicKey(), Block.TestnetGenesisBlock.hash) to emptyList<UnspentItem>()
```
t-bast, line 31: nit: the types are redundant:

```suggestion
                Bitcoin.computeP2WpkhAddress(randomKey().publicKey(), Block.TestnetGenesisBlock.hash) to emptyList(),
                Bitcoin.computeP2WpkhAddress(randomKey().publicKey(), Block.TestnetGenesisBlock.hash) to emptyList()
```

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/WalletStateUnitTest.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        val unsignedTx = Transaction(
+            version = 1,
+            txIn = listOf(TxIn(utxo.outPoint, sequence = 0xffffffffL)),
+            txOut = listOf(
+                TxOut(
+                    3700000L.toSatoshi(),
+                    scriptPubKey1
+                )
+            ), // we reuse the same output script but if could be anything else
```
t-bast, line 44: nit:
```suggestion
            ), // we reuse the same output script but it could be anything else
```

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/WalletStateUnitTest.kt
```diff
@@ -0,0 +1,53 @@
...
+import kotlin.test.assertEquals
+
+class WalletStateUnitTest : LightningTestSuite() {
+
+    @Test
+    fun `sign a p2wpkh tx`() {
+
+        // this tx pays 0.038 BTC to p2wpkh tb1qtkl49wxh4a8mt7dhtwqg7z5zs3yn2vdn6qcz7p
+        val parentTx =
+            Transaction.Companion.read("01000000000101c37055af18a2ab1714e648fe5140982ceec74b6ed9376292662465c9ec3f5bf20000000000ffffffff01c0fb3900000000001600145dbf52b8d7af4fb5f9b75b808f0a8284493531b302483045022100966d17538e775b45bc1a995ff21871e0ecd428895bfe685f9016b1ea0cc2ff810220277e8f64b3827b8c79db34cfaf06af5660d9c3ebcb3f8ae857cf44a138b3cf620121030533e1d2e9b7576fef26de1f34d67887158b7af1b040850aab6024b07925d70a00000000")
```
t-bast, line 16: nit:
```suggestion
            Transaction.read("01000000000101c37055af18a2ab1714e648fe5140982ceec74b6ed9376292662465c9ec3f5bf20000000000ffffffff01c0fb3900000000001600145dbf52b8d7af4fb5f9b75b808f0a8284493531b302483045022100966d17538e775b45bc1a995ff21871e0ecd428895bfe685f9016b1ea0cc2ff810220277e8f64b3827b8c79db34cfaf06af5660d9c3ebcb3f8ae857cf44a138b3cf620121030533e1d2e9b7576fef26de1f34d67887158b7af1b040850aab6024b07925d70a00000000")
```

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/WalletStateUnitTest.kt
```diff
@@ -0,0 +1,53 @@
...
+import fr.acinq.lightning.tests.utils.LightningTestSuite
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class WalletStateUnitTest : LightningTestSuite() {
+
+    @Test
+    fun `sign a p2wpkh tx`() {
+
+        // this tx pays 0.038 BTC to p2wpkh tb1qtkl49wxh4a8mt7dhtwqg7z5zs3yn2vdn6qcz7p
```
t-bast, line 14: Instead of this comment, it would be better to just verify it:

```kotlin
assertEquals(parentTx.txOut.first(), TxOut(3_800_000.sat, Script.pay2wpkh(pub1)))
```

t-bast (commented):
I'm wondering how we will actually use that flow-based model, let me know if that matches what you had in mind:

- at start-up, we create a singleton that registers to this wallet flow
- this singleton waits for wallet state, and once it receives some wallet state, it waits for an additional X seconds to get data that is as up-to-date as possible
- after that duration, it filters spendable utxos and checks which ones are already used in unconfirmed channels
- if there are remaining spendable utxos that aren't used in unconfirmed channels, it starts the channel funding flow with them

If we receive a new wallet state with new utxos after that point, I'm not sure what we should do: repeat that process again and start opening another channel?

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
t-bast: Also, what does `height` mean? I'm not sure whether it's the number of confirmations or the block height at which this utxo was created, it's probably worth renaming to clarify.

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/WalletStateUnitTest.kt
pm47: TIL!

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/WalletStateUnitTest.kt
pm47: Done in [6fcb944](https://github.com/ACINQ/lightning-kmp/pull/350/commits/6fcb944de8b901d3efbb7193f2dd8ec7c1cfe9a5).

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/WalletStateUnitTest.kt
pm47: Done in [6fcb944](https://github.com/ACINQ/lightning-kmp/pull/350/commits/6fcb944de8b901d3efbb7193f2dd8ec7c1cfe9a5).

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
pm47: Done in [6fcb944](https://github.com/ACINQ/lightning-kmp/pull/350/commits/6fcb944de8b901d3efbb7193f2dd8ec7c1cfe9a5). I kept the unsafe method and the API is generally sketchy, but I'd rather revisit when we know more how we are using it.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
pm47: I did the bare minimum in 6b8209b. Notably, parent txs are not cleaned up.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -0,0 +1,188 @@
...
+
+    internal fun addAddress(bitcoinAddress: String, privateKey: PrivateKey? = null) {
+        launch {
+            mailbox.send(WalletCommand.Companion.AddAddress(bitcoinAddress, privateKey))
+        }
+    }
+
+    fun addPay2wpkhAddress(privateKey: PrivateKey) {
+        launch {
+            val bitcoinAddress = Bitcoin.computeP2WpkhAddress(privateKey.publicKey(), Block.TestnetGenesisBlock.hash)
```
t-bast, line 103: We'll need to make the `chainHash` an argument of this function, right? Why not do it now?

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -0,0 +1,188 @@
...
+                                        logger.info { "non-empty status for address=$bitcoinAddress, requesting utxos" }
+                                        client.sendElectrumRequest(ScriptHashListUnspent(msg.scriptHash))
+                                    }
+                                }
+                            }
+                            is ScriptHashListUnspentResponse -> {
+                                scriptHashes[msg.scriptHash]?.let { address ->
+                                    val newUtxos = msg.unspents.minus((_walletStateFlow.value.addresses[address] ?: emptyList()).toSet())
+                                    // request new parent txs
+                                    newUtxos.forEach { utxo -> client.sendElectrumRequest(GetTransaction(utxo.tx_hash)) }
```
t-bast, line 148: How can that work? `GetTransaction` expects the `txid`, not the `tx_hash`?
Or is Electrum's `tx_hash` actually the `txid` (which would be very confusing...)?

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -0,0 +1,188 @@
...
+    data class Utxo(val previousTx: Transaction, val outputIndex: Int, val blockHeight: Long) {
+        val outPoint = OutPoint(previousTx, outputIndex.toLong())
+        val amount = previousTx.txOut[outputIndex].amount
+    }
+
+    fun spendable(): List<Utxo> {
+        return addresses
+            .filter { privateKeys.containsKey(it.key) }
+            .flatMap { it.value }
+            .filter { parentTxs.containsKey(it.tx_hash) }
```
t-bast, line 26: Sane here, isn't there a confusion between `txid` and `tx_hash`?

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/WalletStateUnitTest.kt
```diff
@@ -0,0 +1,55 @@
...
+import fr.acinq.lightning.tests.utils.LightningTestSuite
+import fr.acinq.lightning.utils.sat
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class WalletStateUnitTest : LightningTestSuite() {
+
+    @Test
+    fun `sign a p2wpkh tx`() {
+
```
t-bast, line 14: very nit: why the empty line?

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/WalletStateUnitTest.kt
```diff
@@ -0,0 +1,55 @@
...
+        val address1 = Bitcoin.computeP2WpkhAddress(pub1, Block.TestnetGenesisBlock.hash)
+        assertEquals(address1, "tb1qtkl49wxh4a8mt7dhtwqg7z5zs3yn2vdn6qcz7p")
+        val scriptPubKey1 = ByteVector(Script.write(Script.pay2wpkh(pub1)))
+        assertEquals(scriptPubKey1, ByteVector.fromHex("00145dbf52b8d7af4fb5f9b75b808f0a8284493531b3"))
+
+        // the parent tx pays 0.038 BTC to p2wpkh tb1qtkl49wxh4a8mt7dhtwqg7z5zs3yn2vdn6qcz7p
+        val parentTx =
+            Transaction.read("01000000000101c37055af18a2ab1714e648fe5140982ceec74b6ed9376292662465c9ec3f5bf20000000000ffffffff01c0fb3900000000001600145dbf52b8d7af4fb5f9b75b808f0a8284493531b302483045022100966d17538e775b45bc1a995ff21871e0ecd428895bfe685f9016b1ea0cc2ff810220277e8f64b3827b8c79db34cfaf06af5660d9c3ebcb3f8ae857cf44a138b3cf620121030533e1d2e9b7576fef26de1f34d67887158b7af1b040850aab6024b07925d70a00000000")
+        assertEquals(parentTx.txOut.first(), TxOut(3_800_000L.sat, scriptPubKey1))
+        val utxo = UnspentItem(parentTx.txid, 0, 3_800_000, 0)
```
t-bast, line 26: Here as well, confusion between `txid` and `tx_hash`

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWalletIntegrationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        wallet.addAddress("19b3QuFuYSSYPoLt1AxVQmG3LifgSSPNyA", randomKey())
+
+        val walletState = wallet.walletStateFlow
+            .filter { it.parentTxs.size == 11 }
+            .first()
+
+        assertEquals(4 + 6 + 1, walletState.utxos.size)
+        assertEquals(72_000_000.sat + 30_000_000.sat + 5_000_000.sat, walletState.balance)
+        assertEquals(7, walletState.spendable().size)
+        assertContains(
```
t-bast, line 85: We should probably verify that the previousTx's output correctly sends the right amount to the address we expect, this would ensure the `txid` / `tx_hash` confusion has been fixed. Overall this test should verify more things to ensure that we fetch and return the right data.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
t-bast: That's good enough, I don't think we care about cleaning up the parent txs.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
pm47: > Or is Electrum's tx_hash actually the txid (which would be very confusing...)?

Unfortunately yes...

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
pm47: Damn, fixed in 865eda8.

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/WalletStateUnitTest.kt
pm47: Removed in 462b73e.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
t-bast: I think we should make them private indeed, and we should have tests to very explicitly show that this is the `txid`...

pm47: WDYT? This is _the_ proper way:

```kotlin
data class UnspentItem(@SerialName("tx_hash") val txid: ByteVector32, @SerialName("tx_pos") val outputIndex: Int, val value: Long, @SerialName("height") val blockHeight: Long) {
    val outPoint by lazy { OutPoint(txid.reversed(), outputIndex.toLong()) }
}
```

t-bast: Looks great!

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/WalletStateUnitTest.kt
pm47: Fixed in 6a66e25.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
pm47: Fixed in 6a66e25.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
pm47: Had to do it differently since we're not actually using kotlin's json serializon. Fixed in 6a66e25.

t-bast: Right, that's even better!

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWalletIntegrationTest.kt
pm47: I tried to improve with fbb1eed but I'm not convinced. I think you would prefer (1) a "self-referential test" where e.g. we verify that the output of a given tx indeed corresponds to an address defined in the test, while instead I went with (2) a "test vector" which compares to expected result to data taken from the blockchain.

For this particular test I think (2) is better than (1), because even if (1) is more self-explanatory, it is more verbose and ultimately only verifies a subset of (2).

Tbh I think the best would be to just build a complete `WalletState` and just do an `assertEquals`.

t-bast: You're right, the test vector is probably good enough, I verified it by looking at the blockchain, looks good!

PR #351, opened 2022-08-26
Correctly set `scriptHashSubscriptions` when (re)connecting

pm47:
The field `WatcherRunning.scriptHashSubscriptions`, which keeps track of existing subscriptions in order to not put duplicate watchers, isn't set when transitioning from `WatcherDisconnected`->`WatcherRunning`.

Note: to reproduce the bug in tests I had to increase the visibility of a some classes.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
@@ -62,7 +62,7 @@ private data class BroadcastTxAction(val tx: Transaction) : WatcherAction()
  * +----------+         +-----+
  *
  */
-private sealed class WatcherState {
+internal sealed class WatcherState {
     abstract fun process(event: WatcherEvent): Pair<WatcherState, List<WatcherAction>>
 }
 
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
```diff
@@ -82,17 +82,18 @@ private data class WatcherDisconnected(
...
                             publishQueue.forEach { add(PublishAsapAction(it.tx)) }
                             getTxQueue.forEach { add(AskForTransaction(it.txid, it.channelId)) }
                         }
+                        state = WatcherRunning(
+                            height = message.height,
+                            tip = message.header,
+                            block2tx = block2tx,
+                            watches = watches,
+                            scriptHashSubscriptions = actions.filterIsInstance<RegisterToScriptHashNotification>().map { it.scriptHash }.toSet()
+                        )
```
pm47, line 96: Is it safe to change the order of `actions` and `state` within `newState`?

sstone: Yes it is (`newState` creates a `builder` initialized with `actions` and `states` and then calls its `build()` method)

sstone (approved):

PR #353, opened 2022-08-30
Add dual funding codecs

t-bast:
Add support for encoding/decoding `open_channel2` and `accept_channel2`. The test cases are imported from `eclair`, with a small change of channel type from `anchor_output_zero_fee_htlc_tx` to `anchor_outputs`.

We also add the serialization attributes to interactive-tx messages, as some of them will later need to be serialized in channel state.


pm47 (approved):
This is a quick review, but test coverage looks good and there will be cross-impl testing.

PR #355, opened 2022-08-31
Remove all static loggers

pm47:
Instead we create loggers from a new `LoggerFactory` attribute in `NodeParams`. This allows the separation of logging contexts in tests and hugely facilitates debugging.

NB: some parts are a bit tedious, future improvements in Kotlin will make them much more straightforward (see https://kotlinlang.org/docs/whatsnew1620.html#prototype-of-context-receivers-for-kotlin-jvm).

@dpad85  @robbiehanson The `setLightningLoggerFactory` static method has been removed, instead you need to pass a `LoggingFactory` to the `NodeParams` constructor.


src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClient.kt
```diff
@@ -53,11 +54,11 @@ internal data class BroadcastStatus(val connection: Connection) : ElectrumClient
  *
  */
 internal sealed class ClientState {
-    abstract fun process(event: ClientEvent): Pair<ClientState, List<ElectrumClientAction>>
+    abstract fun process(event: ClientEvent, logger: Logger): Pair<ClientState, List<ElectrumClientAction>>
```
pm47, line 57: This looks tedious but:
- it makes sense to use a logger context attached to the Machine and not the State of the FSM
- it prepares the use of [context receivers](https://kotlinlang.org/docs/whatsnew1620.html#prototype-of-context-receivers-for-kotlin-jvm)

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -94,10 +91,7 @@ interface LightningMessage {
                 PhoenixAndroidLegacyInfo.type -> PhoenixAndroidLegacyInfo.read(stream)
                 PhoenixAndroidLegacyMigrate.type -> PhoenixAndroidLegacyMigrate.read(stream)
                 OnionMessage.type -> OnionMessage.read(stream)
-                else -> {
-                    logger.warning { "unhandled code=${code}, cannot decode input=${Hex.encode(input)}" }
-                    null
-                }
+                else -> UnknownMessage(code.toLong())
```
pm47, line 94: It's now the responsibility of the caller to display the warning.

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -1840,3 +1834,13 @@ data class OnionMessage(
         }
     }
 }
+
+data class UnknownMessage(
+    override val type: Long
+
+) : LightningMessage {
+    override fun write(out: Output) {
+        TODO("Serialization of unknown messages is not implemented")
```
pm47, line 1843: We don't need that right?

src/commonMain/kotlin/fr/acinq/lightning/crypto/sphinx/Sphinx.kt
```diff
@@ -295,11 +294,8 @@ data class DecryptedFailurePacket(val originNode: PublicKey, val failureMessage:
  * +----------------+----------------------------------+-----------------+----------------------+-----+
  * with failure message length + pad length = 256
  */
-@ThreadLocal
```
pm47, line 298: I suppose we can remove all the `@ThreadLocal`

pm47:
Rebased.

pm47:
Rebased.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClient.kt
```diff
@@ -52,11 +53,11 @@ internal data class BroadcastStatus(val connection: Connection) : ElectrumClient
  *
  */
 internal sealed class ClientState {
```
sstone, line 55: Why not extend `LoggingContext` here as in `ChanelState` ?

src/commonMain/kotlin/fr/acinq/lightning/crypto/sphinx/Sphinx.kt
sstone: Yes (it was there because of `FailurePacket.logger`)

src/commonMain/kotlin/fr/acinq/lightning/utils/logger.kt
```diff
@@ -1,34 +1,10 @@
...
-
-@Suppress("ObjectPropertyName")
-private var LightningLoggerFactory: LoggerFactory? = null
-
-fun setLightningLoggerFactory(loggerFactory: LoggerFactory) {
-    require(LightningLoggerFactory == null) { "LightningLoggerFactory has already been accessed." }
-    LightningLoggerFactory = loggerFactory
-}
-
-class LightningLoggerDelegateProvider(val of: KClass<*>) : PropertyDelegateProvider<Any, Lazy<Logger>> {
```
sstone, line 19: Removing this is probably a very good thing wr.t kotlin's new memory model

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
sstone: No I don't see a use-case where we would need to persist unknown LN messages.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClient.kt
pm47: Two reasons:
- The `Logger` should in my opinion not be part of the state data of  an FSM. That is the point I was making here: https://github.com/ACINQ/lightning-kmp/pull/355#discussion_r973153903. If anything we should make `ChannelState` be like `ClientState`, not the opposite. But in the case of the `ChannelState`, the top-level structure already contained a reference to `nodeParams`, so it was really less effort to do it there. Contrast to `ClientState` which is just an interface, we would have to explicitly add the logger to it.
- In the case of the channel, the "machine" is actually the `Peer`, so it would be quite messy to have the `Peer` pass a channel-related `Logger` to `ChannelState`. It's different from the couple `ElectrumState`/`ClientState` where states do not appear outside of the FSM.

dpad85 (approved):
Tested on Android and iOS - on iOS, we currently have unrelated issues coming from the change to kotlin 1.6 and the upgrade to the new memory model that are [still unresolved](https://github.com/ACINQ/phoenix/pull/299#issuecomment-1256343769), so the test was pretty shallow. Still I think we can go ahead with this change and merge this PR, that way we will test everything once.

Regarding the issue fixed in 7de9643 it should be investigated separately. This is probably an issue in the kodein-log library, version 0.11.1, that was not detected beforehand as we did not use kodein-log classes in the iosMain module before this PR.

PR #361, opened 2022-09-05
Use `kotlinx-datetime` instead of custom impls for timestamp

pm47:
This has been extracted from #348 and was successfully tested by @dpad85 (https://github.com/ACINQ/lightning-kmp/pull/348#pullrequestreview-1085859557).


build.gradle.kts
```diff
@@ -39,6 +39,7 @@ kotlin {
             api("org.jetbrains.kotlinx:kotlinx-serialization-core:$serializationVersion")
             api("org.jetbrains.kotlinx:kotlinx-serialization-cbor:$serializationVersion")
             api("org.jetbrains.kotlinx:kotlinx-serialization-json:$serializationVersion")
+            api("org.jetbrains.kotlinx:kotlinx-datetime:0.2.0")
```
t-bast, line 42: Can't we use v0.4.0?

pm47: Possibly, but I went with the safest choice since we already depend on this version:

```
+--- org.kodein.log:kodein-log:0.11.1
    \--- org.kodein.log:kodein-log-jvm:0.11.1
         +--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.5.0 -> 1.5.31 (*)
         \--- org.jetbrains.kotlinx:kotlinx-datetime-jvm:0.2.0
```

t-bast (approved):

PR #362, opened 2022-09-08
Fix electrum state when disconnecting

dpad85:
This PR fixes an issue in the electrum connection `state`. Disconnecting Electrum while connecting would not update the state of the connection, which would get stuck in the `Connecting` state. As such the reconnection logic in Phoenix would not attempt to reconnect to Electrum.

We also enable hierarchical multiplatform and cinterop commonization in gradle properties. This fixes a few issues in the IDE. See https://youtrack.jetbrains.com/issue/KT-40975

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,6 +6,7 @@ kotlin.code.style=official
 kotlin.incremental.multiplatform=true
 kotlin.mpp.stability.nowarn=true
 # https://youtrack.jetbrains.com/issue/KT-36086
-#kotlin.mpp.enableGranularSourceSetsMetadata=true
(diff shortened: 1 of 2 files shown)


robbiehanson:
For branch `fix-electrum-disconnect`, when I try to `publishToMavenLocal` I get:

```
> Task :compileIosMainKotlinMetadata FAILED
w: library included more than once: /Users/robbie/.konan/kotlin-native-prebuilt-macos-x86_64-1.5.31/klib/common/stdlib
e: warnings found and -Werror specified

FAILURE: Build failed with an exception.
```

robbiehanson (approved):

PR #363, opened 2022-09-09
Use IO dispatcher when connecting to a pinned electrum server on the JVM

dpad85:
This PR fixes an issue where connections to an electrum server using a pinned certificate would change dispatcher.

We also update ktor to 1.6.5 which is the latest ktor version supporting kotlin 1.5.31.


robbiehanson (approved):
Looks like we're going to need to upgrade Kotlin soon too.

PR #365, opened 2022-09-13
Rework channel funding mechanism

t-bast:
Instead of asking Electrum to create the funding tx, we start the opening flow with a set of inputs that should be used. This is a better design for how mobile wallets will create channels: the user will choose a funding amount and select utxos that should be used (or swap them into the wallet from an external bitcoin wallet). We currently assume p2wpkh inputs only, but that can easily be changed later.

Note that this channel funder side isn't used in Phoenix yet, so this change shouldn't have any impact on anything else than tests.


src/commonMain/kotlin/fr/acinq/lightning/serialization/v1/ChannelState.kt
```diff
@@ -523,20 +522,36 @@ data class WaitForFundingCreated(
...
     @Serializable(with = SatoshiKSerializer::class) val fundingAmount: Satoshi,
+    val inputs: List<FundingInput>,
+    @Serializable(with = PublicKeyKSerializer::class) val changePubKey: PublicKey?,
+) {
+    constructor(from: fr.acinq.lightning.channel.FundingInputs) : this(from.fundingAmount, from.inputs.map { FundingInput(it) }, from.changePubKey)
+}
+
+@Serializable
+data class InitInitiator(
+    val fundingInputs: FundingInputs,
```
t-bast, line 544: IIUC, we can safely change those classes without impacting serialization because this is only used *before* we actually have commitments, so it's never stored in the DB and only used to provide JSON serialization for the app.

sstone:
Ok I see now how working incrementally could be quicker and easier to review than putting together a huge PR that does everything, and as you point out we'll need to provide local utxos before we initiate the dual funding workflow. So this is fine :)

src/commonMain/kotlin/fr/acinq/lightning/channel/WaitForAcceptChannel.kt
```diff
@@ -79,6 +117,33 @@ data class WaitForAcceptChannel(
         }
     }
 
+    private fun createFundingTx(fundingPubkeyScript: ByteVector): Pair<Transaction, Satoshi> {
+        val txIn = init.fundingInputs.inputs.map { i -> TxIn(i.outpoint, 0) }
```
sstone, line 121: nit: `txIn` is a collection, it should be a plural like `inputs`

src/commonMain/kotlin/fr/acinq/lightning/channel/WaitForAcceptChannel.kt
```diff
@@ -79,6 +117,33 @@ data class WaitForAcceptChannel(
...
+        val unsignedTx = when (val changePubKey = init.fundingInputs.changePubKey) {
+            null -> Transaction(2, txIn, listOf(TxOut(init.fundingAmount, fundingPubkeyScript)), currentBlockHeight.toLong())
```
sstone, line 123: If don't understand why we compute and substract fees only when a change pubkey is provided ?

src/commonMain/kotlin/fr/acinq/lightning/serialization/v1/ChannelState.kt
sstone: Yes, the current code does not distinguish between classes that are serializable because we need to display them as JSON for example and classes that are serializable because they need to be persisted in our database.

src/commonMain/kotlin/fr/acinq/lightning/channel/WaitForAcceptChannel.kt
t-bast: Because if we don't have a change pubkey, we cannot create a change output at all, we have no instructions on where to send the funds?

src/commonMain/kotlin/fr/acinq/lightning/channel/WaitForAcceptChannel.kt
t-bast: Fixed in https://github.com/ACINQ/lightning-kmp/pull/365/commits/79e69505c70b9c4ffd1f3795ee071bf77ca28296

src/commonMain/kotlin/fr/acinq/lightning/channel/WaitForAcceptChannel.kt
t-bast: The reason the change pubkey is optional is that when you're doing a swap-in, you'll want to use all the swapped-in funds and don't want a change output. This also simplifies the wallet which won't have to track as many addresses since unspent funds are generally fully consumed by a channel open without generating new unspent funds.

sstone: Then fees could be subtracted from the amount being sent. This is what bitcoin core does for example.
What seems to be inconsistent here is how fees are handled: caller should select utxos, compute fees and subtract them from `fundingAmount` if they don't include a change pubkey ? how would since work if inputs are combined with other external inputs ?

t-bast: I removed the change pubkey entirely in https://github.com/ACINQ/lightning-kmp/pull/365/commits/1b61102ac777eafb33d5b1aa2914fb5fc942a3b3 for simplicity.

sstone (approved):

PR #368, opened 2022-09-19
Implement dual funding

t-bast:
In this PR, we change the funding flow to use [dual-funding](https://github.com/lightning/bolts/pull/851). We completely remove the previous funding flow, as dual funding is a super-set of it.

This (big) PR is split in several independent commits, which should make the review easier. The design is largely inspired by eclair's implementation, which should also help reviewers.

A few notes:

- we automatically move existing channels to use a 1% reserve (for the side that does have a reserve): this should be ok since this is already the settings used by Phoenix (configured [here](https://github.com/ACINQ/phoenix/blob/76260ebcea6aa2e0ae4ffe7f2e41d5b0a383fdc7/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/NodeParamsManager.kt#L104), note that the comment is inaccurate and just copied from test code)
- when we rework serialization, we should be clever when dealing with multiple `Commitments` object (as the `alternativeCommitments` in the `Closing` state) and serialize the common parts (`localParams`, `remoteParams`, etc) only once
- to correctly support the `send all` feature, we will need to set `max-htlc-value-in-flight-msat` to the full funding amount before we actually know the remote funding contributions: we should implement a `please_open_channel` handshake that takes place before dual funding and lets both parties commit to a shared funding amount


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -611,7 +602,7 @@ class Peer(
                             onChainFeeratesFlow.filterNotNull().first()
                         )
                         val channelConfig = ChannelConfig.standard
-                        // Note that we don't contribute yet to the channel funding (we need dual funding for that).
+                        // We currently don't add any inputs to the funding transaction.
                         val (state1, actions1) = state.process(ChannelEvent.InitNonInitiator(msg.temporaryChannelId, FundingInputs.empty, localParams, channelConfig, theirInit!!))
                         val (state2, actions2) = state1.process(ChannelEvent.MessageReceived(msg))
                         processActions(msg.temporaryChannelId, actions1 + actions2)
```
pm47, line 608: Those lines need to be interverted, otherwise `processActions(ChannelAction.ChannelId.IdAssigned` is a no-op because the `_channels` map is empty:
```scala
_channels = _channels + (msg.temporaryChannelId to state2)
processActions(msg.temporaryChannelId, actions1 + actions2)
```

t-bast: We're definitely missing a test here, I'll fix that and add one when I rebase!

pm47: There is another related issue. The code below (which was in the `HasTemporaryChannelId`) needs to be executed to clean up the old binding in the `_channels` map:

```scala
actions.filterIsInstance<ChannelAction.ChannelId.IdAssigned>().forEach {
  _channels = _channels - it.temporaryChannelId + (it.channelId to state1)
}
```

src/commonMain/kotlin/fr/acinq/lightning/channel/WaitForFundingSigned.kt
```diff
@@ -2,100 +2,126 @@ package fr.acinq.lightning.channel
...
-import fr.acinq.lightning.transactions.Transactions
-import fr.acinq.lightning.utils.Either
-import fr.acinq.lightning.utils.Try
-import fr.acinq.lightning.wire.Error
-import fr.acinq.lightning.wire.FundingCreated
-import fr.acinq.lightning.wire.FundingSigned
+import fr.acinq.lightning.wire.*
+import kotlin.math.absoluteValue
 
 data class WaitForFundingSigned(
```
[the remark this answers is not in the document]
t-bast: This state handles the exchange of `commit_sig` (which makes me realize that I should add comments to each state class to explain what part of the protocol they're supposed to handle).

We could do it in `WaitForFundingCreated`, but it becomes a bit messier because we'd have to keep more variable state inside, I didn't like it...

pm47 (approved):
LGTM, this was successfully tested against eclair on a feature branch.

PR #369, opened 2022-09-19
Remove legacy trampoline feature bit

t-bast:
Since this creates issues with lnd v0.15, we fully remove this legacy feature bit (but still support reading it in invoices).

However, since our serialisation uses Kotlin serialisation for features (instead of using the spec serialisation), this breaks old serialised data that contains this feature bit. So we should only merge this PR once we're confident that all users aren't activating the legacy `TrampolinePayment` feature bit anymore.

The change in the compatibility tests shows that this is breaking backwards-compatibility for those users.

To minimize compatibility issues, we will integrate that change when we introduce the final, spec-compliant trampoline feature bit.


t-bast:
Actually we cannot change that without breaking backwards-compatibility, because this feature is activated in current Phoenix's node params...our serialization mechanism is really limiting here because this will prevent us from properly moving bit 51 to be zero-conf.

What we should do is to just disable the legacy trampoline feature bit in the next release of Phoenix iOS, mandate everyone to update (since it also fixes the payment issue with lnd v0.15), and then we can safely break backwards-compatibility with this PR.

PR #370, opened 2022-09-19
Update to Kotlin 1.6 and use the new Memory Model

pm47:
The old memory model based on freezing objects that moved between threads caused a lot of issues with Phoenix, especially on iOS. With the upgrade to kotlin 1.6, ktor 2.x and coroutines 1.6 it makes sense to also move to the new model which is much more workable than the old one.

To do so we enable an experimental flag and use the adequate kotlinx coroutine library version.

See https://github.com/JetBrains/kotlin/blob/master/kotlin-native/NEW_MM.md


src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
```diff
@@ -368,6 +369,7 @@ data class Closing(
                     when (it) {
                         is Transactions.TransactionWithInputInfo.ClaimHtlcTx.ClaimHtlcSuccessTx -> add(it.tx to "remote-htlc-success")
                         is Transactions.TransactionWithInputInfo.ClaimHtlcTx.ClaimHtlcTimeoutTx -> add(it.tx to "remote-htlc-timeout")
+                        else -> Unit
```
sstone, line 372: same as above

src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
```diff
@@ -378,6 +380,7 @@ data class Closing(
...
```
sstone, line 383: same as above

src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
```diff
@@ -357,6 +357,7 @@ data class Closing(
                     when (it) {
                         is Transactions.TransactionWithInputInfo.HtlcTx.HtlcSuccessTx -> add(it.tx to "local-htlc-success")
                         is Transactions.TransactionWithInputInfo.HtlcTx.HtlcTimeoutTx -> add(it.tx to "local-htlc-timeout")
+                        else -> Unit
```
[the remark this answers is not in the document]
pm47: Done in d75e77b.

src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
pm47: Done in https://github.com/ACINQ/lightning-kmp/commit/d75e77b588fef36607f8b5639cb34e7d1fcaf660.

src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
pm47: Done in https://github.com/ACINQ/lightning-kmp/commit/d75e77b588fef36607f8b5639cb34e7d1fcaf660.

pm47:
Rebased and cleaned up, I  think we're good to go!

dpad85 (approved):
LGTM!

PR #372, opened 2022-09-22
Update Phoenix channel type

t-bast:
We replace the channel type used by Phoenix to use anchor outputs and the experimental zero-reserve feature bit.

We deprecate the previous experimental zero-conf feature bit: we instead decide to use zero-conf or not based on a peer whitelist (and simply send `funding_locked` early when we trust our peer). In the future, that decision can also be based on the details of what we're doing (channel funding, splice) and who contributes what. Such logic will most likely only exist on the LSP side, while the wallet will use zero-conf whenever the LSP is in their whitelist.


src/commonMain/kotlin/fr/acinq/lightning/serialization/v1/ChannelState.kt
```diff
@@ -270,15 +270,7 @@ data class ChannelVersion(@Serializable(with = ByteVectorKSerializer::class) val
...
-        val channelFeatures = fr.acinq.lightning.channel.ChannelFeatures(
-            setOf(
-                Feature.Wumbo,
-                Feature.StaticRemoteKey,
-                Feature.AnchorOutputs,
-                Feature.ZeroReserveChannels,
-                Feature.ZeroConfChannels,
-            )
-        )
+        val channelFeatures = fr.acinq.lightning.channel.ChannelFeatures(setOf(Feature.StaticRemoteKey, Feature.AnchorOutputs, Feature.ZeroReserveChannels))
```
pm47, line 273: I think we should leave old channels as-is.

src/commonMain/kotlin/fr/acinq/lightning/channel/WaitForFundingSigned.kt
```diff
@@ -62,7 +61,7 @@ data class WaitForFundingSigned(
                     is Helpers.Funding.FirstCommitments -> {
                         val (signedFundingTx, commitments) = firstCommitmentsRes
                         logger.info { "c:$channelId funding tx created with txId=${commitments.fundingTxId}. ${fundingTx.localInputs.size} local inputs, ${fundingTx.remoteInputs.size} remote inputs, ${fundingTx.localOutputs.size} local outputs and ${fundingTx.remoteOutputs.size} remote outputs" }
-                        if (commitments.channelFeatures.hasFeature(Feature.ZeroConfChannels)) {
+                        if (staticParams.useZeroConf) {
```
pm47, line 64: We need to send an alias in the `channel_ready` tlv to be able to user zero-conf.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v1/ChannelState.kt
t-bast: Reverted in https://github.com/ACINQ/lightning-kmp/pull/372/commits/d158718b3f49b8e72fe737be4cda2b35c789dcfe

pm47 (approved):

PR #373, opened 2022-09-27
Add `please_open_channel` mechanism for trustless swap-in

t-bast:
This is a new message that lets a (future) non-initiator wallet ask an LSP to open a channel (against a small fee).

This mechanism will be used to replace the previous trusted swap-in protocol: the result will be a fully trustless protocol where the wallet user chooses which utxos to add to the channel (thanks to dual funding).

We also include a temporary tlv that helps provide smooth migration for Phoenix users on android.

NB: this builds on top of #372


src/commonMain/kotlin/fr/acinq/lightning/wire/ChannelTlv.kt
```diff
@@ -244,3 +223,52 @@ sealed class ClosingSignedTlv : Tlv {
...
+
+        companion object : TlvValueReader<MaxFees> {
+            const val tag: Long = 1
+            override fun read(input: Input): MaxFees = MaxFees(LightningCodecs.tu16(input))
+        }
+    }
+
+    // NB: this is a temporary tlv that is only used to ensure a smooth migration to lightning-kmp for the android version of Phoenix.
+    @Serializable
+    data class GrandParents(val outpoints: List<@Contextual OutPoint>) : PleaseOpenChannelTlv() {
```
pm47, line 242: I think a list of `txid` (corresponding to mutual closes) would suffice.

t-bast: Since we have all the information and it doesn't take more space, it's probably better to include everything?

For example, we should guard against inputs that come from a funding tx, but are actually the change output, not the funding outpoint. For old Phoenix channels this cannot be the case (since we're the only funder so they don't have a change output), but maybe in the future this could be exploited, so it's better to be strict since it (almost) doesn't require more code?

pm47: I don't think we need to go all the way up to the funding transaction, we can use the mutual close tx, where there is no change. To me, the grand parent inputs refer to the inputs of the mutual close tx.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v3/ChannelState.kt
```diff
@@ -487,54 +491,82 @@ data class WaitForFundingCreated(
...
-    val outputIndex: Int,
-    @Serializable(with = PrivateKeyKSerializer::class) val privateKey: PrivateKey
+data class UnspentItem(@Serializable(with = ByteVector32KSerializer::class) val txid: ByteVector32, val outputIndex: Int, val value: Long, val blockHeight: Long) {
+    constructor(from: fr.acinq.lightning.blockchain.electrum.UnspentItem) : this(from.txid, from.outputIndex, from.value, from.blockHeight)
+
+    fun export() = fr.acinq.lightning.blockchain.electrum.UnspentItem(txid, outputIndex, value, blockHeight)
+}
+
+@Serializable
+data class WalletState(
```
pm47, line 530: Note to self : 👓 this

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -581,41 +587,56 @@ class Peer(
...
-                        val (state2, actions2) = state1.process(ChannelEvent.MessageReceived(msg))
-                        processActions(msg.temporaryChannelId, actions1 + actions2)
-                        _channels = _channels + (msg.temporaryChannelId to state2)
-                        logger.info { "n:$remoteNodeId c:${msg.temporaryChannelId} new state: ${state2::class}" }
+                    msg is OpenDualFundedChannel -> {
+                        val (wallet, fundingAmount, pushAmount) = msg.tlvStream.get<ChannelTlv.PleaseOpenChannelResponse>()?.let { response ->
+                            channelRequests[response.requestId]?.let { request ->
+                                // We have to pay the fees for our inputs, so we deduce them from our funding amount.
+                                val utxos = request.wallet.spendable()
+                                val fundingFees = Transactions.weight2fee(msg.fundingFeerate, utxos.size * Transactions.p2wpkhInputWeight)
```
pm47, line 601: Note to self: 👓 this

src/commonMain/kotlin/fr/acinq/lightning/wire/ChannelTlv.kt
pm47: It is, in fact, much easier to have a list of grandparents outpoints like you did. It is trivial to find out ancester channel ids that way.

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
```diff
@@ -142,13 +139,17 @@ data class IncomingPayment(val preimage: ByteVector32, val origin: Origin, val r
         /** KeySend payments are spontaneous donations for which we didn't create an invoice. */
         object KeySend : Origin()
 
-        /** Swap-in works by sending an on-chain transaction to a swap server, which will pay us in exchange. We may not know the origin address. */
+        /** DEPRECATED: this is the legacy trusted swap-in, which we keep for backwards-compatibility (previous payments inside the DB). */
         data class SwapIn(val address: String?) : Origin()
 
+        /** Trust-less swap-in based on dual funding. */
+        data class DualSwapIn(val localInputs: Set<OutPoint>) : Origin()
```
t-bast, line 146: @dpad85 let me know if that works for you? I believe this lets us point to explorers that will show the swap transactions.

pm47:
Why can't we clean up those? We're not using them at all IIUC so serialization shouldn't be a problem.
https://github.com/ACINQ/lightning-kmp/blob/08888e6a088d83c0103e959bf0062fb79e0cea48/src/commonMain/kotlin/fr/acinq/lightning/Features.kt#L197-L211

t-bast:
Agreed, we can remove them!

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
dpad85: Yep it's better to have the outpoint/transaction than just the address as we used to with legacy swap-in.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -203,6 +211,26 @@ class Peer(
...
+                    logger.info { "${wallet.spendableBalance} available on final wallet" }
+                }
+            }
+        }
+        launch {
+            swapInWallet.walletStateFlow
+                .takeWhile { wallet ->
+                    if (wallet.spendableBalance > 10_000.sat) {
+                        logger.info { "${wallet.spendableBalance} available on swap-in wallet, requesting channel" }
+                        input.send(RequestChannelOpen(randomBytes32(), wallet, maxFeesBasisPoint = 100))
```
t-bast, line 226: Note that we mustn't forget to fetch `maxFeesBasisPoint` from the user's settings somehow.

pm47 (approved):

PR #375, opened 2022-09-29
Run electrum client and wallet tests

sstone:
We run theses tests on all platforms except linux (because the tcp socket implementation is missing).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -294,7 +294,10 @@ tasks.withType<AbstractTestTask> {
 // Linux native does not support integration tests (sockets are not implemented in Linux native)
 if (currentOs.isLinux) {
     val linuxTest by tasks.getting(KotlinNativeTest::class) {
+        // we don't run these tests on linux because the tcp socket implementation is missing
         filter.excludeTestsMatching("*IntegrationTest")
+        filter.excludeTestsMatching("*ElectrumClientTest")
+        filter.excludeTestsMatching("*ElectrumMiniWalletTest")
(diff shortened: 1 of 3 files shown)


dpad85 (dismissed):

pm47 (commented):
More accurate description:

We can't run integration tests on a native platform because:
- our integration tests need docker and a platform-dependent tcp client 
- iOS has a native tcp impl but CI servers don't have docker
- Linux CI servers have docker but don't have a native tcp impl

However, some integration tests don't actually need docker, and those can run on iOS. To do that, we:
- remove them from the integration tests set by renaming them => they will run on iOS
- manually exclude them from the native Linux platform

Of course all integration tests can run on the jvm platform on Linux. In the end, we end up with:
- iOS native : run non-docker integration tests
- Linux native : no integration tests
- Linux jvm : run all integration tests

sstone:
Rebased, with an updated commit comment.

build.gradle.kts
```diff
@@ -294,7 +294,10 @@ tasks.withType<AbstractTestTask> {
 // Linux native does not support integration tests (sockets are not implemented in Linux native)
 if (currentOs.isLinux) {
     val linuxTest by tasks.getting(KotlinNativeTest::class) {
+        // we don't run these tests on linux because the tcp socket implementation is missing
```
pm47, line 297: Comment is duplicated with line 294
```suggestion
```

pm47 (approved):

PR #377, opened 2022-09-30
Skip JVM tests on macOS GitHub CI

dpad85:
The `check` GitHub action on MacOS is slow. We try to improve that marginally by skipping the JVM tests which are already ran on Ubuntu.

Note that the macOS build is slow mostly because the `iosX64Test` task is slow. This PR does not improve that.

We also remove all nominal outputs from the `PhoenixCrypto` build. Only warning and errors will be printed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,5 +63,5 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: ./gradlew build -PintegrationTests=include
       - name: Check without integration
-        if: matrix.os != 'ubuntu-latest'
-        run: ./gradlew build
+        if: matrix.os == 'macOS-latest'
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -63,5 +63,5 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: ./gradlew build -PintegrationTests=include
       - name: Check without integration
-        if: matrix.os != 'ubuntu-latest'
-        run: ./gradlew build
+        if: matrix.os == 'macOS-latest'
```
pm47, line 66: Nice touch

pm47 (commented):
-20% test time is not bad

pm47 (approved):

PR #378, opened 2022-10-03
Upgrade to Kotlin 1.7.20

dpad85:
This PR upgrade kotlin to 1.7.20, along with several dependencies:

- ktor to 2.1.2
- kotlinx.coroutine to 1.6.4
- kotlinx.serialization to 1.4.0
- kotlinx.datetime to 0.4.0

With kotlin 1.7.20, the memory model is enabled by default and performances should be better. There are also some options (e.g. [call suspend from swift](https://kotlinlang.org/docs/whatsnew1720.html#calling-kotlin-suspending-functions-from-swift-objective-c)) that could be interesting on iOS.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -69,7 +69,6 @@ kotlin {
             api(ktor("network"))
             api(ktor("network-tls"))
             implementation("org.slf4j:slf4j-api:$slf4jVersion")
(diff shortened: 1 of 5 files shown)


dpad85:
The upgrade to 1.7.20 is postponed for now, will reopen this PR when needed.

PR #383, opened 2022-10-05
Revert "Remove migration prototype (#381)"

pm47:
This reverts commit 659f22cd9b5207697fba7c1fcbebbde0b3ca3d9e.

I jumped the gun: independently of the migration method, we still need a way for Legacy Phoenix to tell if they have channels. So we need to keep those messages.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -102,6 +102,8 @@ data class ChannelClosing(val channelId: ByteVector32) : PeerEvent()
 data class SendSwapOutRequest(val amount: Satoshi, val bitcoinAddress: String, val feePerKw: Long) : PeerCommand()
 data class SwapOutResponseEvent(val swapOutResponse: SwapOutResponse) : PeerEvent()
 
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -102,6 +102,9 @@ data class ChannelClosing(val channelId: ByteVector32) : PeerEvent()
 data class SendSwapOutRequest(val amount: Satoshi, val bitcoinAddress: String, val feePerKw: Long) : PeerCommand()
 data class SwapOutResponseEvent(val swapOutResponse: SwapOutResponse) : PeerEvent()
 
+data class SendPhoenixAndroidLegacyMigrate(val newNodeId: PublicKey) : PeerCommand()
```
dpad85, line 105: `SendPhoenixAndroidLegacyMigrate` is a message sent by the phoenix-legacy app to let the peer know when the user started the migration process, so that the peer could start the required migration logic on its side, if there's any.

With dual funding I think this message is not needed anymore as the process is stateless on the peer side.

pm47: Done with [f202d93](https://github.com/ACINQ/lightning-kmp/pull/383/commits/f202d93dfdaa0b427cd04ecc1cfc5de437650d74).

dpad85 (approved):

PR #384, opened 2022-10-07
Support Electrum requests multiplexing

pm47:
This is the minimum change to make `ElectrumClient` support multiple independent "callers" (watcher, wallets).

Requests made by independent callers are identified with a unique id and multiplexed on the same connection. Each caller then listens to the `notifications` flow and only keeps responses for his identifiers. For convenience, this is encapsulated in an inner class `ElectrumClient.Caller` which presents the same interface as before multiplexing.

Note that subscribtion responses are currently _not_ multiplexed. They don't have an associated id and are sent to all callers.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClient.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClient.kt
@@ -210,11 +210,10 @@ class ElectrumClient(
 
     // connection status to the server
     private val _connectionState = MutableStateFlow<Connection>(Connection.CLOSED(null))
-    val connectionState: StateFlow<Connection> get() = _connectionState
+    val connectionState: StateFlow<Connection> get() = _connectionState.asStateFlow()
 
     // subscriptions notifications (headers, script_hashes, etc.)
     private val _notifications = MutableSharedFlow<ElectrumClientNotification>(replay = 0, extraBufferCapacity = 64, onBufferOverflow = BufferOverflow.SUSPEND)
-    val notifications: SharedFlow<ElectrumClientNotification> get() = _notifications.asSharedFlow()
 
     private var state: ClientState = ClientClosed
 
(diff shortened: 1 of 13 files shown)


src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClient.kt
```diff
@@ -25,31 +25,30 @@ import kotlin.time.Duration.Companion.seconds
...
 }
 
 /** Actions */
-internal sealed interface ElectrumClientAction {
+sealed interface ElectrumClientAction {
+    data class SendHeader(val ids: Set<UUID>, val height: Int, val blockHeader: BlockHeader) : ElectrumClientAction
     data class SendRequest(val request: String) : ElectrumClientAction
-    data class SendHeader(val height: Int, val blockHeader: BlockHeader) : ElectrumClientAction
-    data class SendResponse(val response: ElectrumResponse) : ElectrumClientAction
+    data class SendResponse(val ids: Set<UUID>, val response: ElectrumResponse) : ElectrumClientAction
```
sstone, line 40: what is the use case for having multiple different ids here ?

pm47: It started as a way to handle subscriptions, but I abandonned that because it seemed like too much work. Instead I went for a broadcast behavior, but I guess a nullable id would work. Let me try that.

pm47: Good catch, done in 13ca0c4. Also use a simpler id in 1ca2ba4.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClient.kt
```diff
@@ -25,31 +25,30 @@ import kotlin.time.Duration.Companion.seconds
...
 }
 
 /** Actions */
-internal sealed interface ElectrumClientAction {
+sealed interface ElectrumClientAction {
+    data class SendHeader(val callerId: UUID, val height: Int, val blockHeader: BlockHeader) : ElectrumClientAction
     data class SendRequest(val request: String) : ElectrumClientAction
-    data class SendHeader(val height: Int, val blockHeader: BlockHeader) : ElectrumClientAction
-    data class SendResponse(val response: ElectrumResponse) : ElectrumClientAction
+    data class SendResponse(val callerId: UUID?, val response: ElectrumResponse) : ElectrumClientAction
```
sstone, line 40: nit: why not have 2 different messages now:

- `SendSubscriptionResponse` without a caller id
- `SendResponse` with a caller id that cannot be `null`

sstone (approved):

PR #386, opened 2022-10-11
Dual swap-in fixes and improvements

pm47:
Minor fixes, best reviewed commit-by-commit.

The consistency flag is a bit hacky and needs to be revisited later. Another way would have been to use an internal data structure (with electrum types) and only emit clean `WalletState`s (with no electrum types).


t-bast (commented):
Would it be really hard to get rid of the `consistent` flag? It looks like this can be evaluated internally whenever we receive electrum data to ensure we only emit a `WalletState` when we have all the data.

If this is too much work, we can do it later, let me know. Apart from that, LGTM.

pm47:
> Would it be really hard to get rid of the consistent flag? It looks like this can be evaluated internally whenever we receive electrum data to ensure we only emit a WalletState when we have all the data.

> If this is too much work, we can do it later, let me know. Apart from that, LGTM.

Yes it's possible, it's not even very hard but I feel like it's better done later, and make a stable checkpoint now. I'm pretty sure we will revisit the entire wallet when we actually derive keys.

t-bast (approved):

PR #387, opened 2022-10-13
Fix peer state when disconnecting

robbiehanson:
This PR fixes an issue in the peer connection state. Disconnecting Peer while connecting would not update the state of the connection, which would get stuck in the Connecting state. As such the reconnection logic in Phoenix would not attempt to reconnect to Peer.

You'll notice this is the exact same bugfix we performed in #362 for the ElectrumClient :sweat_smile:


dpad85 (approved):
Good catch 🥵

PR #390, opened 2022-10-14
Mark all v1,v2,v3 serialization classes as private

sstone:
Their uses will be restricted to lightning-kmp. For example Phoenix will not be able to use them directly any more and will have to use methods provided by this library for JSON serialization.


pm47:
Let's see: serialization `v2` has been introduced in https://github.com/ACINQ/lightning-kmp/pull/262 and was included at the earliest in tag [v1.0-beta12](https://github.com/ACINQ/lightning-kmp/releases/tag/v1.0-beta12). @dpad85 what was the first publicly accessible mainnet release using and what version of lightning-kmp was it using?

dpad85:
> @dpad85 what was the first publicly accessible mainnet release using and what version of lightning-kmp was it using?

Earliest version Phoenix version on iOS is 1.0.0, using [lightning-kmp 1.0-beta13](https://github.com/ACINQ/phoenix/blob/4c768cb3e837e8fcd3980da6150d8433f8e6c99a/buildSrc/src/main/kotlin/Versions.kt#L2)

Also, 97% of known active iOS users are on phoenix 1.5.0+, depending on lightning-kmp 1.4.1 (which is the latest tag).

sstone:
> Also, 97% of known active iOS users are on phoenix 1.5.0+, depending on lightning-kmp 1.4.1 (which is the latest tag).

Then we can just remove the v1 package

sstone:
Superseded by https://github.com/ACINQ/lightning-kmp/pull/391

PR #391, opened 2022-10-17
Rework channel data model and serialization

pm47:
Strap yourself in, that's a big one!

This PR _has_ to be reviewed commit by commit.

Here are the main changes:
- make `ChannelState` only contain channel state. One can now serialize/deserialize channel data without any additional context
- introduce a new self-contained serialization for channel persistence that doesn't depend on `kotlinx.serialization`
- introduce a new self-contained serialization for json, using external serializers
- clean up legacy serialization code by making it read-only and removing unused code
- make legacy serialization code as self-contained as possible using external serializers, and clean up the codebase from `Serializable` annotations
- add a comprehensive non-reg test for legacy serialization

Don't be scared by the added line count, it's just test data. This PR actually removes a little less than 1000 loc.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -77,12 +77,6 @@ kotlin {
             implementation(kotlin("test-junit"))
             implementation("org.bouncycastle:bcprov-jdk15on:1.64")
             implementation("ch.qos.logback:logback-classic:1.2.3")
-            implementation(ktor("server-netty"))
-            implementation(ktor("serialization"))
-            implementation(ktor("server-status-pages"))
-            implementation(ktor("server-content-negotiation"))
-            implementation(ktor("serialization-kotlinx-json"))
-            implementation("com.typesafe:config:1.4.1")
         }
     }
 
--- a/src/jvmTest/resources/application.conf
+++ b/src/jvmTest/resources/application.conf
@@ -1,6 +1,0 @@
-phoenix {
-  chain = "testnet"
-  #trampoline-node-uri = "02413957815d05abb7fc6d885622d5cdc5b7714db1478cb05813a8474179b83c5c@51.77.223.203:19735"
-  trampoline-node-uri = "03933884aaf1d6b108397e5efe5c86bcf2d8ca8d2f700eda99db9214fc2712b134@13.248.222.197:9735"
-  electrum-server = "testnet1.electrum.acinq.co:51002:tls"
-}


pm47:
@t-bast @sstone I'd like to have your opinion on [ec2ea61](https://github.com/ACINQ/lightning-kmp/pull/391/commits/ec2ea61d11ef1d369bf61fc77e5c9055416320e3) before I go further, as there is significant work to adapt the tests. I have detailed the rationale and explained the change in the commit description. Please disregard previous commits.

t-bast:
> @t-bast @sstone I'd like to have your opinion on [ec2ea61](https://github.com/ACINQ/lightning-kmp/pull/391/commits/ec2ea61d11ef1d369bf61fc77e5c9055416320e3) before I go further, as there is significant work to adapt the tests. I have detailed the rationale and explained the change in the commit description. Please disregard previous commits.

Thanks for the detailed description! I think this is going in the right direction, I was skeptical at first but I now think this is the right model and we should do this.

My only open question is whether we should use an extension method like you did or explicitly add a `context: ChannelContext` parameter to all methods that need it. I find the explicit parameter slightly better, and I'm not sure it requires more code changes, since it removes the need for all the `@SomeChannelState` that were added everywhere. But I don't have a strong feeling on that, and it can be changed in the future without breaking anything, so I'll let you guys choose what you think is best.

pm47:
@t-bast @sstone This wasn't my initial plan, but I went with custom serialization for the `v4` serialization in 39449e1c, and I'm pretty happy with the result. I explained the reasoning in the commit description, please let me know what you think 🙏 .

> My only open question is whether we should use an extension method like you did or explicitly add a context: ChannelContext parameter to all methods that need it.=

Indeed this can be changed easily in theory. The painful part was updating the tests, I'm glad it's over with d02d63da6cf9dad796f004542cdd58976ce75269. They were passing at that point, before I re-broke them with the work on v4.

t-bast:
> @t-bast @sstone This wasn't my initial plan, but I went with custom serialization for the v4 serialization in https://github.com/ACINQ/lightning-kmp/commit/39449e1caa14e54d79a5764f267b1a058210b284, and I'm pretty happy with the result. I explained the reasoning in the commit description, please let me know what you think pray .

I think that's the right decision. It may be painful in the short term, but this will give us more freedom and better assurances in the long run.

src/commonMain/kotlin/fr/acinq/lightning/json/JsonSerializers.kt
```diff
@@ -0,0 +1,437 @@
...
+            polymorphic(UpdateMessage::class) {
+                subclass(UpdateAddHtlc::class, UpdateAddHtlcSerializer)
+                subclass(UpdateFailHtlc::class, UpdateFailHtlcSerializer)
+                subclass(UpdateFailMalformedHtlc::class, UpdateFailMalformedHtlcSerializer)
+                subclass(UpdateFulfillHtlc::class, UpdateFulfillHtlcSerializer)
+                subclass(UpdateFee::class, UpdateFeeSerializer)
+            }
+            polymorphic(Tlv::class) {
+                subclass(ChannelReadyTlv.ShortChannelIdTlv::class, ChannelReadyTlvShortChannelIdTlvSerializer)
+                subclass(ShutdownTlv.ChannelData::class, ShutdownTlvChannelDataSerializer)
```
t-bast, line 151: I think we can omit the `ChannelData` tlv, can't we?

src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationNonRegTestsCommon.kt
```diff
@@ -0,0 +1,58 @@
...
+        regtest("v3", debug = true)
+    }
+
+    companion object {
+        fun fileFor(dir1: String, dir2: String, name: String): WriteableFile {
+            val file: Path = FileSystem.workingDir().resolve("nonreg", dir1, dir2, name)
+            file.parent().createDirs()
+            return file.openWriteableFile(false)
+        }
+    }
```
t-bast, line 57: This is unused...?

src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationNonRegTestsCommon.kt
```diff
@@ -0,0 +1,58 @@
...
+                assertEquals(ref, json, path.toString())
+                if (debug) {
+                    tmpFile.delete()
+                }
+            }
+    }
+
+    @Test
+    fun `non-reg test with v2 serialization`() {
+        regtest("v2", debug = true)
```
t-bast, line 43: We should set `debug` to `false` in this case and the one below before merging.

src/jvmTest/kotlin/fr/acinq/lightning/Node.kt
```diff
@@ -16,7 +16,7 @@ import fr.acinq.lightning.db.OutgoingPayment
 import fr.acinq.lightning.db.sqlite.SqliteChannelsDb
 import fr.acinq.lightning.io.*
 import fr.acinq.lightning.payment.PaymentRequest
-import fr.acinq.lightning.serialization.v1.Serialization
+import fr.acinq.lightning.serialization.v3.Serialization
```
t-bast, line 19: Why are we still using v3 serialization for anything here?

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Serialization.kt
```diff
@@ -0,0 +1,465 @@
...
+        writeCollection(claimAnchorTxs) { writeTransactionWithInputInfo(it) }
+        writeIrrevocablySpent(irrevocablySpent)
+    }
+
+    private fun Output.writeRemoteCommitPublished(o: RemoteCommitPublished) = o.run {
+        writeBtcObject(commitTx)
+        writeNullable(claimMainOutputTx) { writeTransactionWithInputInfo(it) }
+        writeCollection(claimHtlcTxs.entries) {
+            writeBtcObject(it.key)
+            writeNullable(it.value) { writeTransactionWithInputInfo(it) }
```
sstone, line 185:
```suggestion
            writeNullable(it.value) { claimHtlcTx -> writeTransactionWithInputInfo(claimHtlcTx) }
```

sstone (commented):
Nice and most probably easier to maintain and extend than what we have today !

How were non-reg test vectors generated ? (i.e. using the serialization code in this branch or using the actual releases/tags used by released versions of Phoenix)

What is the impact on Phoenix ? (I guess it will just need to use the new JSON serializers defined here)

Wouldn't it be easier to read and modify if serialization/deserialization code was in the same file and serializers/deserializers close to each other ?

src/commonMain/kotlin/fr/acinq/lightning/payment/PaymentRequest.kt
```diff
@@ -282,13 +269,12 @@ data class PaymentRequest(
...
     }
 
-    @Serializable
     sealed class TaggedField {
         abstract val tag: Int5
         abstract fun encode(): List<Int5>
 
         /** @param description a free-format string that will be included in the payment request */
-        @Serializable
+
```
t-bast, line 277: nit: the find-and-replace produced many empty lines between the documentation comment and its class in this file

src/commonMain/kotlin/fr/acinq/lightning/wire/ChannelTlv.kt
```diff
@@ -15,14 +15,11 @@ import fr.acinq.lightning.utils.msat
...
 import fr.acinq.lightning.utils.toByteVector32
-import kotlinx.serialization.Contextual
-import kotlinx.serialization.Serializable
 
-@Serializable
 sealed class ChannelTlv : Tlv {
     /** Commitment to where the funds will go in case of a mutual close, which remote node will enforce in case we're compromised. */
-    @Serializable
-    data class UpfrontShutdownScriptTlv(@Contextual val scriptPubkey: ByteVector) : ChannelTlv() {
+
```
t-bast, line 21: nit: remove empty line between documentation and class?

src/commonMain/kotlin/fr/acinq/lightning/wire/InteractiveTxTlv.kt
```diff
@@ -4,32 +4,30 @@ import fr.acinq.bitcoin.Satoshi
 import fr.acinq.bitcoin.io.Input
 import fr.acinq.bitcoin.io.Output
 import fr.acinq.lightning.utils.sat
-import kotlinx.serialization.Contextual
-import kotlinx.serialization.Serializable
 
-@Serializable
+
```
t-bast, line 8: nit: remove redundant empty lines?

src/commonMain/kotlin/fr/acinq/lightning/wire/MessageOnion.kt
```diff
@@ -7,17 +7,15 @@ import fr.acinq.bitcoin.io.ByteArrayOutput
...
-@Serializable
+
 sealed class OnionMessagePayloadTlv : Tlv {
     /**
      * Onion messages may provide a reply path, allowing the recipient to send a message back to the original sender.
      * The reply path uses route blinding, which ensures that the sender doesn't leak its identity to the recipient.
      */
-    @Serializable
-    data class ReplyPath(@Contextual val blindedRoute: RouteBlinding.BlindedRoute) : OnionMessagePayloadTlv() {
+
```
t-bast, line 17: nit: extra lines between documentation and class (same below)

src/commonMain/kotlin/fr/acinq/lightning/wire/PaymentOnion.kt
```diff
@@ -13,13 +13,11 @@ import fr.acinq.lightning.MilliSatoshi
...
 import fr.acinq.lightning.utils.msat
-import kotlinx.serialization.Contextual
-import kotlinx.serialization.Serializable
 
-@Serializable
+
 sealed class OnionPaymentPayloadTlv : Tlv {
     /** Amount to forward to the next node. */
-    @Serializable
+
```
t-bast, line 20: nit: same in this file

src/commonMain/kotlin/fr/acinq/lightning/wire/RouteBlinding.kt
```diff
@@ -6,14 +6,12 @@ import fr.acinq.bitcoin.io.ByteArrayInput
...
-import kotlinx.serialization.Contextual
-import kotlinx.serialization.Serializable
 
-@Serializable
+
 sealed class RouteBlindingEncryptedDataTlv : Tlv {
     /** Some padding can be added to ensure all payloads are the same size to improve privacy. */
-    @Serializable
-    data class Padding(@Contextual val dummy: ByteVector) : RouteBlindingEncryptedDataTlv() {
+
```
t-bast, line 13: nit: this file as well

src/jvmTest/kotlin/fr/acinq/lightning/db/sqlite/SqliteChannelsDb.kt
```diff
@@ -51,20 +52,21 @@ class SqliteChannelsDb(val nodeParams: NodeParams, val sqlite: Connection) : Cha
...
+//                update.setBytes(2, state.channelId.toByteArray())
+//                if (update.executeUpdate() == 0) {
+//                    using(sqlite.prepareStatement("INSERT INTO local_channels VALUES (?, ?, 0)")) { statement ->
+//                        statement.setBytes(1, state.channelId.toByteArray())
+//                        statement.setBytes(2, data)
+//                        statement.executeUpdate()
+//                    }
+//                }
+//            }
+//        }
```
t-bast, line 69: Why is that commented out?

t-bast (commented):
This was indeed a mouthful! I still need to spend more time on the tests, but this is a great clean-up IMO, this is definitely the right way to go :+1: 

The actual serialization / deserialization of channel states isn't that much code and is quite easy to review. I checked that we length-prefix everything that requires it (we even length-prefix a bit too much, but at least it keeps the code simple).

src/commonTest/kotlin/fr/acinq/lightning/channel/TestsHelper.kt
```diff
@@ -64,12 +60,84 @@ internal inline fun <reified T : ChannelAction> List<ChannelAction>.find() = fin
...
+
+    private fun checkSerialization(actions: List<ChannelAction>) {
+        // we check that serialization works everytime we're supposed to persist channel data
+        actions.filterIsInstance<ChannelAction.Storage.StoreState>().forEach { checkSerialization(it.data) }
+    }
+
+}
+
+// same as process but with added assumptions on exit state
+fun <T : ChannelState> LNChannel<T>.processSameState(event: ChannelCommand): Pair<LNChannel<T>, List<ChannelAction>> {
```
t-bast, line 135: nit: why isn't that function inside `LNChannel` (right below the `process` one)?

src/commonTest/kotlin/fr/acinq/lightning/channel/TestsHelper.kt
```diff
@@ -142,64 +217,64 @@ object TestsHelper {
         alicePushAmount: MilliSatoshi = TestConstants.alicePushAmount,
         bobPushAmount: MilliSatoshi = TestConstants.bobPushAmount,
         zeroConf: Boolean = false,
-    ): Triple<Normal, Normal, Transaction> {
+    ): Triple<LNChannel<Normal>, LNChannel<Normal>, Transaction> {
         val (alice, channelReadyAlice, bob, channelReadyBob) = WaitForChannelReadyTestsCommon.init(channelType, aliceFeatures, bobFeatures, currentHeight, aliceFundingAmount, bobFundingAmount, alicePushAmount, bobPushAmount, zeroConf)
         val (alice1, actionsAlice1) = alice.process(ChannelCommand.MessageReceived(channelReadyBob))
-        assertIs<Normal>(alice1)
+        assertIs<LNChannel<Normal>>(alice1)
```
t-bast, line 223: We could avoid repeating `LNChannel` everywhere by offering a helper method that does the same as `AssertIs` but for the channel state and can be directly called on an `LNChannel` (a slightly annoying find/replace once, but future test writing will be better).

src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationTestsCommon.kt
```diff
@@ -1,69 +1,67 @@
...
+        val check = ChannelStateWithCommitments.from(priv, bytes)
+        assertEquals(alice.state, check)
+
+        val bytes1 = EncryptedChannelData.from(priv, bob.state)
+        val check1 = ChannelStateWithCommitments.from(priv, bytes1)
+        assertEquals(bob.state, check1)
     }
-
+    
+    @Ignore
```
t-bast, line 43: We can probably completely remove that test?

src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationNonRegTestsCommon.kt
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Test
+    fun `non-reg test with v3 serialization`() {
```
t-bast, line 47: Wouldn't it make sense to also introduce test cases for v4 serialization right now?
That would let us easily check for example the json serialization of dual-funding components.

t-bast (commented):
The tests look good and the JSON results look nice!

src/commonMain/kotlin/fr/acinq/lightning/channel/Channel.kt
```diff
@@ -577,7 +554,7 @@ object Channel {
     // since BOLT 1.1, there is a max value for the refund delay of the main commitment tx
     val MAX_TO_SELF_DELAY = CltvExpiryDelta(2016)
 
-    fun handleSync(channelReestablish: ChannelReestablish, d: ChannelStateWithCommitments, keyManager: KeyManager, log: Logger): Pair<Commitments, List<ChannelAction>> {
+    fun ChannelContext.handleSync(channelReestablish: ChannelReestablish, d: ChannelStateWithCommitments, keyManager: KeyManager, log: Logger): Pair<Commitments, List<ChannelAction>> {
```
sstone, line 557: `ChannelContext` is not needed here

pm47:
> How were non-reg test vectors generated ? (i.e. using the serialization code in this branch or using the actual releases/tags used by released versions of Phoenix)

They were generated from tag `1.4.2`, by cherry-picking the json serialization and making all channel state tests write their bin/json serialized data to disk with `v2` and `v3`. I then selected samples for each state and used that on this branch.

In theory, `v2` should already have been read-only in `1.4.2` and I should have gone further back for that older version.

> What is the impact on Phoenix ? (I guess it will just need to use the new JSON serializers defined here)

The API is simplified, and json looks nicer, otherwise there shouldn't be much impact. There is a risk of deserialization error for channel data, that I think is limited due to the non-reg tests.

> Wouldn't it be easier to read and modify if serialization/deserialization code was in the same file and serializers/deserializers close to each other ?

I actually wrote the serializers first, then copy-pasted the file and wrote the deserializers. I quite like it like this but don't have a strong opinion. @t-bast WDYT?

src/commonMain/kotlin/fr/acinq/lightning/serialization/v2/ChannelState.kt
```diff
@@ -7,189 +45,242 @@ import fr.acinq.lightning.crypto.ShaChain
...
-    )
-
     fun export() = fr.acinq.lightning.channel.RevokedCommitPublished(commitTx, remotePerCommitmentSecret, claimMainOutputTx, mainPenaltyTx, htlcPenaltyTxs, claimHtlcDelayedPenaltyTxs, irrevocablySpent)
 }
 
 /**
- * README: by design, we do not include channel private keys and secret here, so they won't be included in our backups (local files, encrypted peer backup, ...), so even
- * if these backups were compromised channel private keys would not be leaked unless the main seed was also compromised.
+ * README: by design, we do not include channel internal keys and secret here, so they won't be included in our backups (local files, encrypted peer backup, ...), so even
+ * if these backups were compromised channel internal keys would not be leaked unless the main seed was also compromised.
```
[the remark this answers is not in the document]
pm47: 🦅 👁️

t-bast:
> I actually wrote the serializers first, then copy-pasted the file and wrote the deserializers. I quite like it like this but don't have a strong opinion. @t-bast WDYT?

I don't have a strong opinion either, to review them I opened both files side-by-side to match them line-by-line, that would work with either separate files or the same file.

src/commonMain/kotlin/fr/acinq/lightning/channel/Channel.kt
```diff
@@ -550,8 +521,14 @@ sealed class ChannelStateWithCommitments : ChannelState() {
             }
         }
     }
+
+    companion object {
+
+    }
```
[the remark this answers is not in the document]
pm47: That one is nasty. I can't remove the companion object because it is referenced indirectly by the external method in `Encryption.kt`:
```kotlin
fun ChannelStateWithCommitments.Companion.from(key: PrivateKey, encryptedChannelData: EncryptedChannelData): ChannelStateWithCommitments
```

I will add a comment. Alternatively I could have defined the external functions as normal functions in `ChannelStateWithCommitments` and in `EncryptedChannelData`, but I think it's better to put them at the same place.

Btw you make me realize that it should be `PersistedChannelState` here.

src/commonMain/kotlin/fr/acinq/lightning/serialization/Encryption.kt
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Convenience method that builds an [EncryptedChannelData] from a [ChannelStateWithCommitments]
+     */
+    fun EncryptedChannelData.Companion.from(key: PrivateKey, state: PersistedChannelState): EncryptedChannelData {
+        val bin = Serialization.serialize(state)
+        val encrypted = encrypt(key.value, bin)
+        // we copy the first 2 bytes as meta-info on the serialization version
+        val data = bin.copyOfRange(0, 2) + encrypted
```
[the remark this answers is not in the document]
pm47: I did think about this approach. The hacky part is that I'm copying _two_ bytes, which means that it includes the serialization version (1B), and the discriminator for the channel state (1B). Doing it "cleanly" would have made this creep a bit too deep in the serializers, and hurt the symmetry between serialization and deserialization. Dealing with bytes sort of guarantees that we're not coupling between the encryption layer and the serialization layer.

src/commonMain/kotlin/fr/acinq/lightning/json/JsonSerializers.kt
```diff
@@ -0,0 +1,437 @@
...
+                subclass(ClosingSignedTlv.FeeRange::class, ClosingSignedTlvFeeRangeSerializer)
+            }
+            contextual(PolymorphicSerializer(PersistedChannelState::class))
+            contextual(PolymorphicSerializer(UpdateMessage::class))
```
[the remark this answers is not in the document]
pm47: This _declares_ a polymorphic serializer for those types, whereas `polymorphic()` _registers_ subclasses of those types to be used by the `PolymorphicSerializer`.

src/jvmTest/kotlin/fr/acinq/lightning/Node.kt
pm47: I forgot 🤷‍♂️ ... but this is why those should be private in theory.

Given the half-finished state of this `Node`, I'm really tempted to nuke it for now, until we redo it better. WDYT? (cc @sstone)

src/commonMain/kotlin/fr/acinq/lightning/payment/PaymentRequest.kt
pm47: I confess I did a global replace 😇

src/jvmTest/kotlin/fr/acinq/lightning/db/sqlite/SqliteChannelsDb.kt
pm47: How are the tests passing? 😱 

Also, this should take `PersistendChannelState`

src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationNonRegTestsCommon.kt
pm47: My idea is to do this right before the next release, to let us freely change the model in the meantime.

src/commonTest/kotlin/fr/acinq/lightning/channel/TestsHelper.kt
pm47: Great idea, and would have saved me time on the first iteration. Will give it a try.

src/jvmTest/kotlin/fr/acinq/lightning/Node.kt
sstone: We could nuke it but it's still very useful for testing, and it just needs to be able to serialize channel states to JSON so that should still be possible with the changes in this PR ?

sstone:
> > I actually wrote the serializers first, then copy-pasted the file and wrote the deserializers. I quite like it like this but don't have a strong opinion. @t-bast WDYT?
> 
> I don't have a strong opinion either, to review them I opened both files side-by-side to match them line-by-line, that would work with either separate files or the same file.

Let's just keep them separate then

src/jvmTest/kotlin/fr/acinq/lightning/Node.kt
pm47: > it's still very useful for testing

I think this is much less true since we have automated integration testing, but I see your point.

> it just needs to be able to serialize channel states to JSON so that should still be possible with the changes in this PR ?

Yes it's possible, but it forces me to expose the `Json` object, which is a bit footgunny for callers because using external serializers cause way more runtime errors depending on how it is called. That's why I defined a well-tested `fun toJsonString(state: ChannelState): String` function that ensure the correct type is used. I'll add a warning and see how it looks.

sstone: Or you could just expose a `toJsonElement()` ?

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Serialization.kt
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v2/ChannelState.kt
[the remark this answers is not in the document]
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -624,7 +625,8 @@ class Peer(
                                 theirInit = msg
                                 _connectionState.value = Connection.ESTABLISHED
                                 _channels = _channels.mapValues { entry ->
-                                    val (state1, actions) = entry.value.process(ChannelCommand.Connected(ourInit, theirInit!!))
+
```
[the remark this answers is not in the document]
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/channel/WaitForAcceptChannel.kt
```diff
@@ -95,33 +87,30 @@ data class WaitForAcceptChannel(
...
             else -> unhandled(cmd)
         }
     }
 
-    override fun handleLocalError(cmd: ChannelCommand, t: Throwable): Pair<ChannelState, List<ChannelAction>> {
+    override fun ChannelContext.handleLocalError(cmd: ChannelCommand, t: Throwable): Pair<ChannelState, List<ChannelAction>> {
         logger.error(t) { "c:$temporaryChannelId error on event ${cmd::class} in state ${this::class}" }
-        val error = Error(init.temporaryChannelId, t.message)
-        return Pair(Aborted(staticParams, currentTip, currentOnChainFeerates), listOf(ChannelAction.Message.Send(error)))
+        val error = Error(init.temporaryChannelId(keyManager), t.message)
```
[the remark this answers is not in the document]
pm47: Done in c5365f4.

src/commonTest/kotlin/fr/acinq/lightning/channel/newmodel/NewChannel.kt
```diff
@@ -0,0 +1,51 @@
+package fr.acinq.lightning.channel.newmodel
```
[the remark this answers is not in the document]
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/channel/Channel.kt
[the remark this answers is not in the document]
pm47: Fixed in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/utils/collections.kt
```diff
@@ -14,3 +14,5 @@ fun <T1, T2, T3, C : MutableCollection<in Triple<T1, T2, T3>>> Triple<Iterable<T
 
 fun <T1, T2, T3> Triple<Iterable<T1>, Iterable<T2>, Iterable<T3>>.zipped(): List<Triple<T1, T2, T3>> =
     zippedTo(ArrayList())
+
+fun <K, V> Map<K, V>.reverse() = this.entries.associateBy({ it.value }) { it.key }
```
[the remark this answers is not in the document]
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/json/JsonSerializers.kt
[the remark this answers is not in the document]
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/json/JsonSerializers.kt
```diff
@@ -0,0 +1,437 @@
...
+            contextual(PolymorphicSerializer(DirectedHtlc::class))
+            // TODO following are serializers defined as @Contextual in project
+            //   classes, which is why they must be explicitly set here. Once the
+            //   @Contextual annotations have been removed this should be cleaned up
```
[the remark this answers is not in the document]
pm47: Tried a better comment in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/json/JsonSerializers.kt
```diff
@@ -0,0 +1,437 @@
...
+    object GenericTlvSerializer
+
+    @Serializable
+    data class TlvStreamSurrogate(val records: List<Tlv>, val unknown: List<GenericTlv> = listOf())
+    class TlvStreamSerializer<T : Tlv> : KSerializer<TlvStream<T>> {
+        private val delegateSerializer = TlvStreamSurrogate.serializer()
+        override val descriptor: SerialDescriptor = delegateSerializer.descriptor
+        override fun serialize(encoder: Encoder, value: TlvStream<T>) =
+            delegateSerializer.serialize(encoder, TlvStreamSurrogate(value.records, value.unknown))
+        override fun deserialize(decoder: Decoder): TlvStream<T> = TODO()
```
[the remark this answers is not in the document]
pm47: Done in c5365f4.

src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationNonRegTestsCommon.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.kodein.memory.text.readString
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class StateSerializationNonRegTestsCommon {
+
+    fun regtest(dir: String, debug: Boolean) {
+        val resourceDir = Environment.findVariable("TEST_RESOURCES_PATH")?.let { Path(it) } ?: FileSystem.workingDir().resolve("src/commonTest/resources")
+        val file = resourceDir.resolve("bolt3_anchor_outputs_test_vectors.json")
+        file.openReadableFile().run { readString(sizeBytes = remaining) }
```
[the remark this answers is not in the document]
pm47: Done in c5365f4.

src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationNonRegTestsCommon.kt
pm47: Done in c5365f4.

src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationNonRegTestsCommon.kt
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/payment/PaymentRequest.kt
pm47: Did a global replace to fix my first global replace (yodawg.jpg).

src/commonMain/kotlin/fr/acinq/lightning/wire/ChannelTlv.kt
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/wire/InteractiveTxTlv.kt
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/wire/MessageOnion.kt
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/wire/RouteBlinding.kt
pm47: Done in c5365f4.

src/jvmTest/kotlin/fr/acinq/lightning/db/sqlite/SqliteChannelsDb.kt
pm47: Done in c5365f4.

src/commonTest/kotlin/fr/acinq/lightning/channel/TestsHelper.kt
pm47: For no good reason, changed in c5365f4.

src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationTestsCommon.kt
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/channel/Channel.kt
pm47: Done in c5365f4.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Serialization.kt
```diff
@@ -0,0 +1,465 @@
...
+            writeBtcObject(it.key)
+            writeNullable(it.value) { claimHtlcTx -> writeTransactionWithInputInfo(claimHtlcTx) }
+        }
+        writeCollection(claimAnchorTxs) { writeTransactionWithInputInfo(it) }
+        writeIrrevocablySpent(irrevocablySpent)
+    }
+
+    private fun Output.writeRevokedCommitPublished(o: RevokedCommitPublished) = o.run {
+        writeBtcObject(commitTx)
+        write(remotePerCommitmentSecret.value.toByteArray())
```
sstone, line 193: nit: every time we use `Output.write()` directly  and not `writeDelimited()` it's because the value being serialized has a fixed, known size (32, 33, 64 bytes ...), why not add explicit methods for serializing/deserializing  Bytevector32/Bytevector64/etc... as using `Output.write()` directly is potentially dangerous ?

src/commonTest/kotlin/fr/acinq/lightning/channel/TestsHelper.kt
pm47: Hmm, for some reason this doesn't work:
```kotlin
inline fun <reified S : ChannelState> assertIsState(value: Any) = assertIs<LNChannel<S>>(value)
```
It doesn't do an implicit cast like the original `assertIs`.

src/jvmTest/kotlin/fr/acinq/lightning/Node.kt
pm47: I updated the `Node` in eede7fd4, but as expected the `GET /channels` method fails because it can't find a serializer for `ChannelState`.

src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationNonRegTestsCommon.kt
```diff
@@ -0,0 +1,54 @@
...
+
+class StateSerializationNonRegTestsCommon {
+
+    /**
+     * If test doesn't pass, set debug to `true`, run the test again and look for the `actual.json` file next to `data.json` in the resources. Then just
+     * compare the two json files (easy within IntelliJ) and see what the difference is.
+     */
+    fun regtest(dir: String, debug: Boolean) {
+        val resourceDir = Environment.findVariable("TEST_RESOURCES_PATH")?.let { Path(it) } ?: FileSystem.workingDir().resolve("src/commonTest/resources")
+        val file = resourceDir.resolve("bolt3_anchor_outputs_test_vectors.json")
```
t-bast, line 20: This is unused, removed?

src/commonMain/kotlin/fr/acinq/lightning/serialization/Encryption.kt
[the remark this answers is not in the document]
t-bast: Why do we need to include the discriminator for the channel state? In theory we only need the serialization version, right? We need to make this clearer and cleaner somehow.

> Doing it "cleanly" would have made this creep a bit too deep in the serializers, and hurt the symmetry between serialization and deserialization.

Why? If serialization returns a magic byte and the serialized data, and deserialization takes as input a magic byte (which lets us choose which deserializer version to use) and the serialized data, this is perfectly symmetrical and clean?

src/commonTest/kotlin/fr/acinq/lightning/channel/TestsHelper.kt
t-bast: > But instead of doing fancy stuff we could also simply go with:
> assertIs<Normal>(alice1.state)

If we do that we won't benefit from the Kotlin's smart-cast, will we? What's nice when doing `assertIs<Normal>(state)` is that `state` in smart-casted to `Normal` for the rest of the scope, but I don't think it works with nested arguments...

If this is too cumbersome that's something we can fix in a follow-up PR, it's not ideal but I don't want this to block the serialization update.

src/jvmTest/kotlin/fr/acinq/lightning/Node.kt
pm47: Phew, upped my json game a bit and got it to work in e6ae89b0. Unhandled states are serialized like so:
```
{
    "type": "fr.acinq.lightning.json.JsonSerializers.ChannelStateSurrogateSerializer",
    "forState": "fr.acinq.lightning.channel.WaitForOpenChannel"
}
```

Note that calling `JsonSerializers.json.encodeToString(state: Normal)` will fail due to [static parent type lookup](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/polymorphism.md#static-parent-type-lookup-for-polymorphism). This is what I was initially afraid of, but I think it's fair game, because we always want the `type` attribute. Any of `JsonSerializers.json.encodeToString(state: ChannelState)` or `JsonSerializers.json.encodeToString(state: ChannelStateWithCommitments)` or `JsonSerializers.json.encodeToString(state: PersistedChannelState)` work now.

⚠️ Just noticed that `Offline` and `Syncing` are `ChannelState`, not `ChannelStateWithCommitments`. That's another design flaw, due to the fact that those states are not persisted. But the confusion has been cleared up with `PersistedChannelState`. I'm afraid of changing the type though, who knows what it can break.

src/commonTest/kotlin/fr/acinq/lightning/channel/TestsHelper.kt
pm47: > If we do that we won't benefit from the Kotlin's smart-cast, will we? What's nice when doing assertIs<Normal>(state) is that state in smart-casted to Normal for the rest of the scope, but I don't think it works with nested arguments...

It does, actually. But the way we then use the object make us lose this cast unfortunately, so the change isn't trivial. Let's leave it like this for now.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Serialization.kt
pm47: Done in [bb24e14](https://github.com/ACINQ/lightning-kmp/pull/391/commits/bb24e14ae59fe4d021bd2749802d721aaa111bdc).

src/commonMain/kotlin/fr/acinq/lightning/serialization/Encryption.kt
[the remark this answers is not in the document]
pm47: I gave it a try in 8f35d013, but I think it's not better:
- this meta-info now impacts 3 files instead of one, and adds coupling between those files
- the way I build `SerializedChannelState(bytes[0], bytes[1], bytes)` is still hacky because the data has already been serialized at this point, we cherry pick bytes from the result
- we still need to do the ugly `data.drop(2)` when decrypting
- due to backward compat with previous versions that didn't have the meta-info, we break the symmetry between the `serialize()` and `deserialize()` methods
- I also tried not writing the version byte in `v4/Serialization`, and it started nice, but then it got hacky at the upper layer

Am I missing something?

t-bast (dismissed):

src/commonMain/kotlin/fr/acinq/lightning/serialization/Encryption.kt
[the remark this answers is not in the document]
t-bast: > this meta-info now impacts 3 files instead of one, and adds coupling between those files

I think that coupling between serialization and deserialization is natural, and the coupling with encryption is there anyway since we pad the encrypted result with that meta...

> the way I build SerializedChannelState(bytes[0], bytes[1], bytes) is still hacky because the data has already been serialized at this point, we cherry pick bytes from the result

This is only true for the `stateByte`, not the version one, which is a static value of the serializer. We know that including the state byte is somewhat hackish anyway but may come in handy!

> due to backward compat with previous versions that didn't have the meta-info, we break the symmetry between the serialize() and deserialize() methods

That's because we still do a hacky deserialization method where we try each version in sequence and catch errors. A better deserialization method would start by popping the meta bytes then call the right version decrypter and deserializer depending on the version bytes. But that can be done in a future improvement.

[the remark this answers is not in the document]
pm47: As discussed offline, I reverted in 9c528ccd as I think the fix is still unsatisfactory. This doesn't affect serialization and can be revisited in a later PR.

t-bast (approved):

src/jvmTest/kotlin/fr/acinq/lightning/Node.kt
```diff
@@ -231,10 +232,7 @@ object Node {
                     }
                 }
                 install(ContentNegotiation) {
-                    json(Json {
-                        serializersModule = Serialization.lightningSerializersModule
-                        allowStructuredMapKeys = true
-                    })
+                    json(JsonSerializers.json)
```
sstone, line 235: It seems that now API calls to `/channels` work but not calls to `/channels/{channelId}`:

```
Serializer for class 'Normal' is not found.
Mark the class as @Serializable or provide the serializer explicitly.
```

pm47: There are no tests on this thing, I should manually test all endpoints I guess.

sstone: Fine with me, it was clumsy to maintain (copying parameters back from phoenix,...) and I was the only one to use it. It means we can also remove `ktor` from our JVM test dependencies.

pm47: I cleaned up dependencies in [09e4957](https://github.com/ACINQ/lightning-kmp/pull/391/commits/09e4957da11e7f087e2f13c303f46e51d8a285b0).

sstone (approved):

t-bast (approved):
:zap: :rocket:

pm47:
🎉

PR #392, opened 2022-10-21
Dual-funding uses txHash and verifies nSequence

t-bast:
This PR contains independent commits that were added to the `eclair` implementation:

- use `tx_hash` instead of `tx_id` in `tx_signatures`
- verify the `nSequence` field of remote `tx_add_input`


pm47:
Do you think we could split and merge the existing fixes, like on eclair? So endurance and kmp/master would be in sync

t-bast:
Yes, that makes sense, especially since the spec discussion on `tx_signatures` ordering is lagging behind. I'll do that shortly.

pm47 (approved):

PR #393, opened 2022-10-25
Allow HTLC over-payment and over-expiry

t-bast:
Recipients must allow HTLC values greater than what the onion payload contains, otherwise they can be trivially identified as being the final recipient.

See https://github.com/lightning/bolts/pull/1032 for more details.


pm47 (approved):

t-bast:
Rebased.

PR #394, opened 2022-10-25
Randomize final expiry

t-bast:
When sending outgoing payments, using an expiry that is very close to the current block height makes it obvious to the next-to-last node who the recipient is.

We add a random expiry to the amount we use to make it plausible that the route contains more hops.

Core Lightning and LDK already do this, LND doesn't.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentHandler.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentHandler.kt
@@ -336,8 +336,8 @@ class OutgoingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
             )
         }
 
-        val finalExpiryDelta = request.paymentRequest.minFinalExpiryDelta ?: Channel.MIN_CLTV_EXPIRY_DELTA
-        val finalExpiry = finalExpiryDelta.toCltvExpiry(currentBlockHeight.toLong())
+        val minFinalExpiryDelta = request.paymentRequest.minFinalExpiryDelta ?: Channel.MIN_CLTV_EXPIRY_DELTA
+        val finalExpiry = nodeParams.paymentRecipientExpiryParams.computeFinalExpiry(currentBlockHeight, minFinalExpiryDelta)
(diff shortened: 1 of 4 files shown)


src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentHandler.kt
```diff
@@ -337,7 +337,8 @@ class OutgoingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
         }
 
         val finalExpiryDelta = request.paymentRequest.minFinalExpiryDelta ?: Channel.MIN_CLTV_EXPIRY_DELTA
-        val finalExpiry = finalExpiryDelta.toCltvExpiry(currentBlockHeight.toLong())
+        // We add a random expiry to ensure that the next-to-last node cannot trivially identify the recipient.
+        val finalExpiry = finalExpiryDelta.toCltvExpiry(currentBlockHeight.toLong() + Lightning.secureRandom.nextLong(100, 300))
```
dpad85, line 341: Using a purely random value could be telling as well, if nodes in the network typically use fixed defined values.

dpad85 (dismissed):
LGTM, though I'm not sure about what random values should be added. Let me know if you need end-to-end tests on Phoenix.

src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentHandler.kt
t-bast: I think that the best way to do this is to add the cltv expiry delta of real hops that are connected to the destination, but that requires access to the graph...I think I'll make those values configurable in `NodeParams`, I'm not a big fan of having them hardcoded like this.

t-bast: I made that configurable in https://github.com/ACINQ/lightning-kmp/pull/394/commits/52cb75d9598b985efd030fe18fcdf193b8d15d0e and I'm happier with the result, it also lets us disable the randomization entirely if needed.

I think that at first Phoenix should use the values from `Node.kt`, which should provide good enough deniability.

pm47 (approved):

PR #396, opened 2022-10-31
WalletState.balance: clarify confirmed vs unconfirmed balance

robbiehanson:
Previously, WalletState had a `var balance: Satoshi` property. However this didn't distinguish between confirmed vs unconfirmed UTXOs.

This was troublesome for the dual-funding flow, which is started like this:

```kotlin
if (wallet.balance > 10_000.sat) {
  // Start swap-in if not already started
}
```

Because initiating the Dual-Funding flow with unconfirmed inputs would ultimately fail, with the server sending a `TxAbort` containing the message:

> "the completed interactive tx contains unconfirmed inputs"

This PR modifies the WalletState, allowing for improved inspection of the balance:

```kotlin
fun balance(includingUnconfirmed: Boolean): Satoshi {
  return if (includingUnconfirmed) {
    utxos.map { it.amount }.sum()
  } else {
    utxos.filter { it.blockHeight > 0L }.map { it.amount }.sum()
  }
}

fun unconfirmedBalance(): Satoshi {
  return utxos.filter { it.blockHeight == 0L }.map { it.amount }.sum()
}
```

And the Peer is then updated accordingly:

```kotlin

val balance = wallet.balance(includingUnconfirmed = false)
if (balance > 10_000.sat) {
  // Start swap-in if not already started
}
```

PR #395, opened 2022-10-28
Add `fundingFee` to `IncomingPayment.ReceivedWith.NewChannel`

robbiehanson:
With the addition of DualFunding, there are now 2 separate fees as shown in `SwapInEvents.Accepted`:

```kotlin
data class Accepted(
  val requestId: ByteVector32,
  val fundingFee: Satoshi, // L1 fees paid to miner
  val serviceFee: MilliSatoshi // L2 fees paid to lightning service provider
) : SwapInEvents
```

However, the fundingFee wasn't exported to `IncomingPayment.ReceivedWith.NewChannel`. Only the serviceFee was included. So, for example, a user sends 15,000 sat to their wallet, and we show:

- received: 11,931 sat
- fees: 3,000 sat

This PR adds both fees to the class:

```kotlin
data class NewChannel(
  val id: UUID,
  override val amount: MilliSatoshi,
  val serviceFee: MilliSatoshi,
  val fundingFee: Satoshi = 0.sat,
  val channelId: ByteVector32?
) : ReceivedWith() {
  override val fees: MilliSatoshi = serviceFee + fundingFee.toMilliSatoshi()
}
```

This allows the UI to properly display each fee to the user. And we extract the fundingFee here:

```kotlin
// oldState is WaitForFundingSigned
when {
  !oldState.localParams.isInitiator -> {
    val amount = oldState.fundingParams.localAmount.toMilliSatoshi() + oldState.remotePushAmount - oldState.localPushAmount
    val fundingFee = oldState.fundingTx.localFees(oldState.fundingParams)
    val localInputs = oldState.fundingTx.localInputs.map { OutPoint(it.previousTx, it.previousTxOutput) }.toSet()
    actions + ChannelAction.Storage.StoreIncomingAmount(amount, fundingFee, localInputs, oldState.channelOrigin) + channelCreated
  }
  else -> actions + channelCreated
}
```

PR #397, opened 2022-11-02
Peer only allowed 1 swap-in per app launch

robbiehanson:
This PR fixes 2 separate (but related) bugs.

### Bug 1:

Previously, the Peer was structured like this:

```kotlin
launch {
  swapInWallet.walletStateFlow
    .filter { it.consistent }
    .fold(false) { swapAlreadyAttempted, wallet ->
        if (wallet.confirmedBalance > 10_000.sat) {
          if (swapAlreadyAttempted) {
            // swap-in already attempted: not doing anything
          } else {
            // send command RequestChannelOpen ...
          }
          true // ^fold
        } else {
          // balance insufficient: waiting for more
          swapAlreadyAttempted // ^fold
        }
    }
}
```

The end result is that the Dual-Funding flow is only initiated once per app launch. Because once the `swapAlreadyAttempted` flag is set, it is never un-set.

### Bug 2:

In PR #396, we fixed a bug related to the confirmed vs unconfirmed balance. We noted that:

> initiating the Dual-Funding flow with unconfirmed inputs would ultimately fail, with the server sending a `TxAbort` message

And although we no longer start the Dual-Funding flow until the `confirmedBalance` is sufficient, we still use the FULL set of `utxos`, regardless of whether or not these are confirmed or unconfirmed.

For example, in `WaitForAcceptChannel`:

```kotlin
val fundingContributions = FundingContributions.create(fundingParams, init.wallet.utxos) // <- ALL utxos
```

If the user is attempting to move funds into Phoenix, there may be a mixture of confirmed vs unconfirmed utxos. And although we wait until `confirmedBalance` is sufficient, this doesn't guarantee there aren't any unconfirmed utxos in the set.

### Proposed solution

First we expose the `confirmedUtxos` in `ElectrumMiniWallet`:

```kotlin
val confirmedUtxos: List<Utxo> = utxos.filter { it.blockHeight > 0L }
val unconfirmedUtxos: List<Utxo> = utxos.filter { it.blockHeight == 0L }
```

And use only `confirmedUtxos` when appropriate:

```kotlin
val fundingContributions = FundingContributions.create(fundingParams, init.wallet.confirmedUtxos)
```

Next we fix the Peer with the following logic:

- when we start a swap-in, what we want to do is **reserve** a set of utxos for the dual-funding flow
- we keep track of this reserved `Set<Utxo>`, and exclude them from consideration when calculating the balance

So it looks like this:

```kotlin
launch {
  swapInWallet.walletStateFlow
    .filter { it.consistent }
    .fold(setOf<WalletState.Utxo>()) { reservedUtxos, wallet ->
      // reservedUtxos are part of a previously issued RequestChannelOpen command
      val availableWallet = wallet.drop(reservedUtxos)
      val balance = availableWallet.confirmedBalance
      if (balance > 10_000.sat) {
        logger.info { "$balance available on swap-in wallet: requesting channel" }
        input.send(RequestChannelOpen(uuid, availableWallet, other, params))
        reservedUtxos + availableWallet.confirmedUtxos // ^fold
      } else {
        logger.info { "$balance available on swap-in wallet but amount insufficient: waiting for more" }
        reservedUtxos.intersect(wallet.utxos) // drop items from reservedUtxos that are no longer in wallet
      }
    }
}
```

So by calling `wallet.drop(reservedUtxos)` we're creating a WalletState subset. Which we then pass to `RequestChannelOpen` ensuring that it will only use what we've newly reserved for it.

And the `reservedUtxos` doesn't grow forever, as we drop items as they get removed from the wallet. I.e. when we spend those Utxos.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
@@ -24,19 +24,30 @@ data class WalletState(val addresses: Map<String, List<UnspentItem>>, val parent
         .filter { parentTxs.containsKey(it.txid) }
         .map { Utxo(parentTxs[it.txid]!!, it.outputIndex, it.blockHeight) }
 
-    val confirmedBalance = balance(includingUnconfirmed = false, includingConfirmed = true)
-    val unconfirmedBalance = balance(includingUnconfirmed = true, includingConfirmed = false)
+    val confirmedUtxos: List<Utxo> = utxos.filter { it.blockHeight > 0L }
+    val unconfirmedUtxos: List<Utxo> = utxos.filter { it.blockHeight == 0L }
+
+    val confirmedBalance = confirmedUtxos.map { it.amount }.sum()
+    val unconfirmedBalance = unconfirmedUtxos.map { it.amount }.sum()
     val totalBalance = confirmedBalance + unconfirmedBalance
-    private fun balance(includingUnconfirmed: Boolean, includingConfirmed: Boolean): Satoshi =
-        utxos
-            .filter { (includingUnconfirmed && it.blockHeight == 0L) || (includingConfirmed && it.blockHeight > 0L) }
-            .map { it.amount }
-            .sum()
+
+    fun drop(reserved: Set<Utxo>): WalletState {
+        val reservedIds = reserved.map {
+            UnspentItemId(it.previousTx.txid, it.outputIndex)
+        }.toSet()
+        return copy(addresses = addresses.mapValues {
+            it.value.filter { item ->
+                !reservedIds.contains(UnspentItemId(item.txid, item.outputIndex))
+            }
+        })
+    }
 
     data class Utxo(val previousTx: Transaction, val outputIndex: Int, val blockHeight: Long) {
         val outPoint = OutPoint(previousTx, outputIndex.toLong())
         val amount = previousTx.txOut[outputIndex].amount
     }
+
+    data class UnspentItemId(val txid: ByteVector32, val outputIndex: Int)
 
     companion object {
         val empty: WalletState = WalletState(emptyMap(), emptyMap())
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
@@ -31,7 +31,7 @@ data class WalletState(val addresses: Map<String, List<UnspentItem>>, val parent
     val unconfirmedBalance = unconfirmedUtxos.map { it.amount }.sum()
     val totalBalance = confirmedBalance + unconfirmedBalance
 
-    fun drop(reserved: Set<Utxo>): WalletState {
+    fun minus(reserved: Set<Utxo>): WalletState {
         val reservedIds = reserved.map {
             UnspentItemId(it.previousTx.txid, it.outputIndex)
         }.toSet()
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -223,19 +223,17 @@ class Peer(
         launch {
             swapInWallet.walletStateFlow
                 .filter { it.consistent }
-                .fold(false) { swapAlreadyAttempted, wallet ->
-                    val balance = wallet.confirmedBalance
+                .fold(setOf<WalletState.Utxo>()) { reservedUtxos, wallet ->
+                    // reservedUtxos are part of a previously issued RequestChannelOpen command
+                    val availableWallet = wallet.drop(reservedUtxos)
+                    val balance = availableWallet.confirmedBalance
                     if (balance > 10_000.sat) {
-                        if (swapAlreadyAttempted) {
-                            logger.info { "$balance available on swap-in wallet but swap-in already attempted: not doing anything" }
-                        } else {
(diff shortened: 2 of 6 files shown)


pm47 (commented):
Bug 1 was actually a known limitation, due the potential race condition on utxos. I should have made it more explicit clearer, _mea culpa_. But you did take this race condition into account and worked around it elegantly in the `fold`. Nice work!

pm47:
Nit, that we could fix in this pass:
> 15:19:42.980 INFO  Peer - 0 sat available on swap-in wallet but amount insufficient: waiting for more

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -223,19 +223,17 @@ class Peer(
...
-                            input.send(RequestChannelOpen(Lightning.randomBytes32(), wallet, maxFeeBasisPoints = 100, maxFeeFloor = 3_000.sat)) // 100 bips = 1 %
-                        }
-                        true
+                        logger.info { "$balance available on swap-in wallet: requesting channel" }
+                        input.send(RequestChannelOpen(Lightning.randomBytes32(), availableWallet, maxFeeBasisPoints = 100, maxFeeFloor = 3_000.sat)) // 100 bips = 1 %
+                        reservedUtxos + availableWallet.confirmedUtxos
                     } else {
                         logger.info { "$balance available on swap-in wallet but amount insufficient: waiting for more" }
-                        swapAlreadyAttempted
+                        reservedUtxos.intersect(wallet.utxos) // drop items from reservedUtxos that are no longer in wallet
```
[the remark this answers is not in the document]
robbiehanson: Good point. Done in [20bdc18](https://github.com/ACINQ/lightning-kmp/pull/397/commits/20bdc188eda0230ff7ee248a0fa1b263e8897683)

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -24,20 +24,31 @@ data class WalletState(val addresses: Map<String, List<UnspentItem>>, val parent
...
+    val confirmedBalance = confirmedUtxos.map { it.amount }.sum()
+    val unconfirmedBalance = unconfirmedUtxos.map { it.amount }.sum()
     val totalBalance = confirmedBalance + unconfirmedBalance
-    private fun balance(includingUnconfirmed: Boolean, includingConfirmed: Boolean): Satoshi =
-        utxos
-            .filter { (includingUnconfirmed && it.blockHeight == 0L) || (includingConfirmed && it.blockHeight > 0L) }
-            .map { it.amount }
-            .sum()
+
+    fun drop(reserved: Set<Utxo>): WalletState {
```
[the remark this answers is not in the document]
robbiehanson: > due to the structure of WalletState which has been designed to work easily with electrum

I thought the structure of WalletState was quite clever, given that it's trying to deal with **asynchronous** Electrum requests. It does what it must, in a rather readable form, and presents the usual suspects to external users (`List<Utxo>` & `balance`)

> I'm not 100% happy with the idea of creating subsets of the `WalletState` class. In fact, we only really need a `List<Utxo>`

Exactly. That was my first idea, but then when I started implementing it I ran into this problem:

> Unfortunately this has impacts on the rest of the codebase and going with this would clash with #391

Yup. A bunch of classes explicitly take a `WalletState` parameter. Given the new approach (of reserving Utxos) it would make more sense to refactor these classes, and instead pass `List<Utxo>`. But I agree with you, it's probably better to revisit this later, after all the conflicting in-flight PRs are closed.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -223,20 +223,18 @@ class Peer(
...
-                        }
-                        true
+                        logger.info { "$balance available on swap-in wallet: requesting channel" }
+                        input.send(RequestChannelOpen(Lightning.randomBytes32(), availableWallet, maxFeeBasisPoints = 100, maxFeeFloor = 3_000.sat)) // 100 bips = 1 %
+                        reservedUtxos.union(availableWallet.confirmedUtxos)
                     } else {
-                        logger.info { "$balance available on swap-in wallet but amount insufficient: waiting for more" }
-                        swapAlreadyAttempted
-                    }
+                        logger.info { "insufficient balance on swap-in wallet ($balance, ${availableWallet.unconfirmedBalance} unconfirmed): waiting for more" }
```
pm47, line 235: It's gonna sound like I'm bikeshedding, but the issue I have with this log is that it makes it implies there is something wrong with having 0.sat ("insufficient balance"), whereas this is not only normal, but the most common case.

robbiehanson: Ah, I didn't understand what you meant initially. That makes sense.

I took another stab at it in [a9be848](https://github.com/ACINQ/lightning-kmp/pull/397/commits/a9be84817ffa36c4716ec4ac18eb93b782e87b40). I actually didn't like the fact that it only logged the full balance (confirmed + unconfirmed) in the `else` branch. So I moved the logs around:

```kotlin
val balance = availableWallet.confirmedBalance
logger.info { "swap-in wallet balance: $balance, ${availableWallet.unconfirmedBalance} unconfirmed" }
if (balance > 10_000.sat) {
    logger.info { "swap-in wallet: requesting channel using confirmed balance: $balance" }
    // ... code ...
} else {
    // ... code ...
}
```

What do you think ?

pm47: 👍

pm47 (approved):

pm47:
We're close to merging #391, we should be able to merge this and #395 early next week

pm47:
This can now be merged.

PR #398, opened 2022-11-07
Update visibility of `privateKey` methods to public

dpad85:
In Phoenix we need to create several custom keys (like the lnurl-auth key or the iCloud encryption key). Those keys are derived from the wallet `master` key. So we have to access `master` key from Phoenix.

To do so we make the `privateKey` methods public in the `LocalKeyManager` class. Those methods derive `master` given a a key path.

Note : the key manager is instantiated by the library client (in our case, Phoenix). Any client will have access to the key manager.

Note 2 : we could implement the logic for all custom keys in lightning-kmp, but I don't think this is a good solution as this would pollute the library with things unrelated to lightning.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
@@ -10,7 +10,7 @@ import fr.acinq.lightning.transactions.Transactions
 
 data class LocalKeyManager(val seed: ByteVector, val chainHash: ByteVector32) : KeyManager {
 
-    val master = DeterministicWallet.generate(seed)
+    private val master = DeterministicWallet.generate(seed)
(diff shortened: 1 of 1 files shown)


src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
```diff
@@ -10,7 +10,7 @@ import fr.acinq.lightning.transactions.Transactions
 
 data class LocalKeyManager(val seed: ByteVector, val chainHash: ByteVector32) : KeyManager {
 
-    private val master = DeterministicWallet.generate(seed)
+    val master = DeterministicWallet.generate(seed)
```
sstone, line 13: We could instead make ` privateKey(keyPath: KeyPath)` and `fun publicKey(keyPath: KeyPath)` public ?

dpad85: Fixed in da16e221

It is more consistent to make the `privateKey` methods public.

pm47 (approved):

PR #400, opened 2022-11-09
Put back deserialization of  `FundingCreated`/`FundingSigned`

pm47:
Deserializers for messages `FundingCreated` and `FundingSigned` were removed in dd6c84cb5ec84ebfec222f67270197be92af80f0 as part of implementing dual funding.

But we now use LN protocol format in v4 serialization. This means that we need to be able to read/write all LN messages that are persisted. Even if those messages have been deprecated by dual funding, we may have existing channels in the process of being created at the time when the user upgrades Phoenix. This is the purpose of
`LegacyWaitForFundingConfirmed`/`LegacyWaitForFundingLocked`. If these states are written back to disk, they will use v4 format, therefore we need to be able to serialize and deserialize all LN messages that are part of those states, which includes `FundingCreated` and `FundingSigned`.

Why wasn't this caught in tests?

Channel state tests do a serialization roundtrip at each state transition, but we never serialize
`LegacyWaitForFundingConfirmed`, it is only read and then we move to the next state. To be fair, in reality there is probably no code path where we would need to actually serialize this state, but this seems a bit risky to take chances for such a little change.

Static non-reg tests only deserializes data and compare it to reference json. This PR also adds a serialization roundtrip test, which succesfully reproduces the issue.

Why is the error in the v2 serialization whereas the bug affects v4?

Because of the try-and-fallback logic that we used for deserializing data. All deserialization failures end up with an error in the last codec in the list, which happens to be v2 (cc @robbiehanson). This logic has been changed and we now explicitly try the correct serialization based on the version discriminator.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/serialization/Serialization.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/serialization/Serialization.kt
@@ -3,7 +3,6 @@ package fr.acinq.lightning.serialization
 import fr.acinq.lightning.channel.PersistedChannelState
 import fr.acinq.lightning.serialization.v4.Deserialization.fromBinV4
 import fr.acinq.lightning.serialization.v4.Serialization.toBinV4
-import fr.acinq.lightning.utils.runTrying
 
 object Serialization {
 
@@ -12,13 +11,14 @@ object Serialization {
     }
 
     fun deserialize(bin: ByteArray): PersistedChannelState {
-        return runTrying {
-            bin.fromBinV4()
-        }.recoverWith {
-            runTrying { fr.acinq.lightning.serialization.v3.Serialization.deserialize(bin) }
-        }.recoverWith {
-            runTrying { fr.acinq.lightning.serialization.v2.Serialization.deserialize(bin) }
-        }.get()
+        return when {
+            // v4 uses a 1-byte version discriminator
+            bin[0].toInt() == 4 -> bin.fromBinV4()
+            // v2/v3 use a 4-bytes version discriminator
+            bin.sliceArray(0..3).contentEquals(byteArrayOf(0, 0, 0, 3)) -> fr.acinq.lightning.serialization.v3.Serialization.deserialize(bin)
+            bin.sliceArray(0..3).contentEquals(byteArrayOf(0, 0, 0, 2)) -> fr.acinq.lightning.serialization.v2.Serialization.deserialize(bin)
+            else -> error("unknown serialization version")
+        }
     }
 
 }
--- a/src/commonMain/kotlin/fr/acinq/lightning/serialization/Serialization.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/serialization/Serialization.kt
@@ -1,5 +1,6 @@
 package fr.acinq.lightning.serialization
 
+import fr.acinq.bitcoin.crypto.Pack
 import fr.acinq.lightning.channel.PersistedChannelState
 import fr.acinq.lightning.serialization.v4.Deserialization.fromBinV4
 import fr.acinq.lightning.serialization.v4.Serialization.toBinV4
@@ -15,8 +16,8 @@ object Serialization {
             // v4 uses a 1-byte version discriminator
             bin[0].toInt() == 4 -> bin.fromBinV4()
             // v2/v3 use a 4-bytes version discriminator
-            bin.sliceArray(0..3).contentEquals(byteArrayOf(0, 0, 0, 3)) -> fr.acinq.lightning.serialization.v3.Serialization.deserialize(bin)
-            bin.sliceArray(0..3).contentEquals(byteArrayOf(0, 0, 0, 2)) -> fr.acinq.lightning.serialization.v2.Serialization.deserialize(bin)
+            Pack.int32BE(bin) == 3 -> fr.acinq.lightning.serialization.v3.Serialization.deserialize(bin)
+            Pack.int32BE(bin) == 2  -> fr.acinq.lightning.serialization.v2.Serialization.deserialize(bin)
             else -> error("unknown serialization version")
         }
     }
--- a/src/commonMain/kotlin/fr/acinq/lightning/serialization/Serialization.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/serialization/Serialization.kt
@@ -2,22 +2,20 @@ package fr.acinq.lightning.serialization
 
 import fr.acinq.bitcoin.crypto.Pack
 import fr.acinq.lightning.channel.PersistedChannelState
-import fr.acinq.lightning.serialization.v4.Deserialization.fromBinV4
-import fr.acinq.lightning.serialization.v4.Serialization.toBinV4
 
 object Serialization {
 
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/fr/acinq/lightning/serialization/Serialization.kt
```diff
@@ -12,13 +11,14 @@ object Serialization {
...
-            runTrying { fr.acinq.lightning.serialization.v2.Serialization.deserialize(bin) }
-        }.get()
+        return when {
+            // v4 uses a 1-byte version discriminator
+            bin[0].toInt() == 4 -> bin.fromBinV4()
+            // v2/v3 use a 4-bytes version discriminator
+            bin.sliceArray(0..3).contentEquals(byteArrayOf(0, 0, 0, 3)) -> fr.acinq.lightning.serialization.v3.Serialization.deserialize(bin)
+            bin.sliceArray(0..3).contentEquals(byteArrayOf(0, 0, 0, 2)) -> fr.acinq.lightning.serialization.v2.Serialization.deserialize(bin)
+            else -> error("unknown serialization version")
+        }
```
pm47, line 21: This is quite ugly, I don't know how to do it better.

We could make v4 use 4 bytes for its version to be a bit more consistent, but I don't really see the point of having 4+ billion possible serialization versions.

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -50,6 +50,8 @@ interface LightningMessage {
                 Pong.type -> Pong.read(stream)
                 OpenDualFundedChannel.type -> OpenDualFundedChannel.read(stream)
                 AcceptDualFundedChannel.type -> AcceptDualFundedChannel.read(stream)
+                FundingCreated.type -> FundingCreated.read(stream)
+                FundingSigned.type -> FundingSigned.read(stream)
```
pm47, line 54: Actual fix.

src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationNonRegTestsCommon.kt
```diff
@@ -36,6 +38,20 @@ class StateSerializationNonRegTestsCommon {
...
+                    is WaitForChannelReady -> state.copy(commitments = state.commitments.copy(remoteChannelData = EncryptedChannelData.empty))
+                    is Normal -> state.copy(commitments = state.commitments.copy(remoteChannelData = EncryptedChannelData.empty))
+                    is ShuttingDown -> state.copy(commitments = state.commitments.copy(remoteChannelData = EncryptedChannelData.empty))
+                    is Negotiating -> state.copy(commitments = state.commitments.copy(remoteChannelData = EncryptedChannelData.empty))
+                    is Closing -> state.copy(commitments = state.commitments.copy(remoteChannelData = EncryptedChannelData.empty))
+                    is WaitForRemotePublishFutureCommitment -> state.copy(commitments = state.commitments.copy(remoteChannelData = EncryptedChannelData.empty))
+                    is Closed -> state
+                }
+                val state2 = Serialization.deserialize(Serialization.serialize(state))
+                assertEquals(state1, state2, path.toString())
```
pm47, line 54: More uglyness. We can't just compare `state` with `state2` because test data was generated using `lightning-kmp` integration tests, which have two Phoenix peers, which cause them to both have a `remoteChannelData` (which will be dropped at serialization).

I could regenerate the test data, or simpler (and maybe better?): serialize `remoteChannelData`. Why not after all?

src/commonMain/kotlin/fr/acinq/lightning/serialization/Serialization.kt
sstone: maybe something like:
```scala
Pack.int32BE(bin) == 3 -> fr.acinq.lightning.serialization.v3.Serialization.deserialize(bin)
```

pm47: That looks better, done in [1d39974](https://github.com/ACINQ/lightning-kmp/pull/400/commits/1d39974c91b897b40c84fa1fb4ee50e63b95d995).

t-bast:
> If these states are written back to disk, they will use v4 format, therefore we need to be able to serialize and deserialize all LN messages that are part of those states, which includes FundingCreated and FundingSigned.

I don't think that is true. `LegacyWaitForFundingConfirmed` does contain `FundingCreated` and `FundingSigned`, but `LegacyWaitForFundingLocked` doesn't. There are no code paths where we actually serialize `LegacyWaitForFundingConfirmed`, so we shouldn't need serializers for `FundingCreated` and `FundingSigned`, we only need to be able to deserialize them in v2 and v3 serialization.

The only flows that are possible are:

- The channel was in `LegacyWaitForFundingLocked` before the app update:
  - It will be correctly deserialized by v2 or v3 serialization
  - The only state that we may then serialize is `Normal` once we receive the remote `channel_ready`
- The channel was in `LegacyWaitForFundingConfirmed` before the app update:
  - It will be correctly deserialized by v2 or v3 serialization
  - When the funding tx confirms, we go to `LegacyWaitForFundingLocked` and serialize it using v4: it's ok because it doesn't contain either `FundingCreated` or `FundingSigned`
  - Then we're back the first case

I like the change to the discriminator in `Serialization.kt`, but I think the other changes are unnecessary.

pm47:
> There are no code paths where we actually serialize LegacyWaitForFundingConfirmed, so we shouldn't need serializers for FundingCreated and FundingSigned

I agree with this assessment, I came to the same conclusion, but I don't see a very good reason to voluntarily remove our ability to deserialize those two standard LN messages, if only for the sake of symmetry with other messages.

> I like the change to the discriminator in Serialization.kt, but I think the other changes are unnecessary.

I reverted the changed to the static non-reg test in c89b4dbe. Wdyt about serializing `remoteChannelData` in v4?

t-bast:
> I don't see a very good reason to voluntarily remove our ability to deserialize those two standard LN messages

I agree, we can keep those around for a while :+1: 

> Wdyt about serializing remoteChannelData in v4?

I thought about this during #391, I think we should do it if it isn't too much hassle.

pm47:
>> Wdyt about serializing remoteChannelData in v4?

>I thought about this during https://github.com/ACINQ/lightning-kmp/pull/391, I think we should do it if it isn't too much hassle.

Done in efec2344. Not only is it very easy, it also removes the same hack I had to do for the round-trip test in channels tests. It is now so straightforward I put back the round-trip test in the static non-reg test as well.

t-bast (approved):

pm47:
Now that I think of it, maybe the proper fix was to remove serialization of the whole `LegacyWaitForFundingConfirmed` state. We only need to write `LegacyWaitForFundingLocked`. Well...

src/commonMain/kotlin/fr/acinq/lightning/serialization/Serialization.kt
pm47: Improved further in d5ad4a81. Not sure why I initially diverged on naming.

t-bast (approved):

PR #403, opened 2022-11-10
Emit `Confirmed` event when channel reaches normal

pm47:
When a new channel has been created, it is not immediately usable for sending. Due to previously always using zero-conf, the time during which the channel was created but not yet usable was almost instantly. But with dual funding, we need to wait for a few confirmations, making the "created-but-not-usable-yet" window much larger (tens of minutes).

Note that even in the zero-conf case it is cleaner to handle this explicitly, because even if the "created-but-not-usable-yet" was very small, it was still non-zero.


dpad85 (approved):
@robbiehanson We'll add a new query in the payments db that updates the `confirmed` flag when we receive the `NodeEvents.Confirmed` event. We will also have to update the UI to discriminate payments that were received with at least 1 `NewChannel` type that is not yet confirmed.

src/commonTest/kotlin/fr/acinq/lightning/channel/states/WaitForChannelReadyTestsCommon.kt
```diff
@@ -71,12 +71,12 @@ class WaitForChannelReadyTestsCommon : LightningTestSuite() {
         val (alice, channelReadyAlice, bob, channelReadyBob) = init()
         val (alice1, actionsAlice1) = alice.process(ChannelCommand.MessageReceived(channelReadyBob))
         assertIs<LNChannel<Normal>>(alice1)
-        assertEquals(2, actionsAlice1.size)
+        assertEquals(3, actionsAlice1.size)
```
t-bast, line 74: It would be useful to explicitly test that we're sending the `Confirmed` event here - to be added in a next PR whenever we have the occasion

pm47: Good catch, noted

PR #404, opened 2022-11-15
Updated logic for `NewChannel.confirmed`

robbiehanson:
In PR #403, we made a few changes to account for the new dual-funding flow:

```kotlin
// PaymentsDb.kt:
data class IncomingPayment.ReceivedWith.NewChannel(
  // ...
  val confirmed = false
)

// NodeEvents.kt:
sealed interface ChannelEvents : NodeEvents {
  // ...
  data class Confirmed(val state: Normal) : ChannelEvents
}
```

This PR builds upon these changes.

~~First, we connect `ChannelEvents.Confirmed` to a database action that will set the `confirmed` flag to true.~~

And we noted in 403:

> We will also have to update the UI to discriminate payments that were received with at least 1 `NewChannel` type that is not yet confirmed.

We do this via `WalletPayment.completedAt()`, which now returns zero for this specific case.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/db/InMemoryPaymentsDb.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/db/InMemoryPaymentsDb.kt
@@ -56,6 +56,26 @@ class InMemoryPaymentsDb : PaymentsDb {
         }
     }
 
+    override suspend fun updateNewChannelConfirmed(
+        preimage: ByteVector32,
+        channelId: ByteVector32
+    ) {
+        val paymentHash = preimage.sha256()
+        val payment = incoming[paymentHash]
+        when (payment?.received?.receivedWith) {
+            null -> Unit // no-op
+            else -> incoming[paymentHash] = run {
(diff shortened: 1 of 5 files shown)


robbiehanson:
The majority of this logic has now been moved to the Phoenix layer. However, there is still one thing we need to change:

`WalletPayment.completedAt()` must return zero for this specific case (when `NewChannel.confirmed == false`)

Without this change, the method is worthless within Phoenix. And we'll have to change all code that uses it to instead use some other function that includes this change.

However, it's worth noting that `completedAt()` is **unused** within lightning-kmp, except within `InMemoryPaymentsDb` (which only exists for unit-test purposes). So another possible code change would be:

- remove `WalletPayment.completedAt`
- add it back as a private method within `InMemoryPaymentsDb`

dpad85 (approved):
LGTM

We can keep `completedAt()` in lightning-kmp. We'll reassess when we clean up `PaymentsDb` (remove all Phoenix-specific methods and concepts from lightning-kmp). For example I think `InMemoryPaymentsDb` should be defined in the tests, not the main codebase.

PR #405, opened 2022-11-18
`assertIs<LNChannel<X>>` doesn't work as expected

robbiehanson:
According to the docs for [assertIs](https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-is.html):

> Note that due to type erasure the type check may be partial (e.g. `assertIs<List<String>>(value)` only checks for the class being `List` and not the type of its elements because it's erased).

And we had a **LOT** of `assertIs<LNChannel<X>>` checks :grimacing:

Note that this PR isn't finished, because we still have code like this that we need to fix:

```
fun <T : ChannelState> someFun(): LNChannel<T> {
  // ...
  assertIs<LNChannel<T>>(state)
  return state
}
```


pm47:
Dammit, you're right. We even touched this topic during the review: https://github.com/ACINQ/lightning-kmp/pull/391#discussion_r1011921582.

I believe that most of the time, due to us using the inner `state` through smart cast, the test would break in case of type mismatch though.

robbiehanson:
From the previous [discussion](https://github.com/ACINQ/lightning-kmp/pull/391#discussion_r1011921582), it sounds like there may be better ways to solve this issue.

The problem with this simple approach is that this:
```kotlin
assertIs<Closing>(alice.state)
```

does NOT tell the compiler that `alice is LNChannel<Closing>`. And so you get several compiler errors. Which means you either have to add a redundant check:
```kotlin
assertIs<LNChannel<Closing>>(alice)
```

or refactor a bunch of tests, somehow.

I'm closing this PR, and replacing it with issue #412

PR #406, opened 2022-11-21
Handle disconnects during dual-funding flow

robbiehanson:
During the dual-funding flow, a disconnect will lead to a bad state. (e.g. disconnect occurs while in state `WaitForFundingCreated`) Specifically, the UTXOs remain "reserved", and the corresponding sats are stuck in the user's swap-in wallet until they restart the app.

This PR addresses the problem by assigning the UTXOs to a `reserved.pending` state when the flow is started. And from there:

- if `ChannelEvents.Created` is seen, UTXOs move to `reserved.reserved` state
- if `Disconnected` is seen, UTXOs are removed from `reserved.pending`

It would be nice to add a unit test for this. But, if I understand correctly, this would require injecting Electrum messages into the ElectrumClient. Which isn't possible now (it's a private API). But the `electrum-client-rework` branch might help us here.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
@@ -7,6 +7,7 @@ import fr.acinq.lightning.Lightning.nodeFee
 import fr.acinq.lightning.blockchain.fee.OnChainFeeConf
 import fr.acinq.lightning.crypto.KeyManager
 import fr.acinq.lightning.utils.toMilliSatoshi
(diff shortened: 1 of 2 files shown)


t-bast:
This code change is potentially quite dangerous, there is a risk that we end up creating multiple dual-funding attempts in parallel with the same utxos, which means none of them will work because they double-spend each other, and Phoenix doesn't have double-spend detection so the user will only see that channels are stuck in a weird state.

If we really want to go in that direction, that needs to be thoroughly tested, otherwise it will introduce more issues than the one it solves.

PR #401, opened 2022-11-10
Use MDC instead of hardcoding "key:value" in log messages.

pm47:
It should make debugging easier.

Formatting is not necessarily great, by default kodein logs puts one indented MDC item per line below the message. Although it may not be a bad thing on small mobile screens.

Examples:

```
13:38:52.835 INFO  IncomingPaymentHandler - generated payment request lnbcrt1500n1p3keuxupp5unfkc90kv6rz24djsf2e5ycl6zgge22qen29dxr7d9fels7njyxsdq0w4hxjapqw3jhxaqcqpjsp5lf7yykx3prkmr8d6z5nm3hyvu2pdhzrkqe9qtdlgvadllm6h8l4q9q7sqqqqqqqqqqqqqqqqqqqsqqqqqysgqmqz9g3lpvuzjwzeerq6mr7u08ge6pjcgd60gs65fwkq23wj2qfvpm2mdjrsa02939g2039jcyred823l4c4x06w59nz9e4j99ugpe2u608jqpv3gfat
    paymentHash: e4d36c15f666862555b282559a131fd0908ca940ccd456987e69539fc3d3910d
13:38:52.837 INFO  IncomingPaymentHandler - processing htlc part expiry=CltvExpiry(400144)
    paymentHash: e4d36c15f666862555b282559a131fd0908ca940ccd456987e69539fc3d3910d
    amount: 100000 msat
    totalAmount: 150000 msat
13:38:52.839 INFO  IncomingPaymentHandler - processing pay-to-open part amount=30000 msat funding=100000 sat fees=3 sat
    paymentHash: e4d36c15f666862555b282559a131fd0908ca940ccd456987e69539fc3d3910d
    amount: 30000 msat
    totalAmount: 150000 msat
13:38:52.841 INFO  IncomingPaymentHandler - processing pay-to-open part amount=20000 msat funding=100000 sat fees=2 sat
    paymentHash: e4d36c15f666862555b282559a131fd0908ca940ccd456987e69539fc3d3910d
    amount: 20000 msat
    totalAmount: 150000 msat
13:38:52.841 WARN  IncomingPaymentHandler - amount too low for a pay-to-open: 50000 msat, min is 75000 msat
    paymentHash: e4d36c15f666862555b282559a131fd0908ca940ccd456987e69539fc3d3910d
    amount: 20000 msat
    totalAmount: 150000 msat
```

```
[...]

PR #410, opened 2022-11-23
Include channel backup whenever it fits

t-bast:
We previously omitted the first channel backup because it risked overflowing the `tx_signatures` message. This is an issue for users who do a swap-in, never use the channel, and expect to be able to restore their seed on a different device while keeping the channel open.

We now always try to include the channel backup, and simply omit it when it doesn't fit in the message. This should work for most users since transaction sizes are usually well below 64kB. It won't work when the swap transaction is too large (e.g. when doing a large CoinJoin), but in that case our node will send an `update_fee` immediately after the funding tx confirms, which will automatically generate a post-funding backup.

NB: this builds on top of #409


robbiehanson:
I just ran into this bug. I had been testing swap-in, and had created several channels. Then I restored the app, and discovered there was only a single channel. From what I can tell, it was the channel used to send a normal payment.

pm47:
On [d658f23](https://github.com/ACINQ/lightning-kmp/pull/410/commits/d658f2323cc42b3376d0cf2cca51c5f9c705e936), why would there be a deadlock if the ordering of signatures is properly specified?

t-bast:
> On [d658f23](https://github.com/ACINQ/lightning-kmp/pull/410/commits/d658f2323cc42b3376d0cf2cca51c5f9c705e936), why would there be a deadlock if the ordering of signatures is properly specified?

This is a safeguard to be able to ship that code even though the spec PR isn't merged, since the specified ordering of `tx_signatures` may still change. I think it won't change, but by always sending our `tx_signatures` first we're sure it will work even if the spec changes!

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -154,11 +155,22 @@ data class EncryptedChannelData(val data: ByteVector) {
     }
 }
 
-interface HasEncryptedChannelData : LightningMessage {
+interface HasEncryptedChannelData : HasChannelId {
     val channelData: EncryptedChannelData
-    fun withChannelData(data: ByteVector): HasEncryptedChannelData = withChannelData(EncryptedChannelData(data))
-    fun withChannelData(ecd: EncryptedChannelData): HasEncryptedChannelData = if (ecd.isEmpty()) this else withNonEmptyChannelData(ecd)
     fun withNonEmptyChannelData(ecd: EncryptedChannelData): HasEncryptedChannelData
+    fun withChannelData(data: ByteVector, logger: Logger? = null): HasEncryptedChannelData = withChannelData(EncryptedChannelData(data), logger)
```
pm47, line 161: This is a bit convoluted but I don't have better to propose. Ideally `HasEncryptedChannelData` should be a `sealed class` and `withNonEmptyChannelData` a `protected abstract fun`, but double inheritance is forbidden so we are stuck with an interface.

pm47 (approved):
LGTM

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -154,11 +155,22 @@ data class EncryptedChannelData(val data: ByteVector) {
...
+    fun withChannelData(ecd: EncryptedChannelData, logger: Logger? = null): HasEncryptedChannelData {
+        // The maximum size for lightning messages is 65535 bytes, so we only include a channel backup when we have enough space available.
+        // Our peer needs to be able to send the backup back in their channel_reestablish message, and we don't know the size of that message.
+        // We use an upper bound of 64000 bytes, hoping that it will be enough to make it work in all cases.
+        val sizeWithoutBackup = write().size
+        return if (!ecd.isEmpty() && ecd.data.size() + sizeWithoutBackup <= 64_000) {
```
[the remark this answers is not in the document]
t-bast: > A simple hard limit at 60_000B seems a better option I think.

That's not enough, that can still create invalid backups if the message without it is more than 6kB. I'll update that code to do both: a hard limit on the back at 60_000 byte and a check that it doesn't overflow the message.

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -154,11 +155,22 @@ data class EncryptedChannelData(val data: ByteVector) {
...
+            withNonEmptyChannelData(ecd)
+        } else {
+            logger?.warning { "c:${this.channelId} could not include encrypted backup in ${this::class.simpleName}: too large (${ecd.data.size()} bytes)" }
```
pm47, line 170: On an actual restore with an eclair peer, this prints `too large (0 bytes)`:

```
18:11:27.648 INFO  f.a.l.c.ChannelContext - c:c84d2256d3441abb0e0782116e1ff5b262175bc85a6c90daad7c93063987d248 channel_reestablish includes a peer backup
18:11:27.649 INFO  f.a.l.c.ChannelContext - c:c84d2256d3441abb0e0782116e1ff5b262175bc85a6c90daad7c93063987d248 ignoring their older backup
18:11:27.661 WARN  f.a.l.c.ChannelContext - c:c84d2256d3441abb0e0782116e1ff5b262175bc85a6c90daad7c93063987d248 could not include encrypted backup in ChannelReestablish: too large (0 bytes)
```

There is another faulty log `ignoring their older backup` that I will fix in #401.

t-bast: Good catch, I'm printing that error also for the case where the backup is empty, it's a mistake.

pm47: > There is another faulty log ignoring their older backup that I will fix in https://github.com/ACINQ/lightning-kmp/pull/401.

Done in [bbc3456](https://github.com/ACINQ/lightning-kmp/pull/401/commits/bbc3456f00cffe2e580a175e5d65e62267aa6132).

t-bast: I avoid logging anything on empty backups in https://github.com/ACINQ/lightning-kmp/pull/410/commits/973e9ce48d2708a8f8d5f07b31ab71d5d8cb7ef7

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/410/commits/973e9ce48d2708a8f8d5f07b31ab71d5d8cb7ef7

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -154,11 +155,24 @@ data class EncryptedChannelData(val data: ByteVector) {
...
+        // We use an upper bound of 60000 bytes, hoping that it will be enough to make it work in all cases.
+        val sizeWithoutBackup = write().size
+        return when {
+            ecd.isEmpty() -> this
+            ecd.data.size() + sizeWithoutBackup <= 60_000 -> withNonEmptyChannelData(ecd)
+            else -> {
+                logger?.warning { "c:${this.channelId} could not include encrypted backup in ${this::class.simpleName}: too large (${ecd.data.size()} bytes)" }
+                this
+            }
+        }
```
pm47, line 174: Very concise and neat!

pm47 (dismissed):

pm47 (approved):

PR #411, opened 2022-11-23
Rework and simplify electrum client, miniwallet and watcher

sstone:
CompletableDeferred<> is used to handle multiple concurrent calls and provide suspend methods which are much easier to use. For example, to retrieve a transaction you can simply call `client.getTx(txid)` and you can chain such calls in your message handling loops: this eliminate the need to send messages to yourself and create intermediate states that are just waiting for a response from the electrum client.

I've also remove the `NotifyTxEvent` event which I believe is not used anywhere.


robbiehanson:
This looks like a nice cleanup!

I tested it in branch [test-pr-411](https://github.com/ACINQ/phoenix/tree/testing-pr-411) and I ran into one issue.

The problem we're having on the Phoenix side is that ElectrumClient's `connect()`/`disconnect()` functions have been replaced with `start()`/`stop()`, but they don't do the same thing. From what I'm reading, `start()` is expected to be invoked once, and has it's own reconnection logic. And `stop()`, if invoked, prevents ElectrumClient from working again.

The issue is that, Phoenix has a better understanding of the network state than lightning-kmp. Specifically, it has a [NetworkMonitor](https://github.com/ACINQ/phoenix/blob/master/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/NetworkMonitor.kt), which knows whether or not the device has an Intenet connection. This is obviously OS-specific (for example, here's the [iOS version](https://github.com/ACINQ/phoenix/blob/master/phoenix-shared/src/iosMain/kotlin/fr/acinq/phoenix/managers/NetworkMonitorIos.kt)). And since Phoenix receives notifications from the OS about the network state, it can better determine when to attempt a re-connect.

And there are other complications. On iOS, when the app is sent into the background, the OS will kill any open network connections (after a brief delay). The Phoenix layer knows when the app is going to be sent to the background, and prefers to perform a clean disconnect on its own.

Unless of course an outgoing payment is in progress when the user background's the app. In which case Phoenix asks the OS for permission to perform a "long-lived background task". If granted, it then keeps the connection(s) open until the payment completes, or we run out of granted time.

Long story short, there are a bunch of complications surrounding connect/disconnect/reconnect. All this logic is implemented in [AppConnectionsDaemon](https://github.com/ACINQ/phoenix/blob/master/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/AppConnectionsDaemon.kt). (See `AppConnectionsDaemon.TrafficControl`) So I think it makes sense to remove the auto-reconnect logic from within ElectrumClient, and allow the external layer (e.g. Phoenix) to handle it instead. (This is also how Peer works - it outsources the reconnection logic.)

sstone:
Sorry I missed that, I'll restore `connect()/disconnect()` to their original behaviour.

sstone:
Rebased and updated to be 100% compatible with the current client and watcher (tested on Android but not on iOS).

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
```diff
@@ -1,597 +1,303 @@
...
-                is GetTransaction -> true
-                is GetMerkle -> true
-                is BroadcastTransaction -> true
-                else -> false
+                    is WatcherCommand.NotifyIfReady -> {
+                        if (state.isConnected) {
+                            state.idleSince?.let {
+                                if (it < currentTimestampMillis() - 5000) {
+                                    // no requests in progress and watcher has been idle for more than 5s
+                                    logger.debug { "Watcher is up-to-date and idle for more than 5s" }
```
dpad85, line 281: This log is very spammy (every 2sec) and I don't understand the purpose.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClient.kt
```diff
@@ -202,75 +44,64 @@ class ElectrumClient(
...
+            }
+        }
     }
 
-    // Warning : lateinit vars have to be used AFTER their init to avoid any crashes
-    //
-    // This shouldn't be used outside the establishedConnection() function
-    // Except from the disconnect() one that check if the lateinit var has been initialized
-    private lateinit var socket: TcpSocket
     private fun establishConnection(serverAddress: ServerAddress) = launch {
```
dpad85, line 102: On Android with Tor enabled and a flaky connection, I've been able to sometimes crash the app because of `linesFlow` below. `linesFlow` relies on `JvmSocket.receiveAvailable` (on Android) which may throw `TcpSocket.IOException.ConnectionClosed`. This exception is not caught and will crash the `establishConnection` coroutine (and the app).

This is probably a race condition that Tor makes more obvious. I don't have a good solution for this for now and this is probably outside the scope of this PR. But in any case, I think we should add a `CoroutineExceptionHandler` here that will catch errors happening during initialisation, log the error and make sure the socket is closed.

dpad85 (commented):
Tested on Phoenix Android and iOS, it works fine. There are issues with Tor in edge cases but I think they should be investigated separately.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClient.kt
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/411/commits/f66fd8221e492a36567df03027935e8879691679

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
sstone: removed in https://github.com/ACINQ/lightning-kmp/pull/411/commits/f66fd8221e492a36567df03027935e8879691679

pm47:
I have added a few nits on branch [electrum-client-rework-pm](https://github.com/ACINQ/lightning-kmp/commits/electrum-client-rework-pm).

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
pm47: What is the purpose of `NotifyIfReady`? What are the hardcoded 2s and 5s durations about?

sstone: It is used to replicate what is done on the master branch to send out `NotifyUpToDateEvent` events. With the new watcher there is no concept of "pending requests": whenever the watcher is ready to process a command or event is means that is has processed all the calls it needs to make to the electrum server, so we don't use request timestamps anymore but instead know when the watcher has been idle for more than X seconds. The hard-coded 2s and 5s duration are copied from the master branch, so the new watcher should behave exactly like the current one.

pm47: I can't find any reference of `NotifyUpToDateEvent` in the Phoenix codebase. If we are not using it, let's drop it. @dpad85 @robbiehanson please confirm.

sstone:
> I have added a few nits on branch [electrum-client-rework-pm](https://github.com/ACINQ/lightning-kmp/commits/electrum-client-rework-pm).

Thanks, I used these changes and also updated ElectrumClient's notification flow to a ` Flow<ElectrumSubscriptionResponse>` 

> Why is this PR draft, what's left to work on?

The only thing that I'm unsure of is the handling of `NotifyUpToDateEvent` but otherwise there is nothing missing that I know of or was reported so far, I'll update the PR status

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
dpad85: I could not find any references to `NotifyIfReady` or  `NotifyUpToDateEvent` either on phoenix `master` branch or the iOS production branch. We don't use these events, I think we can drop them.

sstone: https://github.com/ACINQ/lightning-kmp/pull/411/commits/3e16024d77a450a22b15d13aff5249547caf1cfd removes ` NotifyUpToDateEvent`

pm47: Phew, feels lighter.

pm47 (approved):
Code LGTM, but this requires an ACK from the Phoenix team.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
robbiehanson: [ElectrumWatcher.openUpToDateFlow](https://github.com/ACINQ/lightning-kmp/blob/57f619f3605cc8c7eba1ae1bf29342103c5429c0/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt#L429) is being by the WatchTower [here](https://github.com/ACINQ/phoenix/blob/9306dc669bf33a059ccae7984f0e216ce8b1d9f4/phoenix-ios/phoenix-ios/officers/WatchTower.swift#L168) (via a wrapping [publisher](https://github.com/ACINQ/phoenix/blob/9306dc669bf33a059ccae7984f0e216ce8b1d9f4/phoenix-ios/phoenix-ios/kotlin/KotlinPublishers.swift#L287))

sstone: @robbiehanson can you please test commit https://github.com/ACINQ/lightning-kmp/pull/411/commits/08640a6c61435f96163862c35e1f7c89b8370168 ?

robbiehanson: :+1: I was able to compile phoenix master branch (with some minor tweaks) against https://github.com/ACINQ/lightning-kmp/commit/08640a6c61435f96163862c35e1f7c89b8370168

dpad85 (approved):
LGTM

sstone:
> LGTM

@dpad85 with or without `ElectrumWatcher.openUpToDateFlow`  (see https://github.com/ACINQ/lightning-kmp/pull/411#discussion_r1149812387) ? Or do you want to add that notification back in another PR ?

dpad85:
So the removal of `openUpToDateFlow` should be reverted as it is needed by the background watcher in the iOS app. This is ad hoc code that should be cleaned up, but it's outside the scope of this PR.

sstone:
I've restored `openUpToDateFlow` and simplified the watcher a bit (there is now one flow for watch events, one flow for up-to-date events)

PR #414, opened 2022-11-28
Dual funding isn't mandatory yet

t-bast:
While we only support dual-funding, we cannot yet make it mandatory in our `init` message because it would force our peer to activate it, whereas the specification hasn't been merged yet and cln nodes are using the feature bit with an incompatible implementation.

@dpad85 Phoenix should activate the dual-funding feature bit, but only make it `optional` for now, not `mandatory`, otherwise it won't be able to connect to the ACINQ node.


pm47:
I think it works: eclair will refuse incoming `open_channel2`, but will happily send `open_channel2` even if it doesn't support it in the init features.

pm47:
There is unfortunately some tweaks to do in eclair here: https://github.com/ACINQ/eclair/blob/6569eaf07dcf0768ea09e5eb0ce2f69d8ecbbdd5/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala#L414

t-bast:
Why? Eclair will always be funder, it will never receive `open_channel2`, so we shouldn't need to change that?

pm47:
Sorry wrong paste, it's https://github.com/ACINQ/eclair/blob/6569eaf07dcf0768ea09e5eb0ce2f69d8ecbbdd5/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala#L161

dpad85:
> Phoenix should activate the dual-funding feature bit, but only make it optional

I'm ready to update the node parameters in Phoenix as soon as this PR is merged.

pm47:
Since spec has converged and implementations should be compatible, there is a chance we will not need this PR. In that case Phoenix will set `DualFunding=Mandatory` as it is currently.

PR #416, opened 2022-12-13
Fix JVM socket `startTLS` when TLS is disabled

dpad85:
The `startTLS` method is used for Tor, where we first connect to the Tor proxy then enable TLS if necessary. This PR fixes this method so that we do not make the socket connection secure if TLS is actually disabled (which is the case with the peer connection).


pm47 (dismissed):
LGTM

pm47 (approved):

PR #420, opened 2023-02-28
Prepare `InteractiveTx` to support splicing

t-bast:
Adding splicing to the InteractiveTx protocol requires:

- adding a shared input (the current funding output)
- adding outputs (when splicing out)

The `localAmount` and `remoteAmount` provided are the amounts each peer contributes to the new funding output. Those amounts should be computed by the caller depending on what they intend to do (splice funds in/out).
This model allows batching operations: if the caller wants to do several splices in and several splices out, this is easy to do by iterating over these operations and updating the targeted `localAmount` accordingly.

When splicing, the `tx_add_input` message we send for the shared input doesn't contain the previous transaction. There are two reasons for that:

- it potentially doesn't fit into a lightning message (if > 65kB)
- we don't need it, we know it correctly uses segwit


t-bast:
Rebased.

pm47 (approved):

PR #421, opened 2023-02-28
Serialize HTLCs only once

t-bast:
HTLCs are duplicated in local and remote commit(s), and across commitments when splices are ongoing. We now serialize them only once, which reduces the size of the encrypted backup that needs to be stored.


pm47 (dismissed):
LGTM, roundtrip is tested in `StateSerializationNonRegTestsCommon`

t-bast:
Rebased.

pm47 (approved):

PR #424, opened 2023-03-06
Publish `NodeEvent` when a pay-to-open is rejected

dpad85:
We publish new pay-to-open events in `NodeParams.nodeEvents`. Those events can be used by a front app like Phoenix to know when a pay-to-open succeeds or is rejected.

For now, only the rejection scenario implemented, and only when it's because the amount is below minimum. This event is emitted by the incoming payment handler.

```kotlin
sealed interface PayToOpenEvents : NodeEvents {
    sealed class Rejected: PayToOpenEvents {
        data class BelowMin(val request: PayToOpenRequest): Rejected()
    }
}
```

The event contains the `PayToOpenRequest` so that the front can display a nice message like so:

image

Note: this PR has been significantly reworked, hence the revert commits that can be dismissed.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -99,7 +99,6 @@ data class PurgeExpiredPayments(val fromCreatedAt: Long, val toCreatedAt: Long) 
 
 sealed class PeerEvent
 data class PaymentRequestGenerated(val receivePayment: ReceivePayment, val request: String) : PeerEvent()
-data class PaymentNotReceived(val actions: List<PeerCommand>, val incomingPayment: IncomingPayment?) : PeerEvent()
 data class PaymentReceived(val incomingPayment: IncomingPayment, val received: IncomingPayment.Received) : PeerEvent()
 data class PaymentProgress(val request: SendPayment, val fees: MilliSatoshi) : PeerEvent()
 data class PaymentNotSent(val request: SendPayment, val reason: OutgoingPaymentFailure) : PeerEvent()
@@ -560,7 +559,6 @@ class Peer(
             is Either.Left -> incomingPaymentHandler.process(item.value, currentBlockHeight)
         }
         when (result) {
-            is IncomingPaymentHandler.ProcessAddResult.Rejected -> _eventsFlow.emit(PaymentNotReceived(result.actions, result.incomingPayment))
             is IncomingPaymentHandler.ProcessAddResult.Accepted -> _eventsFlow.emit(PaymentReceived(result.incomingPayment, result.received))
             else -> Unit
         }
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -99,6 +99,7 @@ data class PurgeExpiredPayments(val fromCreatedAt: Long, val toCreatedAt: Long)
 
 sealed class PeerEvent
 data class PaymentRequestGenerated(val receivePayment: ReceivePayment, val request: String) : PeerEvent()
+data class PaymentNotReceived(val actions: List<PeerCommand>, val incomingPayment: IncomingPayment?) : PeerEvent()
```
pm47, line 102: Nit: I think the name is confusing, I understood it first as "Payment not yet received".

But most importantly, there is an inconsistency between `PaymentReceived` and `PaymentNotReceived`:
- `PaymentReceived` is for a whole logical payment
- `PaymentNotReceived` will be sent for each rejected payment part

So the two messages don't operate at the same level, and you would have to do some filtering to extract the cases that you are interested in, which I believe is: all parts of a payment have been received, but we are rejecting the payment due to \<reasons\>.

dpad85: The inconsistency is a good point and reveals further issues, I reverted the change. Listening to `ProcessAddResult` is not a good idea.

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -256,11 +258,13 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
                                     is PayToOpenPart -> actionForPayToOpenFailure(privateKey, failureMsg, part.payToOpenRequest) // NB: this will fail all parts, we could only return one
                                 }
                             }
+                            payment.parts.filterIsInstance<PayToOpenPart>().firstOrNull()?.let {
```
pm47, line 261: The `firstOrNull()` will select a single payment part, it kind of create the same issue as before. You should use the aggregated amount `payToOpenAmount`, which is what we actually base the decision on. I think the error message should also contain the reference value.

Maybe break up the attributes?

```kotlin
sealed interface PayToOpenEvents : NodeEvents {
    abstract val paymentHash: ByteVector32
    abstract val totalAmount: MilliSatoshi
    abstract val payToOpenAmount: MilliSatoshi
    sealed class Rejected: PayToOpenEvents {
        data class BelowMin(override val paymentHash: ByteVector32, override val totalAmount: MilliSatoshi, override val payToOpenAmount: MilliSatoshi, val minPayToOpenAmount: MilliSatoshi): Rejected()
    }
}
```

dpad85: Fixed in 25039c9

pm47 (approved):

PR #426, opened 2023-03-22
Store channel state after sending `commit_sig`

t-bast:
We must store the channel state after sending our `tx_signatures`, because our peer may broadcast the transaction at that point. But it's useful to store it earlier than that, to allow resuming the signatures exchange if we get disconnected, otherwise we could end up in a state where one peer has forgotten the channel while the other has sent `tx_signatures` and must thus wait for the channel to be spent or double-spent.

With that change, we can cleanly handle any disconnection:

- if we get disconnected before any peer sent `commitment_signed`, the channel is forgotten by both peers
- if we get disconnected after one peer send `commitment_signed` but the other peer didn't, the channel will be forgotten when reconnecting (this is safe because the peer that sent `commitment_signed` did *not* send `tx_signatures` since the other peer didn't send `commitment_signed`)
- if we get disconnected after both peers sent `commitment_signed`, we simply resume the signatures exchange on reconnection

We introduce a new TLV field to `channel_reestablish` that contains the `txid` of the funding transaction that is partially signed: this lets peers figure out which messages to send back on reconnection.

See https://github.com/lightning/bolts/pull/851#discussion_r1135205227 for more details.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
@@ -101,7 +101,7 @@ sealed class LocalFundingStatus {
     abstract val signedTx: Transaction?
     abstract val txId: ByteVector32
 
-    data class UnconfirmedFundingTx(val sharedTx: SignedSharedTransaction, val fundingParams: InteractiveTxParams, val localCommitSig: CommitSig, val createdAt: Long) : LocalFundingStatus() {
+    data class UnconfirmedFundingTx(val sharedTx: SignedSharedTransaction, val fundingParams: InteractiveTxParams, val createdAt: Long) : LocalFundingStatus() {
         override val signedTx: Transaction? = sharedTx.signedTx
         override val txId: ByteVector32 = sharedTx.localSigs.txId
     }
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
@@ -88,10 +88,23 @@ data class CommitmentChanges(val localChanges: LocalChanges, val remoteChanges: 
 
 data class HtlcTxAndSigs(val txinfo: HtlcTx, val localSig: ByteVector64, val remoteSig: ByteVector64)
 data class PublishableTxs(val commitTx: CommitTx, val htlcTxsAndSigs: List<HtlcTxAndSigs>)
+
 /** The local commitment maps to a commitment transaction that we can sign and broadcast if necessary. */
 data class LocalCommit(val index: Long, val spec: CommitmentSpec, val publishableTxs: PublishableTxs)
+
 /** The remote commitment maps to a commitment transaction that only our peer can sign and broadcast. */
-data class RemoteCommit(val index: Long, val spec: CommitmentSpec, val txid: ByteVector32, val remotePerCommitmentPoint: PublicKey)
+data class RemoteCommit(val index: Long, val spec: CommitmentSpec, val txid: ByteVector32, val remotePerCommitmentPoint: PublicKey) {
+    fun sign(keyManager: KeyManager, params: ChannelParams, commitInput: Transactions.InputInfo): CommitSig {
+        val (remoteCommitTx, htlcTxs) = Commitments.makeRemoteTxs(keyManager, index, params.localParams, params.remoteParams, commitInput, remotePerCommitmentPoint, spec)
+        val channelKeys = params.localParams.channelKeys(keyManager)
+        val sig = keyManager.sign(remoteCommitTx, channelKeys.fundingPrivateKey)
+        // we sign our peer's HTLC txs with SIGHASH_SINGLE || SIGHASH_ANYONECANPAY
+        val sortedHtlcsTxs = htlcTxs.sortedBy { it.input.outPoint.index }
+        val htlcSigs = sortedHtlcsTxs.map { keyManager.sign(it, channelKeys.htlcKey, remotePerCommitmentPoint, SigHash.SIGHASH_SINGLE or SigHash.SIGHASH_ANYONECANPAY) }
+        return CommitSig(params.channelId, sig, htlcSigs.toList())
+    }
+}
+
 /** We have the next remote commit when we've sent our commit_sig but haven't yet received their revoke_and_ack. */
 data class NextRemoteCommit(val sig: CommitSig, val commit: RemoteCommit)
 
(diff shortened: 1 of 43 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -587,11 +601,139 @@ data class InteractiveTxSession(
...
+
+/**
+ * Once a shared transaction has been created, peers exchange signatures for the commitment and the shared transaction.
+ * We store the channel state once we reach that step. Once we've sent tx_signatures, we cannot forget the channel
+ * until it has been spent or double-spent.
+ */
+data class InteractiveTxSigningSession(
+    val fundingParams: InteractiveTxParams,
+    val fundingTx: PartiallySignedSharedTransaction,
+    val localCommitSig: CommitSig,
```
pm47, line 631: I believe `localCommitSig` should be returned as an action `InteractiveTxSigningSessionAction.SendCommitSig` by the `create` method.

src/commonMain/kotlin/fr/acinq/lightning/channel/WaitForFundingSigned.kt
```diff
@@ -7,109 +7,59 @@ import fr.acinq.lightning.MilliSatoshi
...
 import fr.acinq.lightning.blockchain.WatchConfirmed
-import fr.acinq.lightning.blockchain.electrum.WalletState
 import fr.acinq.lightning.crypto.ShaChain
 import fr.acinq.lightning.utils.Either
 import fr.acinq.lightning.wire.*
 import kotlin.math.absoluteValue
 
 /*
  * We exchange signatures for a new channel.
+ * If we send our tx_signatures first, the protocol flow is:
```
pm47, line 17:
```suggestion
 * We have already sent our commit_sig.
 * If we send our tx_signatures first, the protocol flow is:
```

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -99,7 +101,7 @@ sealed class LocalFundingStatus {
     abstract val signedTx: Transaction?
     abstract val txId: ByteVector32
 
-    data class UnconfirmedFundingTx(val sharedTx: SignedSharedTransaction, val fundingParams: InteractiveTxParams, val createdAt: Long) : LocalFundingStatus() {
+    data class UnconfirmedFundingTx(val sharedTx: SignedSharedTransaction, val fundingParams: InteractiveTxParams, val localCommitSig: CommitSig, val createdAt: Long) : LocalFundingStatus() {
```
pm47, line 104: I understand why you put this here (need to resend at reconnection in the rbf case), but it's inconsistent with the `WaitForFundingSigned` case, and probably footgunny because this may not be in sync with the actual commitment in the zero-conf case.

I think we should persist the `InteractiveTxSigningSession` in the rbf (and splice) cases.

src/commonMain/kotlin/fr/acinq/lightning/channel/Syncing.kt
```diff
@@ -46,15 +50,35 @@ data class Syncing(val state: ChannelStateWithCommitments, val waitForTheirReest
...
-                    val yourLastPerCommitmentSecret = nextState.commitments.remotePerCommitmentSecrets.lastIndex?.let { nextState.commitments.remotePerCommitmentSecrets.getHash(it) } ?: ByteVector32.Zeroes
-                    val myCurrentPerCommitmentPoint = keyManager.commitmentPoint(state.commitments.params.localParams.channelKeys(keyManager).shaSeed, nextState.commitments.localCommitIndex)
-                    val channelReestablish = ChannelReestablish(
-                        channelId = nextState.channelId,
-                        nextLocalCommitmentNumber = nextState.commitments.localCommitIndex + 1,
-                        nextRemoteRevocationNumber = nextState.commitments.remoteCommitIndex,
-                        yourLastCommitmentSecret = PrivateKey(yourLastPerCommitmentSecret),
-                        myCurrentPerCommitmentPoint = myCurrentPerCommitmentPoint
-                    ).withChannelData(nextState.commitments.remoteChannelData, logger)
+                    val channelReestablish = when (nextState) {
```
pm47, line 53: This part should probably be factored between `Offline` and `Syncing`. Maybe we should initialize an `Offline` state here and make it process the `channel_reestablish`?

t-bast: Refactored in https://github.com/ACINQ/lightning-kmp/pull/426/commits/93c214b7cdcde0149f64389590a4ed940fa9a55c, I simply created a helper function in `PersistedChannelState` that create `channel_reestablish`.

src/commonMain/kotlin/fr/acinq/lightning/channel/Syncing.kt
```diff
@@ -31,11 +31,15 @@ data class Syncing(val state: ChannelStateWithCommitments, val waitForTheirReest
                         logger.info { "channel_reestablish includes a peer backup" }
                         when (val decrypted = runTrying { PersistedChannelState.from(staticParams.nodeParams.nodePrivateKey, cmd.message.channelData) }) {
                             is Try.Success -> {
-                                if (decrypted.get().commitments.isMoreRecent(state.commitments)) {
-                                    logger.warning { "they have a more recent commitment, using it instead" }
-                                    decrypted.get()
-                                } else {
-                                    state
+                                val decryptedState = decrypted.get()
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/426/commits/12b76082b07f0566af13dd26b051deabd7621c9d

src/commonMain/kotlin/fr/acinq/lightning/channel/Offline.kt
```diff
@@ -28,26 +28,50 @@ data class Offline(val state: ChannelStateWithCommitments) : ChannelState() {
                     staticParams.nodeParams.features.hasFeature(Feature.ChannelBackupClient) -> {
                         // We wait for them to go first, which lets us restore from the latest backup if we've lost data.
                         logger.info { "syncing ${state::class}, waiting fo their channelReestablish message" }
-                        val nextState = state.updateCommitments(state.commitments.copy(params = state.commitments.params.updateFeatures(cmd.localInit, cmd.remoteInit)))
+                        val nextState = when (state) {
+                            is ChannelStateWithCommitments -> state.updateCommitments(state.commitments.copy(params = state.commitments.params.updateFeatures(cmd.localInit, cmd.remoteInit)))
+                            else -> state
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/426/commits/12b76082b07f0566af13dd26b051deabd7621c9d

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
t-bast: Removed in https://github.com/ACINQ/lightning-kmp/pull/426/commits/46f9be9a8e46585432ba51f9b503e7eb22b4c3e7, we now re-compute our `commit_sig` when necessary, which is better.

pm47: At first I thought there would be an opportunity for factorization between `RemoteCommit.sign` and `Commitment.sendCommit` but there is a chicken-and-egg issue.

pm47 (approved):
LGTM 🔨 🔧 🧰

PR #428, opened 2023-03-30
Allow negative contributions in `InteractiveTx`

t-bast:
When splicing over an existing channel, it is simpler to specify the amount we will add or remove from the channel, rather than the resulting amount that takes the previous balance into account. It removes the need for truncating msat balances and trivially carries over those balances to the new commitment.

It also works with dual funding, where these amounts will always be positive (since we don't have previous balances).

This requires changing the `tx_init_rbf` and `tx_ack_rbf` messages to use signed amounts, which needs to be discussed in the specification since it impacts the dual funding spec PR.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
@@ -65,16 +65,12 @@ data class InteractiveTxParams(
     constructor(channelId: ByteVector32, isInitiator: Boolean, localContribution: Satoshi, remoteContribution: Satoshi, fundingPubkeyScript: ByteVector, lockTime: Long, dustLimit: Satoshi, targetFeerate: FeeratePerKw) :
             this(channelId, isInitiator, localContribution, remoteContribution, null, fundingPubkeyScript, listOf(), lockTime, dustLimit, targetFeerate)
 
+    /** The amount of the new funding output, which is the sum of the shared input, if any, and both sides' contributions. */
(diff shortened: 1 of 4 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -48,27 +53,28 @@ sealed class SharedFundingInput {
...
     // BOLT 2: MUST set `feerate` greater than or equal to 25/24 times the `feerate` of the previously constructed transaction, rounded down.
     val minNextFeerate: FeeratePerKw = targetFeerate * 25 / 24
     // BOLT 2: the initiator's serial IDs MUST use even values and the non-initiator odd values.
     val serialIdParity = if (isInitiator) 0 else 1
+
+    /**
+     * @param previousFundingAmount if we're splicing over an existing channel, amount of the previous funding output.
+     * @return the amount of the new funding output.
+     */
+    fun fundingAmount(previousFundingAmount: Satoshi): Satoshi = previousFundingAmount + localContribution + remoteContribution
```
pm47, line 77: 😎 

```suggestion

       /**
     * @return the amount of the new funding output, which the sum of the shared input, if any, and both side's contributions.
     */
    fun fundingAmount: Satoshi = (sharedInput?.info?.txOut?.amount ?: 0.sat) + localContribution + remoteContribution
```

t-bast: Nice, good catch, I found it annoying to make this a function! Fixed in https://github.com/ACINQ/lightning-kmp/pull/428/commits/af7e7ef33cc2ecad04f24f83b1367722e68a514c

pm47 (approved):

PR #429, opened 2023-03-30
Store restored channel immediately

pm47:
Builds on #448.

When we restore a channel from a peer backup, we need to store the channel in database as soon as possible, so we are exactly in the same state as if the channel was restored from local db.

This applies whether we are restoring from empty or outdated local db.

This is the backport of 626deed7e2c9e207f085b3e0b53c9e78e42d45e2 on master.


src/commonMain/kotlin/fr/acinq/lightning/channel/Syncing.kt
```diff
@@ -49,7 +46,13 @@ data class Syncing(val state: PersistedChannelState, val waitForTheirReestablish
                         state
                     }
                     val channelReestablish = nextState.run { createChannelReestablish() }
-                    val actions = listOf(ChannelAction.Message.Send(channelReestablish))
+                    val actions = buildList {
+                        if (nextState != state) {
+                            // we just restored from backup
+                            add(ChannelAction.Storage.StoreState(nextState))
```
t-bast, line 52: We can probably update one of our restore unit tests to check that `StoreState` is emitted here?

pm47: Clearly, I'm in-between rebases and this was meant as a reminder. Let put it back to Draft.

pm47: Done in [8b37758](https://github.com/ACINQ/lightning-kmp/pull/429/commits/8b377580ebbdf362b2a285b63be7a172d2267668).

pm47:
Rebased on #448.

pm47:
Superseded by #454.

PR #435, opened 2023-04-04
Remove public channel updates support

t-bast:
Mobile wallets aren't relaying payments, so they don't need to advertise the fact that a channel is disabled when a disconnection happens. We also don't need to watch for `deeply buried` events and keep an (optional) channel announcement in our channel state.

Less code! :broom: :broom:


pm47 (approved):
LGTM 🧹

t-bast:
Rebased.

pm47 (approved):

pm47 (approved):

PR #437, opened 2023-04-04
Expose swap-in and final wallets' path in a single place

dpad85:
This PR adds the path of the final wallet (= the wallet used for the closing script) and the swap-in wallet in the `LocalKeyManager` object, in order to make them easily available and also to have a single source of truth:
- the final wallet uses account 0;
- the swap-in wallet uses account 1;
- both are on the BIP-84 path;
- they should be defined in one single place and we should never directly work with the account number.

This PR also adds a `p2wpkhXpub` method that returns the key manager's extended public key for a given path.

---

Note: [ElectrumMiniWallet](https://github.com/ACINQ/lightning-kmp/blob/4253a23ee6b53c2c6c2ed643c39a3c4248c7c60d/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt#L83) directly uses the account, AFAICT it should be using our new `LocalKeyManager.swapInXpubPath`.

I think this could be fixed in another PR, the key manager deserves a refactoring (many unused method for example), but I tried to keep changes as small as possible to avoid merge conflicts with the current work on splices.


src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
```diff
@@ -50,12 +50,23 @@ data class LocalKeyManager(val seed: ByteVector, val chainHash: ByteVector32) :
...
+                Block.LivenetGenesisBlock.hash -> DeterministicWallet.zpub
+                Block.TestnetGenesisBlock.hash -> DeterministicWallet.vpub
+                else -> throw IllegalArgumentException("unhandled chain hash $chainHash")
+            }
+        )
     }
 
     override fun closingPubkeyScript(fundingPubKey: PublicKey): Pair<PublicKey, ByteArray> {
-        val priv = bip84PrivateKey(0, 0)
+        val priv = derivePrivateKey(master, finalAddressPath(chainHash)).privateKey
```
dpad85, line 69: This change is not necessary but I feel that we should avoid working directly with the `account` and `index` when possible.

pm47 (approved):

PR #438, opened 2023-04-06
Add splice tests and fix a few issues

t-bast:
This PR is based on #430, it was easier for me to really dive into the code by writing tests and fixing issues that I found during the review.

One test doesn't pass, where we have a race condition between `splice_locked` and `commit_sig`. I haven't had time yet to look into it and figure out how to fix it.

I still need to add another commit to refactor local contributions computation (https://github.com/ACINQ/lightning-kmp/pull/430/files#r1156133948).

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
@@ -124,13 +124,13 @@ data class Closing(
                             // counterparty may attempt to spend its next commit tx at any time
                             handleRemoteSpentNext(watch.tx, commitments.latest)
                         }
-                        watch.tx.txIn.map { it.outPoint.txid }.contains(commitments.latest.fundingTxId) -> {
+                        watch.tx.txIn.map { it.outPoint }.contains(commitments.latest.commitInput.outPoint) -> {
                             // counterparty may attempt to spend a revoked commit tx at any time
                             handleRemoteSpentOther(watch.tx)
                         }
                         else -> when (val commitment = commitments.resolveCommitment(watch.tx)) {
                             is Commitment -> {
-                                logger.warning { "a commit tx for an older commitment has been published txid=${commitment.fundingTxId} fundingTxIndex=${commitment.fundingTxIndex}" }
+                                logger.warning { "a commit tx for an older commitment has been published fundingTxId=${commitment.fundingTxId} fundingTxIndex=${commitment.fundingTxIndex}" }
                                 Pair(this@Closing, listOf(ChannelAction.Blockchain.SendWatch(WatchConfirmed(channelId, watch.tx, staticParams.nodeParams.minDepthBlocks.toLong(), BITCOIN_ALTERNATIVE_COMMIT_TX_CONFIRMED))))
                             }
                             else -> {
@@ -141,14 +141,15 @@ data class Closing(
                     }
                     watch is WatchEventConfirmed && watch.event is BITCOIN_ALTERNATIVE_COMMIT_TX_CONFIRMED -> when (val commitment = commitments.resolveCommitment(watch.tx)) {
                         is Commitment -> {
-                            logger.warning { "a commit tx for fundingTxIndex=${commitment.fundingTxIndex} fundingTxid=${commitment.fundingTxIndex} has been confirmed" }
+                            logger.warning { "a commit tx for fundingTxIndex=${commitment.fundingTxIndex} fundingTxId=${commitment.fundingTxId} has been confirmed" }
                             val commitments1 = commitments.copy(
                                 active = listOf(commitment),
                                 inactive = emptyList()
                             )
                             val newState = this@Closing.copy(commitments = commitments1)
+                            // This commitment may be revoked: we need to verify that its index matches our latest known index before overwriting our previous commitments.
                             when {
-                                watch.tx.txid == commitments1.latest.localCommit.publishableTxs.commitTx.tx.txid -> {
+                                watch.tx.txid == commitments1.latest.localCommit.publishableTxs.commitTx.tx.txid && commitments1.localCommitIndex == commitments.localCommitIndex -> {
                                     // our local commit has been published from the outside, it's unexpected but let's deal with it anyway
                                     newState.run { spendLocalCurrent() }
                                 }
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
@@ -236,6 +236,13 @@ data class Commitment(
 
     fun isIdle(changes: CommitmentChanges): Boolean = hasNoPendingHtlcs() && changes.localChanges.all.isEmpty() && changes.remoteChanges.all.isEmpty()
 
+    /** The commitment is locked once we've sent and received splice_locked. */
+    fun ChannelContext.isLocked(): Boolean {
+        val localLocked = staticParams.useZeroConf || localFundingStatus is LocalFundingStatus.ConfirmedFundingTx
+        val remoteLocked = remoteFundingStatus == RemoteFundingStatus.Locked
+        return localLocked && remoteLocked
+    }
+
     fun timedOutOutgoingHtlcs(blockHeight: Long): Set<UpdateAddHtlc> {
         fun expired(add: UpdateAddHtlc) = blockHeight >= add.cltvExpiry.toLong()
 
@@ -792,7 +799,7 @@ data class Commitments(
                 )
                 val commitment = commitments1.all.find { it.fundingTxId == fundingTxId }!! // NB: this commitment might be pruned at the next line
                 val commitments2 = commitments1.run { deactivateCommitments() }.run { pruneCommitments() }
-                logger.info { "commitments active=${commitments2.active.map { it.fundingTxIndex}} inactive=${commitments2.inactive.map { it.fundingTxIndex}}" }
+                logger.info { "commitments active=${commitments2.active.map { it.fundingTxIndex }} inactive=${commitments2.inactive.map { it.fundingTxIndex }}" }
                 Either.Right(Pair(commitments2, commitment))
             }
             else -> {
@@ -805,7 +812,7 @@ data class Commitments(
     fun ChannelContext.updateLocalFundingStatus(fundingTxId: ByteVector32, status: LocalFundingStatus): Either<Commitments, Pair<Commitments, Commitment>> =
         updateFundingStatus(fundingTxId) { c: Commitment, _: Long ->
             if (c.fundingTxId == fundingTxId) {
-                logger.debug { "setting localFundingStatus=${status::class.simpleName} for funding fundingTxId=$fundingTxId" }
+                logger.debug { "setting localFundingStatus=${status::class.simpleName} for fundingTxId=$fundingTxId" }
                 c.copy(localFundingStatus = status)
             } else c
         }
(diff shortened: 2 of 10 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -236,6 +236,13 @@ data class Commitment(
 
     fun isIdle(changes: CommitmentChanges): Boolean = hasNoPendingHtlcs() && changes.localChanges.all.isEmpty() && changes.remoteChanges.all.isEmpty()
 
+    /** The commitment is locked once we've sent and received splice_locked. */
+    fun ChannelContext.isLocked(): Boolean {
+        val localLocked = staticParams.useZeroConf || localFundingStatus is LocalFundingStatus.ConfirmedFundingTx
+        val remoteLocked = remoteFundingStatus == RemoteFundingStatus.Locked
+        return localLocked && remoteLocked
+    }
+
```
pm47, line 245: This isn't equivalent to the code that was in `deactivateCommitments` and I think is incorrect. In the case where there are several funding tx chained, we may not receive a splice_locked for every single one and need to handle the intersection between local and remote.

t-bast: Good catch! But there was an issue in `deactivateCommitments` (or at least something that bothered me):

```kotlin
val lastLocalLockedIndex: Long = active.firstOrNull { staticParams.useZeroConf || it.localFundingStatus is LocalFundingStatus.ConfirmedFundingTx }?.fundingTxIndex ?: -1
val lastRemoteLockedIndex: Long = active.firstOrNull { it.remoteFundingStatus is RemoteFundingStatus.Locked }?.fundingTxIndex ?: -1
val lastLockedIndex = min(lastLocalLockedIndex, lastRemoteLockedIndex)
return when (val lastLocked = active.find { it.fundingTxIndex == lastLockedIndex }) {
```

The `lastLocked` commitment could be a different commitment from the one we use to compute `lastLocalLockedIndex` and `lastRemoteLockedIndex`, if we have RBF attempts, and I don't think this is correct...but that can probably be fixed differently, I'll take another look at this.

src/commonTest/kotlin/fr/acinq/lightning/channel/states/SpliceTestsCommon.kt
```diff
@@ -43,106 +108,822 @@ class SpliceTestsCommon : LightningTestSuite() {
...
+        val (bob2, actionsBob2) = bob1.process(ChannelCommand.MessageReceived(SpliceLocked(bob.channelId, spliceTx.hash)))
+        assertEquals(actionsBob2.size, 1)
+        actionsBob2.has<ChannelAction.Storage.StoreState>()
+        assertEquals(bob2.commitments.active.size, 1)
+        assertNotEquals(bob2.commitments.latest.fundingTxId, bob.commitments.latest.fundingTxId)
+        assertIs<LocalFundingStatus.UnconfirmedFundingTx>(bob2.commitments.latest.localFundingStatus)
+    }
+
+    @Test
+    fun `use channel before splice_locked -- zero-conf`() {
```
pm47, line 184: The test doesn't pass because this corner case has not been implemented. We did discuss it though! The check below should actually be `commits.size < active.size`, and then we should only consider the N first sigs, N being the number of active commitments:
https://github.com/ACINQ/lightning-kmp/blob/1b376e37571ea74089400886e0999c221ddaaeb6/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt#L708-L711

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
t-bast: Fixed in https://github.com/ACINQ/lightning-kmp/pull/438/commits/6fb40021dd021b1ce9a9e1651a8a54224ff9ca90

src/commonTest/kotlin/fr/acinq/lightning/channel/states/SpliceTestsCommon.kt
t-bast: This was indeed trivial to fix! Done in https://github.com/ACINQ/lightning-kmp/pull/438/commits/71b0841611b8a3004eee0fba397954e52820e0f7

src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
```diff
@@ -141,14 +141,15 @@ data class Closing(
...
+                            logger.warning { "a commit tx for fundingTxIndex=${commitment.fundingTxIndex} fundingTxId=${commitment.fundingTxId} has been confirmed" }
                             val commitments1 = commitments.copy(
                                 active = listOf(commitment),
                                 inactive = emptyList()
                             )
                             val newState = this@Closing.copy(commitments = commitments1)
+                            // This commitment may be revoked: we need to verify that its index matches our latest known index before overwriting our previous commitments.
                             when {
-                                watch.tx.txid == commitments1.latest.localCommit.publishableTxs.commitTx.tx.txid -> {
+                                watch.tx.txid == commitments1.latest.localCommit.publishableTxs.commitTx.tx.txid && commitments1.localCommitIndex == commitments.localCommitIndex -> {
```
pm47, line 152: I think the previous code was correct: at this point there is nothing we can do, our local commitment (revoked or not) has been published and confirmed. We may as well try our luck and maybe we'll get away with it. With your addition, we will end up in the `handleRemoteSpentOther` and fail to recognize the tx.

src/commonMain/kotlin/fr/acinq/lightning/channel/Normal.kt
```diff
@@ -205,8 +211,8 @@ data class Normal(
                                     is Either.Right -> {
                                         val nextState = this@Normal.copy(commitments = result.value.first)
                                         val actions = mutableListOf<ChannelAction>()
-                                        actions.add(ChannelAction.Message.Send(result.value.second))
                                         actions.add(ChannelAction.Storage.StoreState(nextState))
+                                        actions.add(ChannelAction.Message.Send(result.value.second))
```
pm47, line 215: 🦅 👁️

src/commonMain/kotlin/fr/acinq/lightning/channel/Normal.kt
```diff
@@ -404,9 +410,9 @@ data class Normal(
                             when (val fundingContributions = FundingContributions.create(
                                 params = fundingParams,
                                 sharedUtxo = Pair(sharedInput, SharedFundingInputBalances(toLocal = parentCommitment.localCommit.spec.toLocal, toRemote = parentCommitment.localCommit.spec.toRemote)),
-                                walletUtxos = spliceStatus.command.spliceIn?.wallet?.confirmedUtxos ?: emptyList(),
+                                walletInputs = spliceStatus.command.spliceIn?.wallet?.confirmedUtxos ?: emptyList(),
                                 localOutputs = spliceStatus.command.spliceOutputs,
-                                changePubKey = null // we're spending every funds available TODO: check this
+                                changePubKey = null // we don't want a change output: we're spending every funds available
```
pm47, line 415: 👍

src/commonMain/kotlin/fr/acinq/lightning/channel/Normal.kt
```diff
@@ -627,17 +643,17 @@ data class Normal(
...
                 }
-                // if we are splicing and are earlyè in the process, then we cancel it
+                // If we are splicing and are early in the process, then we cancel it.
+                val replyTo = when (spliceStatus) {
+                    is SpliceStatus.Requested -> spliceStatus.command.replyTo
+                    is SpliceStatus.InProgress -> spliceStatus.replyTo
+                    else -> null
+                }
+                replyTo?.complete(Command.Splice.Response.Failure.Disconnected)
+                // If we didn't reach the signatures exchange step, we implicitly abort the splice.
```
pm47, line 653: I don't think this refactoring makes things clearer, previous version made it obvious that it's either replyTo+forget splice, or keep going. If we keep this, maybe rename `replyTo` to `spliceReplyTo`?

src/commonMain/kotlin/fr/acinq/lightning/channel/Normal.kt
```diff
@@ -611,7 +618,16 @@ data class Normal(
             }
             is ChannelCommand.CheckHtlcTimeout -> checkHtlcTimeout()
             is ChannelCommand.WatchReceived -> when (val watch = cmd.watch) {
-                is WatchEventConfirmed -> updateFundingTxStatus(watch)
+                is WatchEventConfirmed -> {
+                    val (nextState, actions) = updateFundingTxStatus(watch)
+                    if (!staticParams.useZeroConf && nextState.commitments.active.any { it.fundingTxId == watch.tx.txid && it.fundingTxIndex > 0 }) {
+                        // We're not using 0-conf and a splice transaction is confirmed, so we send splice_locked.
+                        val spliceLocked = SpliceLocked(channelId, watch.tx.hash)
+                        Pair(nextState, actions + ChannelAction.Message.Send(spliceLocked))
```
pm47, line 626: Nice catch!

src/commonMain/kotlin/fr/acinq/lightning/channel/Normal.kt
```diff
@@ -671,7 +687,10 @@ data class Normal(
     /** If we haven't completed the signing steps of an interactive-tx session, we will ask our peer to retransmit signatures for the corresponding transaction. */
     fun getUnsignedFundingTxId(): ByteVector32? = when (spliceStatus) {
         is SpliceStatus.WaitingForSigs -> spliceStatus.session.fundingTx.txId
-        else -> null
+        else -> when (commitments.latest.localFundingStatus.signedTx) {
+            null -> commitments.latest.localFundingStatus.txId
+            else -> null
+        }
```
pm47, line 693: Nice catch again! I think rewriting the method like so is more explicit (same style as your refactoring for `CommitSig`):
```kotlin
fun getUnsignedFundingTxId(): ByteVector32? = when {
        spliceStatus is SpliceStatus.WaitingForSigs -> spliceStatus.session.fundingTx.txId
        commitments.latest.localFundingStatus is LocalFundingStatus.UnconfirmedFundingTx && commitments.latest.localFundingStatus.sharedTx is PartiallySignedSharedTransaction -> commitments.latest.localFundingStatus.txId
        else -> null
    }
```

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -170,60 +150,55 @@ sealed class FundingContributionFailure {
 /** Inputs and outputs we contribute to the funding transaction. */
 data class FundingContributions(val inputs: List<InteractiveTxInput.Outgoing>, val outputs: List<InteractiveTxOutput.Outgoing>) {
     companion object {
+        /** Compute our local splice contribution using all the funds available in our wallet. */
+        fun computeSpliceContribution(isInitiator: Boolean, commitment: Commitment, walletInputs: List<WalletState.Utxo>, localOutputs: List<TxOut>, targetFeerate: FeeratePerKw): Satoshi {
```
pm47, line 154: This factorization will save us a lot of future headaches 🤞.

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
pm47: I think it's still wrong, because we are not handling the intersection. There has to be a `min` somewhere because we want the "common denominator".

src/commonMain/kotlin/fr/acinq/lightning/channel/Closing.kt
t-bast: Right, if that happens our peer can spend all outputs anyway because they know our revocation secret, so we're screwed...

src/commonMain/kotlin/fr/acinq/lightning/channel/Normal.kt
t-bast: Agreed, done in https://github.com/ACINQ/lightning-kmp/pull/438/commits/9b9810ffc67da8f0b40825b2126d1dc54ee6b0dd

src/commonMain/kotlin/fr/acinq/lightning/channel/Normal.kt
t-bast: I re-grouped them in https://github.com/ACINQ/lightning-kmp/pull/438/commits/9b9810ffc67da8f0b40825b2126d1dc54ee6b0dd
I think it's important to explicitly list every case, as `SpliceStatus.Aborted` must be converted to `SpliceStatus.None` on disconnection.

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -705,8 +705,10 @@ data class Commitments(
...
     fun receiveCommit(commits: List<CommitSig>, keyManager: KeyManager, log: MDCLogger): Either<ChannelException, Pair<Commitments, RevokeAndAck>> {
-        // first we make sure that we have exactly one commit_sig for each active commitment
-        if (commits.size != active.size) {
+        // We may receive more commit_sig than the number of active commitments, because there can be a race where we send splice_locked
+        // while our peer is sending us a batch of commit_sig. When that happens, we simply need to discard the commit_sig that belong
+        // to commitments we deactivated.
+        if (commits.size < active.size) {
             return Either.Left(CommitSigCountMismatch(channelId, active.size, commits.size))
         }
         val active1 = active.zip(commits).map {
```
pm47, line 714: There are two subtelties that make it work: ordering of sigs and behavior of `zip`.

```suggestion
        // Signatures are sent in order (most recent first), calling `zip` will drop trailing sigs that are for deactivated/pruned commitments.
        val active1 = active.zip(commits).map {
```

t-bast: I thought the behavior of zip was standard enough, but an explicit comment is always good :+1:

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
t-bast: I also added tests for this behavior in https://github.com/ACINQ/lightning-kmp/pull/438/commits/82ca8f5556ac46d8249909c9b34f35a0140b0fce

pm47: Very nice! I verified that the test indeed fails with the previous `deactivateCommitments`.

pm47 (approved):

PR #439, opened 2023-04-07
startTLS negotiation with expectedHostName

robbiehanson:
This PR adds an option to TcpSocket.TLS.TRUSTED_CERTIFICATES:

```kotlin
/** Used for Electrum servers when expecting a valid certificate */
data class TRUSTED_CERTIFICATES(
  /**
   * Specify an expectedHostName when it's different than the `host` value you used
   * within TcpSocket.Builder.connect(). This may be the case when using Tor.
   */
  val expectedHostName: String? = null // <- NEW
) : TLS()
```

This is needed by iOS to properly perform the TLS handshake when Tor is being used. Because the `socket.connect` method is passed the parameters for the proxy (e.g. host="127.0.0.1"). And the TLS validation subsequently fails, unless we can override the expected host name (e.g. expectedHostName="electrum.acinq.co").

### DIFF
--- a/PhoenixCrypto/PhoenixCrypto/Classes/NativeSocket.swift
+++ b/PhoenixCrypto/PhoenixCrypto/Classes/NativeSocket.swift
@@ -230,7 +230,36 @@ public class NativeSocket: NSObject {
 		
 		// NWConnection doesn't support startTLS (or SOCKS).
 		// So we have to manually code a workaround.
-		// Here's how it works:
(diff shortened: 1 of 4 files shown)


t-bast (approved):
I haven't tested it, but the code change looks good to me :+1:

dpad85 (approved):
Successfully tested on phoenix production branch.

We'll have to make a 1.4.5 release.

PR #440, opened 2023-04-11
Use a `Set<Tlv>` in tlv streams

t-bast:
TLV streams are inherently sets, not lists. We also take this opportunity to introduce a constructor using `vararg`, which is more convenient.


pm47 (dismissed):

t-bast:
Rebased.

pm47 (approved):

PR #441, opened 2023-04-13
Fix RBF attempts deactivation

t-bast:
There was an issue with the following scenario:

- we create fundingTx1
- we RBF it to create fundingTx2
- fundingTx1 confirms
- our peer sends splice_locked for fundingTx1

The previous code from `deactivateCommitments` would select fundingTx2 and deactivate fundingTx1, which is invalid. We fix that by preserving order among RBF attempts (even though they share the same fundingIndex).

There was another issue with the initial funding (index=0). Since there was no explicit locking from our peer, we were keeping all RBF attempts even after one confirmed and invalidated the other ones. We're never exchanging `commit_sig` so we should eagerly deactivate RBF attempts.

This change is also being added to eclair in https://github.com/ACINQ/eclair/pull/2632/commits/6be3220bd701e821dd32579a9c79283e2cbededf


pm47 (approved):
LGTM

PR #444, opened 2023-04-13
Handle exceptions when decoding lightning messages

### LINKED ISSUE #443: Catch exceptions thrown during lightning message decoding
We currently don't wrap `LightningMessage.decode` in a `try/catch` block, which means that invalid messages will create a crash.
We should probably add a `try/catch` inside the `decode` method.

t-bast:
When we receive a lightning message from the TCP connection, we must wrap the decoding logic in a try/catch, otherwise invalid messages will crash the coroutine (and potentially the wallet application).

We don't need to do that when reading messages that we previously decoded or encoded ourselves, that's why we didn't change the signature of the `LightningMessage.decode` function. But maybe it would be cleaner to make it return a `Try` or a `LightningMessage?`

Fixes #443


pm47 (approved):
LGTM

> But maybe it would be cleaner to make it return a Try or a LightningMessage?

I think it's fine that way.

PR #446, opened 2023-04-17
`KeyManager` cleanup

pm47:
The `KeyManager` is now simply a repository of keys, it is not responsible for signing transactions. 

The interface has been simplified and returns classes of type `NodeKeys`, `ChannelKeys` and `Bip84OnChainKeys`. The latter is used for final and swap-in wallets.


src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
```diff
@@ -3,168 +3,166 @@ package fr.acinq.lightning.crypto
...
-            paymentKey = privateKey(paymentPoint(channelKeyPath).path).privateKey,
-            delayedPaymentKey = privateKey(delayedPaymentPoint(channelKeyPath).path).privateKey,
-            htlcKey = privateKey(htlcPoint(channelKeyPath).path).privateKey,
-            revocationKey = privateKey(revocationPoint(channelKeyPath).path).privateKey,
-            shaSeed = shaSeed(channelKeyPath)
+            paymentKey = privateKey(channelKeyPrefix I hardened(2)),
+            delayedPaymentKey = privateKey(channelKeyPrefix I hardened(3)),
+            htlcKey = privateKey(channelKeyPrefix I hardened(4)),
+            revocationKey = privateKey(channelKeyPrefix I hardened(1)),
+            shaSeed = privateKey(channelKeyPrefix I hardened(5)).value.concat(1).sha256()
```
pm47, line 92: Hardcoded-paths are all defined in this one place to minimize potential for footgun.

src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
```diff
@@ -3,168 +3,166 @@ package fr.acinq.lightning.crypto
...
+        fun nodeKeyBasePath(chain: Chain) = when (chain) {
+            Chain.Regtest, Chain.Testnet -> KeyPath.empty I hardened(48) I hardened(0)
+            Chain.Mainnet -> KeyPath.empty I hardened(50) I hardened(0)
         }
     }
 }
+
+infix fun KeyPath.I(index: Long): KeyPath = this.append(index)
+
+infix fun KeyPath.I(other: KeyPath): KeyPath = this.append(other)
```
pm47, line 168: This is may not 100% orthodox, but `/` is forbiden as a function name, and it renders nicely in Intellij and makes the paths human-readable.

src/commonMain/kotlin/fr/acinq/lightning/channel/Channel.kt
```diff
@@ -369,6 +368,8 @@ sealed class ChannelStateWithCommitments : PersistedChannelState() {
     val isInitiator: Boolean get() = commitments.params.localParams.isInitiator
     val remoteNodeId: PublicKey get() = commitments.remoteNodeId
 
+    fun ChannelContext.channelKeys(): KeyManager.ChannelKeys = commitments.params.localParams.channelKeys(keyManager)
```
pm47, line 371: This method is not strictly necessary, but comes in handy and reduces boiler plate. Unfortunately we can't use it in state `WaitForFundingSigned`, but that's fine.

Note that keys are recomputes every single time. They could easily be cached if performance is an issue.

src/commonMain/kotlin/fr/acinq/lightning/crypto/Bolt3Derivation.kt
```diff
@@ -4,41 +4,45 @@ import fr.acinq.bitcoin.ByteVector32
...
     fun perCommitPoint(seed: ByteVector32, index: Long): PublicKey = perCommitSecret(seed, index).publicKey()
 
-    fun derivePrivKey(secret: PrivateKey, perCommitPoint: PublicKey): PrivateKey {
+    private fun derivePrivKey(secret: PrivateKey, perCommitPoint: PublicKey): PrivateKey {
         // secretkey = basepoint-secret + SHA256(per-commitment-point || basepoint)
         return secret + (PrivateKey(sha256(perCommitPoint.value + secret.publicKey().value)))
     }
 
-    fun derivePubKey(basePoint: PublicKey, perCommitPoint: PublicKey): PublicKey {
+    fun PrivateKey.derive(perCommitPoint: PublicKey): PrivateKey = derivePrivKey(this, perCommitPoint)
```
t-bast, line 22: I'm not sure we should define this directly on `PrivateKey`, it means that when we use this we don't necessarily know what kind of derivation this is doing (this is a very custom type of key derivation). I'd rather explicitly call `Bolt3Derivation.derivePrivKey` to keep this explicit. Same thing for `PublicKey.derive`.

Or maybe we could keep these extension methods but name them more explicitly? For example `PrivateKey.deriveForCommitment`?

src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
t-bast: Actually you can override `/`, which is a bit more intuitive:

```suggestion
infix operator fun KeyPath.div(index: Long): KeyPath = this.append(index)

infix operator fun KeyPath.div(other: KeyPath): KeyPath = this.append(other)
```

src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
```diff
@@ -1,128 +1,103 @@
...
+
+    data class Bip84OnChainKeys(
+        private val chain: NodeParams.Chain,
+        val account: Long,
+        val xpriv: DeterministicWallet.ExtendedPrivateKey
+    ) {
+        constructor(chain: NodeParams.Chain, master: DeterministicWallet.ExtendedPrivateKey, account: Long) : this(
+            chain, account,
+            xpriv = DeterministicWallet.derivePrivateKey(master, bip84BasePath(chain) I hardened(account))
+        )
```
t-bast, line 70: I think there's a confusion between this constructor and the default one, since they take the same parameters but in a different order and this one actually does one hardened derivation using `account`. For example, `val xpub` re-derives using `account`, which means we derived twice, which is incorrect?

Do we really need to keep the `account` field in the `data class`? Can't we have the following instead?

```kotlin
data class Bip84OnChainKeys(
        private val chain: NodeParams.Chain,
        val xpriv: DeterministicWallet.ExtendedPrivateKey
    ) {
        companion object {
            fun derive(chain: NodeParams.Chain, master: DeterministicWallet.ExtendedPrivateKey, account: Long) = Bip84OnChainKeys(
                chain,
                DeterministicWallet.derivePrivateKey(master, bip84BasePath(chain) I hardened(account))    
            )
        }
    }
```

src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
```diff
@@ -1,128 +1,103 @@
...
+            return Script.write(script).toByteVector()
+        }
+
+        fun address(addressIndex: Long): String {
+            return Bitcoin.computeP2WpkhAddress(privateKey(addressIndex).publicKey(), chain.chainHash)
         }
 
-        fun channelKeyPath(fundingPubKey: ExtendedPublicKey): KeyPath = channelKeyPath(fundingPubKey.publicKey)
+        val xpub: String = DeterministicWallet.encode(
+            input = DeterministicWallet.publicKey(DeterministicWallet.derivePrivateKey(xpriv, KeyPath.empty I hardened(account))),
```
t-bast, line 88: Shouldn't this just be:

```suggestion
            input = DeterministicWallet.publicKey(xpriv),
```

src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
```diff
@@ -3,168 +3,173 @@ package fr.acinq.lightning.crypto
...
+ * ```
+ *  node key:
+ *       50' / 0'
+ *
+ *  funding keys:
+ *   - initial funding tx (*):
+ *       50' / 1' / <fundingKeyPath> / <0' or 1'> / 0'
+ *   - splice funding txs (index > 0):
+ *       50' / 1' / <fundingKeyPath> / <0' or 1'> / 6' / <index>
+ *   (*) The initial funding tx is computed differently for historical reasons.
```
t-bast, line 30: Same as https://github.com/ACINQ/eclair/pull/2639, I think we should just do:

```suggestion
 *   - initial funding tx:
 *       50' / 1' / <fundingKeyPath> / <0' or 1'> / 0'
 *   - splice funding txs (index > 0):
 *       50' / 1' / <fundingKeyPath> / <0' or 1'> / <index'>
```

t-bast (commented):
A first pass mostly on the key derivation files.

src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
pm47: Done in 92eadb53.

src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
pm47: Nice catch! Fixed and added test in aa5dc819.

src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
pm47: I think it's useful to expose the account to the library integrators. I refactored in 5dfb0556, which cleared up the confusion.

src/commonMain/kotlin/fr/acinq/lightning/crypto/Bolt3Derivation.kt
pm47: Went with `deriveForCommitment`/`deriveForRevocation` in cc3eaac9.

src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
pm47: Agreed. I've updated the doc in 701ea979, but the actual implementation of the funding tx derivation is done in the next PR.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -744,21 +738,7 @@ class Peer(
...
-                                    nodeParams.dustLimit,
-                                    nodeParams.maxHtlcValueInFlightMsat,
-                                    nodeParams.htlcMinimum,
-                                    nodeParams.toRemoteDelayBlocks,
-                                    nodeParams.maxAcceptedHtlcs,
-                                    false,
-                                    closingPubkeyScript.toByteVector(),
-                                    features
-                                )
+                                val localParams = LocalParams(nodeParams, isInitiator = false)
```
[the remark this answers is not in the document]
t-bast: That's strange, I can't remember why we did that (it was like this for channels even before dual-funding)...I believe we want to fix it and use the 8 bytes path provided by `newFundingKeyPath`.

t-bast (commented):
Looks mostly good!

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -80,10 +79,11 @@ data class WalletState(val addresses: Map<String, List<UnspentItem>>, val parent
         }
 
         /** Find the private key corresponding to this script, assuming this is a p2wpkh owned by us. */
-        private fun script2PrivateKey(keyManager: KeyManager, publicKeyScript: ByteVector): PrivateKey? {
-            val priv = keyManager.bip84PrivateKey(account = 1, addressIndex = 0)
-            val script = Script.write(Script.pay2wpkh(priv.publicKey())).toByteVector()
-            return if (script == publicKeyScript) priv else null
+        private fun script2PrivateKey(onChainKeys: KeyManager.Bip84OnChainKeys, publicKeyScript: ByteVector): PrivateKey? {
+            for (addressIndex in 0L..100L) {
```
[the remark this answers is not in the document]
pm47: Right, we are kind of in-between on this. I left the loop to show the intent but restricted to index 0 in 7d305ffb.

src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
```diff
@@ -3,168 +3,168 @@ package fr.acinq.lightning.crypto
...
+            val buffer = fundingPubKey.value.sha256().toByteArray()
+
+            val path = sequence {
+                var i = 0
+                while (true) {
+                    yield(Pack.int32BE(buffer, i).toUInt().toLong())
+                    i += 4
+                }
+            }
+            return KeyPath(path.take(8).toList())
```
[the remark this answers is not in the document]
pm47: Added in 0b1a5886, also reverified independently.

src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
```diff
@@ -3,168 +3,173 @@ package fr.acinq.lightning.crypto
...
+     */
     companion object {
 
-        fun channelKeyBasePath(chainHash: ByteVector32) = when (chainHash) {
-            Block.RegtestGenesisBlock.hash, Block.TestnetGenesisBlock.hash -> listOf(hardened(48), hardened(1))
-            Block.LivenetGenesisBlock.hash -> listOf(hardened(50), hardened(1))
-            else -> throw IllegalArgumentException("unknown chain hash $chainHash")
+        /**
+         * Create a BIP32 path from a public key. This path will be used to derive channel keys.
+         * Having channel keys derived from the funding public keys makes it very easy to retrieve your funds when've you've lost your data:
```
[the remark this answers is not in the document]
pm47: Fixed in 3fc10e1b.

src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
```diff
@@ -3,168 +3,173 @@ package fr.acinq.lightning.crypto
...
-            prefix = when (chainHash) {
-                Block.LivenetGenesisBlock.hash -> DeterministicWallet.zpub
-                Block.TestnetGenesisBlock.hash -> DeterministicWallet.vpub
-                else -> throw IllegalArgumentException("unhandled chain hash $chainHash")
-            }
-        )
-    }
+    /**
+     * This method offers direct access to the master key derivation. It should only be used for some advanced usage
+     * like (LNURL-auth, data encryption.
```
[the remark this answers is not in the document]
pm47: Fixed in 3fc10e1b.

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -417,9 +418,9 @@ data class Commitment(
         // signatures are now optional in the commit message, and will be sent only if the other party is actually
         // receiving money i.e its commit tx has one output for them
         val spec = CommitmentSpec.reduce(localCommit.spec, changes.localChanges.acked, changes.remoteChanges.proposed)
-        val localPerCommitmentPoint = keyManager.commitmentPoint(params.localParams.channelKeys(keyManager).shaSeed, localCommit.index + 1)
-        val (localCommitTx, htlcTxs) = Commitments.makeLocalTxs(keyManager.channelKeys(params.localParams.fundingKeyPath), localCommit.index + 1, params.localParams, params.remoteParams, commitInput, localPerCommitmentPoint, spec)
-        val keys = params.localParams.channelKeys(keyManager)
+        val keys = channelKeys
```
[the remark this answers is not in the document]
pm47: Removed in 603a21f4.

pm47:
🎉

PR #453, opened 2023-05-03
Persist channel state on `tx_abort` in status `WaitForTxSigned`

pm47:
We must persist the channel state if a splice is aborted in status `WaitForTxSigned`.


t-bast (approved):
Good catch!

PR #454, opened 2023-05-04
Remove `ErrorInformationLeak` state and fix channel recovery

pm47:
Splicing makes the handling of multi devices more difficult, because an unknown spending tx could in fact be a splice tx. There was no real utility to the `InformationLeak` state so we remove it entirely.

There is another subtelty: when we recover from an outdated state, we also need to update the watches, because the recovered state may be one or several splices ahead. The easiest way to do that is to recreate the channel, just like if we recovered from zero.

This leads to a nice refactoring of the recovery process, where both cases (recover from scratch / recover from outdated) happen at the same place, in the Peer.


t-bast (commented):
LGTM, I slightly improved the peer tests in https://github.com/ACINQ/lightning-kmp/commit/8d4da59bd9505e595462b9548b1faa4638aa41f5 to verify that the restored state matches the latest state for which we sent a backup. Can you cherry-pick that commit?

pm47:
> I slightly improved the peer tests in https://github.com/ACINQ/lightning-kmp/commit/8d4da59bd9505e595462b9548b1faa4638aa41f5 to verify that the restored state matches the latest state for which we sent a backup. Can you cherry-pick that commit?

Thanks, done in [f9a574a](https://github.com/ACINQ/lightning-kmp/pull/454/commits/f9a574a36bdd4b536968ec56ba8d6e042b120c90).

t-bast (approved):

PR #456, opened 2023-05-05
Adjust splice feerate with unconfirmed parents

pm47:
Also expose a `spliceCpfp()` method which doesn't take any input or output.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -407,6 +407,10 @@ class Peer(
     /**
      * Estimate the actual feerate to use (and corresponding fee to pay) in order to reach the target feerate
      * for a cpfp splice.
+     * @return The adjusted feerate to use in [spliceCpfp], such that the whole transaction chain has a feerate equivalent
+     *         to [targetFeerate].
+     *         NB: if the output feerate is equal to the input feerate then the cpfp is useless and
+     *         should not be attempted.
      */
     suspend fun estimateFeeForSpliceCpfp(channelId: ByteVector32, targetFeerate: FeeratePerKw): Pair<FeeratePerKw, Satoshi>? {
         return channels.values


src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClient.kt
```diff
@@ -212,31 +212,34 @@ class ElectrumClient(
...
+    override suspend fun send(request: ElectrumRequest, replyTo: CompletableDeferred<ElectrumResponse>) {
         mailbox.send(Action.SendToServer(Pair(request, replyTo)))
     }
 
     suspend inline fun <reified T : ElectrumResponse> rpcCall(request: ElectrumRequest): T {
         val replyTo = CompletableDeferred<ElectrumResponse>()
         send(request, replyTo)
-        return replyTo.await() as T
+        return when (val res = replyTo.await()) {
+            is ServerError -> error(res)
```
t-bast, line 223: Won't it crash the wallet if we throw an exception here? There are a lot of calls that we seem to make without catching exceptions.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClientExtensions.kt
```diff
@@ -0,0 +1,49 @@
...
+    return tx?.let { getConfirmations(tx) }
+}
+
+suspend fun IElectrumClient.getConfirmations(tx: Transaction): Int? {
+    val scriptHash = ElectrumClient.computeScriptHash(tx.txOut.first().publicKeyScript)
+    val scriptHashHistory = getScriptHashHistory(scriptHash)
+    val item = scriptHashHistory.find { it.txid == tx.txid }
+    val blockHeight = startHeaderSubscription().blockHeight
+    return item?.let { if (item.blockHeight > 0) blockHeight - item.blockHeight + 1 else 0 }
+}
```
t-bast, line 25: We could probably return a non-null value here: if anything fails, that means the transaction is not in the blockchain and thus has 0 confirmations?

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClientExtensions.kt
```diff
@@ -0,0 +1,49 @@
...
+                is LocalFundingStatus.UnconfirmedFundingTx -> commitment.localFundingStatus.signedTx?.weight() ?: commitment.localFundingStatus.sharedTx.tx.buildUnsignedTx().weight()
+                is LocalFundingStatus.ConfirmedFundingTx -> commitment.localFundingStatus.signedTx.weight()
+            }
+            Pair(parentsWeight + weight, parentsFees + commitment.localFundingStatus.fee)
+        }
+    val totalWeight = parentsWeight + spliceWeight
+    val totalFees = Transactions.weight2fee(targetFeerate, totalWeight)
+    val actualFee = totalFees - parentsFees
+    val minFeerate = Transactions.fee2rate(actualFee, spliceWeight)
+    // if ancestors have a higher feerate than target, min feerate could be negative
```
t-bast, line 42: Isn't this a case where we would cancel that CPFP, because it's actually useless (and makes our transaction chain longer)? More precisely, if `Transactions.fee2rate(parentsFee, parentsWeight) >= targetFeerate` we shouldn't do a CPFP?

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClient.kt
pm47: Yes it would, but this change doesn't introduce a new behavior.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClientExtensions.kt
t-bast: True, but we should document the fact that `null` should be interpreted as `unconfirmed` as well.

pm47: Improved the comment here: d206a452.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClientExtensions.kt
pm47: So, this would have to be done in `Peer.estimateFeeForSpliceCpfp`, because this estimation is also used for regular splice-in and splice-out.

However, we would then need additional typing to allow integrators to make a difference between error cases (e.g. channel not found), and invalid feerate. This would then lead to a general refactoring of other methods related to splices in `Peer` for consistency. Maybe simpler to let the integrators handle this corner case for now?

pm47:
Rebased on master.

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
```diff
@@ -127,7 +127,7 @@ data class IncomingPayment(val preimage: ByteVector32, val origin: Origin, val r
     override val completedAt: Long?
         get() = when {
             received == null -> null // payment has not yet been received
-            received.receivedWith.any { it is ReceivedWith.OnChainIncomingPayment && it.confirmedAt == null } -> null // payment has been received, but there is at least one unconfirmed on-chain part
+            received.receivedWith.any { it is ReceivedWith.OnChainIncomingPayment && it.lockedAt == null } -> null // payment has been received, but there is at least one unconfirmed on-chain part
```
t-bast, line 130: Shouldn't we still be looking at `confirmedAt` here?

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -876,7 +933,10 @@ class Peer(
...
                 when (val channel = channels.values.firstOrNull { it is Normal }) {
                     is ChannelStateWithCommitments -> {
-                        val feerate = onChainFeeratesFlow.filterNotNull().first().fundingFeerate
+                        val targetFeerate = onChainFeeratesFlow.filterNotNull().first().fundingFeerate
+                        val weight = FundingContributions.computeWeightPaid(isInitiator = true, commitment = channel.commitments.active.first(), walletInputs = cmd.wallet.confirmedUtxos, localOutputs = emptyList())
+                        val (feerate, _) = watcher.client.computeSpliceCpfpFeerate(channel.commitments, targetFeerate, spliceWeight = weight, logger)
+
                         logger.info { "requesting splice-in using confirmed balance=$balance feerate=$feerate" }
 
                         val feeEstimate = Transactions.weight2fee(feerate, 610 + cmd.wallet.confirmedUtxos.size * Transactions.p2wpkhInputWeight)
```
t-bast, line 942: We should use `weight` here instead of this hard-coded value.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClientExtensions.kt
t-bast: Sounds good, then we should probably update the comment of the `estimateFeeForSpliceCpfp` function to tell callers that if the feerate it returns is the same as the feerate they provided as input, it means that the CPFP attempt will be useless and they shouldn't make it.

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
pm47: When defined, `completedAt` means that the received payment is usable, so it's indeed `lockedAt`.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
pm47: We should use `weight` instead of `610 + cmd.wallet.confirmedUtxos.size * Transactions.p2wpkhInputWeight` right?

t-bast: Exactly, and we probably don't even need this `feeEstimate`: can't we use the fee provided by `computeSpliceCpfpFeerate`?

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
pm47: That field is also used for pure LN payments, where there is no concept of locking.

t-bast: But you agree it's very confusing for readers of that code (and most likely for wallets integrating it as well)? So we should do something, at least a comment to explain that confusion until we find a better way of expressing the distinction between `locked` / `confirmed` in terms of UX

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
pm47: Indeed, I noticed that too. It's now much cleaner in c4d3cd6d.

src/commonTest/kotlin/fr/acinq/lightning/channel/states/SpliceTestsCommon.kt
```diff
@@ -42,6 +42,12 @@ class SpliceTestsCommon : LightningTestSuite() {
         spliceIn(alice, bob, listOf(30_000.sat, 40_000.sat, 25_000.sat))
     }
 
+    @Test
+    fun `splice cpfp`() {
+        val (alice, bob) = reachNormal()
+        spliceCpfp(alice, bob)
```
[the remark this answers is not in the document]
pm47: I added a couple splices in 020405d1, but the goal of this test is only to make sure that we supported a splice with no input and no output. It would be even more interesting to check feerates, but that requires testing at the peer level, which is quite hacky currently and needs a more general improvement.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClientExtensions.kt
pm47: Done in 2be2abe9.

PR #457, opened 2023-05-09
Upgrade to kotlin 1.8

sstone:


dpad85:
I've been able to test this branch (locally merged with `splices-cpfp`) on Phoenix Android. A few libs had to be upgraded (e.g. SQLDelight to 1.3.5) to support kotlin 1.8, but otherwise there have been no issues.

We'll have to test on iOS as well but this looks good.

dpad85:
Tests also look good on iOS. There are a few Swift-related changes that will need Robbie's touch, but no blockers. I think we can go ahead with kotlin 1.8 on lightning-kmp 👍

sstone:
Nice! I'll publish released version of secp256k1-kmp and bitcoin-kmp and update this PR when it's done

build.gradle.kts
```diff
@@ -14,7 +14,7 @@ allprojects {
     version = "1.5.0-SNAPSHOT"
 
     repositories {
-        mavenLocal()
+        //mavenLocal()
```
pm47, line 17: Why is this commented?

publishing/lightning-kmp-snapshot-deploy.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash -x
```
pm47, line 1: What is the purpose of this file?

build.gradle.kts
sstone: I added a comment in https://github.com/ACINQ/lightning-kmp/pull/457/commits/4798b41df3534a750c0474746de9da69bdfe4488 (we could also just remove this line)

publishing/lightning-kmp-snapshot-deploy.sh
sstone: To publish snapshot libraries to sonatype so they can be available globally (for other devs/machines, github CI, ...)

pm47 (approved):

pm47:
Thanks!

PR #460, opened 2023-05-15
Remove `PleaseOpenChannelRejected`

pm47:
The rejection is now done on lightning-kmp's side, based on the local liquidity policy.


t-bast (approved):
Less code!

PR #462, opened 2023-05-16
Fix computation of fee in cpfp feerate estimator

pm47:
If user attempts a cpfp with a smaller target feerate than the feerate of the existing ancestors, the projected feerate would be negative. This is why we correct it by taking the max between the requested feerate and the projecte feerate.

But we forgot to do that for the absolute fee value.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClientExtensions.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClientExtensions.kt
@@ -40,13 +40,15 @@ suspend fun IElectrumClient.computeSpliceCpfpFeerate(commitments: Commitments, t
         }
     val totalWeight = parentsWeight + spliceWeight
     val totalFees = Transactions.weight2fee(targetFeerate, totalWeight)
-    val actualFee = totalFees - parentsFees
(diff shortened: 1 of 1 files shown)


t-bast (approved):
I don't think this is relevant for CPFP, because we simply shouldn't do any CPFP in that case, right?

However, I see why it could be useful for splice-out, where the user may want to use a smaller feerate than the ancestor feerate (because they're not trying to CPFP, they're just making an on-chain payment).

pm47:
> I don't think this is relevant for CPFP, because we simply shouldn't do any CPFP in that case, right?
> However, I see why it could be useful for splice-out, where the user may want to use a smaller feerate than the ancestor feerate (because they're not trying to CPFP, they're just making an on-chain payment).

Correct, this is mainly fixing the fee presented to the user when they attempt a splice-out.

dpad85:
> I don't think this is relevant for CPFP, because we simply shouldn't do any CPFP in that case, right?

We now check in Phoenix (android at least) that the effective feerate would not be below what the user intends (@robbiehanson see [this commit](https://github.com/ACINQ/phoenix/blob/7c59cfe2a243b2a385a413a914b19b93257d1a3b/phoenix-android/src/main/kotlin/fr/acinq/phoenix/android/payments/cpfp/CpfpViewModel.kt#L75)). I think that will prevent such issues from happening.

PR #468, opened 2023-05-31
Phoenix 2-of-2 swap in

t-bast:
This PR implements a new kind of swap-in based on [swap-in-potentiam](https://lists.linuxfoundation.org/pipermail/lightning-dev/2023-January/003810.html):

- users swap funds to a script that can be spent by either a 2-of-2 multisig with the LSP, or unilaterally by them after a relative delay
- once those swaps are confirmed, we can use the corresponding utxos in dual-funding / splicing with zero-conf, since the user cannot unilaterally double-spend

Note that the serialization changes are *not* backwards-compatible: wallets with channels using dual funding / splicing will need to be reset.

This is best reviewed commit-by-commit (commit messages should be detailed enough).


src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
```diff
@@ -177,8 +177,7 @@ data class Normal(
                             val (signingSession1, action) = spliceStatus.session.receiveCommitSig(channelKeys(), commitments.params, cmd.message, currentBlockHeight.toLong())
                             when (action) {
                                 is InteractiveTxSigningSessionAction.AbortFundingAttempt -> {
-                                    logger.warning { "splice attempt failed: ${action.reason.message}" }
-                                    logger.warning { "funding tx was txId=${spliceStatus.session.fundingTx.txId} tx=${spliceStatus.session.fundingTx.tx.buildUnsignedTx()}" }
+                                    logger.warning { "splice attempt failed: ${action.reason.message} (fundingTxId=${spliceStatus.session.fundingTx.txId})" }
```
pm47, line 180: The full log was useful for debugging, but I guess we are past that.

src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
```diff
@@ -692,7 +694,7 @@ data class Normal(
             // If we added some funds ourselves it's a swap-in
             if (action.fundingTx.sharedTx.tx.localInputs.isNotEmpty()) add(
                 ChannelAction.Storage.StoreIncomingPayment.ViaSpliceIn(
-                    amount = action.fundingTx.sharedTx.tx.localInputs.map { i -> i.previousTx.txOut[i.previousTxOutput.toInt()].amount }.sum().toMilliSatoshi() - action.fundingTx.sharedTx.tx.fees.toMilliSatoshi(),
+                    amount = action.fundingTx.sharedTx.tx.localInputs.map { i -> i.txOut.amount }.sum().toMilliSatoshi() - action.fundingTx.sharedTx.tx.fees.toMilliSatoshi(),
```
pm47, line 697: Nice!

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -770,17 +828,13 @@ data class InteractiveTxSigningSession(
...
-                    null -> Either.Left(ChannelFundingError(channelParams.channelId))
-                    else -> {
-                        val unsignedLocalCommit = UnsignedLocalCommit(commitmentIndex, firstCommitTx.localSpec, firstCommitTx.localCommitTx, listOf())
-                        val remoteCommit = RemoteCommit(commitmentIndex, firstCommitTx.remoteSpec, firstCommitTx.remoteCommitTx.tx.txid, remotePerCommitmentPoint)
-                        Either.Right(Pair(InteractiveTxSigningSession(fundingParams, fundingTxIndex, signedFundingTx, Either.Left(unsignedLocalCommit), remoteCommit), commitSig))
-                    }
-                }
+                val unsignedLocalCommit = UnsignedLocalCommit(commitmentIndex, firstCommitTx.localSpec, firstCommitTx.localCommitTx, listOf())
+                val remoteCommit = RemoteCommit(commitmentIndex, firstCommitTx.remoteSpec, firstCommitTx.remoteCommitTx.tx.txid, remotePerCommitmentPoint)
+                val signedFundingTx = sharedTx.sign(keyManager, fundingParams, channelParams.localParams)
```
pm47, line 836: Looks much nicer. I guess the idea is that if the signatures don't match an error will be raised by the peer instead of here with the `ChannelFundingError(channelParams.channelId)`?

src/commonMain/kotlin/fr/acinq/lightning/wire/InteractiveTxTlv.kt
```diff
@@ -24,6 +22,25 @@ sealed class TxAddInputTlv : Tlv {
             override fun read(input: Input): SharedInputTxId = SharedInputTxId(LightningCodecs.bytes(input, 32).toByteVector32().reversed())
         }
     }
+
+    /** When adding a swap-in input to an interactive-tx, the user needs to provide the corresponding script parameters. */
+    data class SwapInParams(val userKey: PublicKey, val serverKey: PublicKey, val refundDelay: Int) : TxAddInputTlv() {
```
pm47, line 27: 👍

pm47 (commented):
LGTM

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
t-bast: Exactly, we checked that we have the correct number of signatures, and we will verify that the transaction correctly spends its inputs later and fail at that point if one of the signatures was invalid. If we forgot to send some of our signatures, our peer will send back an error.

t-bast:
Rebased on master.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -200,26 +202,38 @@ class Peer(
         }
         launch {
             finalWallet.walletStateFlow
-                .distinctUntilChangedBy { Pair(it.unconfirmedBalance, it.confirmedBalance) }
+                .distinctUntilChangedBy { it.totalBalance }
```
pm47, line 205: This is not equivalent, the previous code displayed a log line when a tx was confirmed.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -200,26 +202,38 @@ class Peer(
...
-                    val balance = availableWallet.confirmedBalance
-                    logger.info { "swap-in wallet balance: confirmed=$balance, unconfirmed=${availableWallet.unconfirmedBalance}" }
-                    if (balance > 0.sat) {
-                        logger.info { "swap-in wallet: requesting channel using confirmed balance: $balance" }
-                        input.send(RequestChannelOpen(Lightning.randomBytes32(), availableWallet))
-                        reservedUtxos.union(availableWallet.confirmedUtxos)
-                    } else {
-                        reservedUtxos
-                    }.intersect(wallet.utxos.toSet()) // drop utxos no longer in wallet
+                    when (val currentBlockHeight = currentTipFlow.value?.first) {
```
pm47, line 214: I'd rather remove this `when` and simply do `val currentBlockHeight = currentTipFlow.filterNotNull().first().first`. It's simpler, consistent with the rest of the file, and that case will never happen anyway since if we receive a balance event we have to be connected in the first place.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -200,26 +202,38 @@ class Peer(
...
+                            logger.warning { "cannot process wallet state: current block height not available" }
+                            reservedUtxos
+                        }
+                        else -> {
+                            // reservedUtxos are part of a previously issued RequestChannelOpen command
+                            val availableWallet = wallet.minus(reservedUtxos).withConfirmations(currentBlockHeight, walletParams.swapInConfirmations)
+                            logger.info { "swap-in wallet balance: deeplyConfirmed=${availableWallet.deeplyConfirmed.balance}, weaklyConfirmed=${availableWallet.weaklyConfirmed.balance}, unconfirmed=${availableWallet.unconfirmed.balance}" }
+                            val utxos = when {
+                                // When migrating from the legacy android app, we use all utxos, even unconfirmed ones.
+                                isMigrationFromLegacyApp -> availableWallet.all
```
pm47, line 225: I think logging differently based on `isMigrationFromLegacyApp` like I did on my branch could help debugging.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
t-bast: True, but do we need that?

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
t-bast: We can't, the compiler won't let me do that because `.first()` needs a `suspend`...but if you find the right way to do this, that's what I initially wanted to write!

pm47: This code is already inside a coroutine so the compiler shouldn't give you a hard time.

The following compiles on my machine™:

```kotlin
        launch {
            swapInWallet.walletStateFlow.filter { it.consistent }.fold(emptySet<WalletState.Utxo>()) { reservedUtxos, wallet ->
                    val currentBlockHeight = currentTipFlow.filterNotNull().first().first
                    // reservedUtxos are part of a previously issued RequestChannelOpen command
                    val availableWallet = wallet.minus(reservedUtxos).withConfirmations(currentBlockHeight, walletParams.swapInConfirmations)
                    logger.info { "swap-in wallet balance: deeplyConfirmed=${availableWallet.deeplyConfirmed.balance}, weaklyConfirmed=${availableWallet.weaklyConfirmed.balance}, unconfirmed=${availableWallet.unconfirmed.balance}" }
                    val utxos = when {
                        // When migrating from the legacy android app, we use all utxos, even unconfirmed ones.
                        isMigrationFromLegacyApp -> availableWallet.all
                        else -> availableWallet.deeplyConfirmed
                    }
                    if (utxos.balance > 0.sat) {
                        logger.info { "swap-in wallet: requesting channel using ${utxos.size} utxos with balance=${utxos.balance}" }
                        input.send(RequestChannelOpen(Lightning.randomBytes32(), utxos))
                        reservedUtxos.union(utxos.utxos)
                    } else {
                        reservedUtxos
                    }.intersect(wallet.utxos.toSet()) // drop utxos no longer in wallet
                }
        }
```

t-bast: I don't know what I previously did, that's exactly what I tried but the compiler was complaining...I must have gotten it wrong, fixed in https://github.com/ACINQ/lightning-kmp/pull/468/commits/e9dbdbcc3d4b5c602d27de20e24cf066c64b70aa

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
pm47: When we get tired of seeing `(migration=false)` in the logs we'll remove that special case.

src/commonMain/kotlin/fr/acinq/lightning/channel/states/WaitForOpenChannel.kt
```diff
@@ -24,7 +24,7 @@ data class WaitForOpenChannel(
     val temporaryChannelId: ByteVector32,
     val fundingAmount: Satoshi,
     val pushAmount: MilliSatoshi,
-    val wallet: WalletState,
+    val utxos: List<WalletState.Utxo>,
```
pm47, line 27: I'm nitpicking but it would really be clearer if those were named `walletInputs` in all commands and channel states

```suggestion
    val walletInputs: List<WalletState.Utxo>,
```

pm47 (approved):

pm47 (approved):

PR #469, opened 2023-05-31
Fix channel reestablish in some corner cases

pm47:
The main change is re-sending `commit_sig`/`tx_signatures` in `WaitChannelReady`, which is needed in the zero-conf case.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Syncing.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Syncing.kt
@@ -61,9 +61,9 @@ data class Syncing(val state: PersistedChannelState) : ChannelState() {
                                                 state.commitments.params,
                                                 fundingTxIndex = 0,
                                                 state.commitments.latest.remoteFundingPubkey,
-                                                state.commitments.latest.commitInput)
-                                            add(ChannelAction.Message.Send(commitSig)
+                                                state.commitments.latest.commitInput
                                             )
+                                            add(ChannelAction.Message.Send(commitSig))
                                         }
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/states/Syncing.kt
```diff
@@ -56,9 +56,14 @@ data class Syncing(val state: PersistedChannelState) : ChannelState() {
...
-                                                state.commitments.latest.remoteCommit.sign(channelKeys(), state.commitments.params, fundingTxIndex = 0, state.commitments.latest.remoteFundingPubkey, state.commitments.latest.commitInput)
-                                            add(ChannelAction.Message.Send(commitSig))
+                                            val commitSig = state.commitments.latest.remoteCommit.sign(
+                                                channelKeys(),
+                                                state.commitments.params,
+                                                fundingTxIndex = 0,
+                                                state.commitments.latest.remoteFundingPubkey,
+                                                state.commitments.latest.commitInput)
+                                            add(ChannelAction.Message.Send(commitSig)
+                                            )
```
t-bast, line 66:
```suggestion
                                            add(ChannelAction.Message.Send(commitSig))
```

pm47:
> We should add a test in WaitForChannelReadyTestsCommon, similar to the disconnection tests in SpliceTestsCommon. This should be easy to add since the fixture in WaitForChannelReadyTestsCommon is already in a state where:

Done in 36a54691. I had to use `WaitForFundingSignedTestsCommon` because we need to be in an assymetrical state where only one side has received the `commit_sig`.

t-bast (approved):

PR #473, opened 2023-06-07
Skip waiting for reestablish for unsigned channels

t-bast:
When we are a backup client, we wait for our peer to send their `channel_reestablish` first to make sure we haven't lost data.

This can lead to a deadlock if we're waiting for a `channel_reestablish`, but our peer has forgotten that channel because we got disconnected before they were able to send their signatures.

To avoid this deadlock, we send our `channel_reestablish` immediately for channels where we're waiting for signatures.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Syncing.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Syncing.kt
@@ -1,6 +1,5 @@
 package fr.acinq.lightning.channel.states
 
-import fr.acinq.lightning.Feature
 import fr.acinq.lightning.ShortChannelId
 import fr.acinq.lightning.blockchain.*
 import fr.acinq.lightning.channel.*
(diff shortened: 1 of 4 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/states/Syncing.kt
```diff
@@ -270,9 +270,12 @@ data class Syncing(val state: PersistedChannelState) : ChannelState() {
                     }
                     else -> unhandled(cmd)
                 }
+                val sendChannelReestablish = when (state) {
```
pm47, line 273: We could add this boolean in the `Syncing` state, which is not persisted. It would save us from keeping consistency between `Offline` and `Syncing`.

t-bast: Good idea, done in https://github.com/ACINQ/lightning-kmp/pull/473/commits/80045cda2602f7d28100f95b430dee46b013decd

PR #474, opened 2023-06-07
Avoid double-spending channels on restart

t-bast:
When we restart and scan our swap-in address, we must make sure we don't initiate a channel funding with utxos that we're already using in concurrent sessions, otherwise we will end up double-spending ourselves which we cannot easily detect after the fact, leading to channels that appear stuck.

This is a bit hard to test since we don't have good ways of testing the wallet-watching coroutine. I'll add some tests in our integration tests with `eclair`, but I'm not sure what we can do to better unit test this here.


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -208,11 +208,16 @@ class Peer(
                 }
         }
         launch {
+            // Wallet utxos that are already used in channel funding attempts should be ignored, otherwise we would double-spend ourselves.
+            // If the electrum server we connect to has our channel funding attempts in their mempool, those utxos wouldn't be added to our wallet anyway.
+            // But we cannot rely only on that, since we may connect to a different electrum server after a restart, or transactions may be evicted from their mempool.
+            // Since we don't have an easy way of asking electrum to check for double-spends, we would end up with channels that are stuck waiting for confirmations.
+            // This generally wouldn't be a security issue (only one of the funding attempts would succeed), unless 0-conf is used and our LSP is malicious.
+            val initiallyReservedUtxos: Set<OutPoint> = Helpers.reservedWalletInputs(db.channels.listLocalChannels())
```
t-bast, line 216: We're now reading channels from the DB twice, here and in the coroutine below. Should we group this coroutine and the one below as children of a shared coroutine to list channels only once?

pm47: Also, we could be disconnected and switch electrum servers while the app stays up, meaning that for best effort we should check channels everytime we are notified of a new wallet state.

A solution could be:
1) isolate the spawning of the swap-in coroutine in a method, say `launchSwapInWallet()`
2) call `launchSwapInWallet()` after `db.channels.listLocalChannels()` in the coroutine that reads the db (it would fit nicely near the "check htlc timeout" coroutine)
3) call `Helpers.reservedWalletInputs(_channels.value)` in the swap-in coroutine

WDYT?

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -24,10 +24,9 @@ data class WalletState(val addresses: Map<String, List<UnspentItem>>, val parent
         .map { Utxo(parentTxs[it.txid]!!, it.outputIndex, it.blockHeight) }
     val totalBalance = utxos.map { it.amount }.sum()
 
-    fun minus(reserved: Set<Utxo>): WalletState {
-        val reservedIds = reserved.map { UnspentItemId(it.previousTx.txid, it.outputIndex) }.toSet()
+    fun minus(reserved: Set<OutPoint>): WalletState {
```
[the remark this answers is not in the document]
t-bast: But it makes a lot more sense though...we had to create an otherwise unused type `UnspentItemId`!

src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
```diff
@@ -246,6 +244,38 @@ object Helpers {
...
+                    val fundingInputs = channel.commitments.all.flatMap { commitment ->
+                        when (val fundingStatus = commitment.localFundingStatus) {
+                            is LocalFundingStatus.UnconfirmedFundingTx -> fundingStatus.sharedTx.tx.localInputs.map { it.outPoint }
+                            is LocalFundingStatus.ConfirmedFundingTx -> listOf() // utxos spent by confirmed transactions will never appear in our wallet
+                        }
+                    }
+                    unsignedInputs + fundingInputs
+                }
+            }
+        }.toSet()
```
pm47, line 276: Same behavior, written in two separate stages instead of imbricated. I think it conveys the intent better.

```suggestion
        val fundingTxs: List<SignedSharedTransaction> = buildList {
            for (channel in channels) {
                // add all unsigned inputs currently used to build a funding tx (creation, rbf, splice)
                when {
                    channel is WaitForFundingSigned -> add(channel.signingSession.fundingTx)
                    channel is WaitForFundingConfirmed && channel.rbfStatus is RbfStatus.WaitingForSigs -> channel.rbfStatus.session.fundingTx
                    channel is Normal && channel.spliceStatus is SpliceStatus.WaitingForSigs -> channel.spliceStatus.session.fundingTx
                    else -> {}
                }
                // add all inputs in unconfirmed funding txs (utxos spent by confirmed transactions will never appear in our wallet)
                when {
                    channel is ChannelStateWithCommitments -> channel.commitments.all
                        .map { it.localFundingStatus }
                        .filterIsInstance<LocalFundingStatus.UnconfirmedFundingTx>()
                        .forEach { add(it.sharedTx) }
                    else -> {}
                }
            }
        }
        val localInputs = fundingTxs.flatMap { fundingTx -> fundingTx.tx.localInputs.map { it.outPoint } }
        return localInputs.toSet()
```

t-bast: This is better, thanks! Done in https://github.com/ACINQ/lightning-kmp/pull/474/commits/88f3c72e4bcc145faccfcb236df7bc3b4512e475

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
t-bast: I applied 1) and 2) in https://github.com/ACINQ/lightning-kmp/pull/474/commits/88f3c72e4bcc145faccfcb236df7bc3b4512e475

As discussed offline, even if we disconnect from electrum, we don't need to re-scan our channels because we already added the utxos for each `RequestChannelOpen` we made since launch to our `reservedUtxos` set.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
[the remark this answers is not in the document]
t-bast: That makes sense, done in https://github.com/ACINQ/lightning-kmp/pull/474/commits/eba5b38861c984b98b5621e8c4cb22923ce42996

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -395,6 +374,34 @@ class Peer(
         listen() // This suspends until the coroutines is cancelled or the socket is closed
     }
 
+    private fun launchSwapInWallet(channels: List<PersistedChannelState>) = launch {
```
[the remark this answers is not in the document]
t-bast: Agreed, it bothered me a bit that this wasn't consistent with the `checkHtlcTimeout` block. Done in https://github.com/ACINQ/lightning-kmp/pull/474/commits/59b458c91fa599bf333a2b6d5e9504f345aef160

pm47 (approved):

PR #475, opened 2023-06-07
Do not recursively call `ChannelState.process`

pm47:
Otherwise reasoning about transitions becomes very difficult and error-prone.

In this instance, we were emitting a duplicate `StoreOutgoingPayment.ViaClose` when force-closing in offline `Offline` (reported by @dpad85). It is safer to directly call the `handleLocalError()` function, and also more consistent with e.g. `checkHtlcTimeout()`.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Offline.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Offline.kt
@@ -5,6 +5,7 @@ import fr.acinq.lightning.ShortChannelId
 import fr.acinq.lightning.blockchain.*
 import fr.acinq.lightning.channel.ChannelAction
 import fr.acinq.lightning.channel.ChannelCommand
+import fr.acinq.lightning.channel.ForcedLocalCommit
 import fr.acinq.lightning.channel.PleasePublishYourCommitment
 import fr.acinq.lightning.utils.Either
 import fr.acinq.lightning.wire.ChannelReady
@@ -107,7 +108,7 @@ data class Offline(val state: PersistedChannelState) : ChannelState() {
                 }
             }
             cmd is ChannelCommand.Close.ForceClose -> {
-                val (newState, actions) = state.run { process(cmd) }
+                val (newState, actions) = state.run { handleLocalError(cmd, ForcedLocalCommit(channelId)) }
                 when (newState) {
                     // NB: it doesn't make sense to try to send outgoing messages if we're offline.
                     is Closing -> Pair(newState, actions.filterNot { it is ChannelAction.Message.Send })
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/states/Offline.kt
```diff
@@ -99,15 +100,7 @@ data class Offline(val state: PersistedChannelState) : ChannelState() {
...
-            cmd is ChannelCommand.Close.ForceClose -> {
-                val (newState, actions) = state.run { process(cmd) }
-                when (newState) {
-                    // NB: it doesn't make sense to try to send outgoing messages if we're offline.
-                    is Closing -> Pair(newState, actions.filterNot { it is ChannelAction.Message.Send })
-                    is Closed -> Pair(newState, actions.filterNot { it is ChannelAction.Message.Send })
-                    else -> Pair(Offline(newState as PersistedChannelState), actions)
-                }
-            }
+            cmd is ChannelCommand.Close.ForceClose -> state.run { handleLocalError(cmd, ForcedLocalCommit(channelId)) }
```
t-bast, line 110: It's probably fine to that we're not filtering the `error` message anymore, since we only handle `ChannelAction.Message.Send` if `_connectionState.value == Connection.ESTABLISHED`. One thing that surprises me though is that I don't see what will happen to this action in the disconnected case: the `when` in `processActions` (in `Peer.kt`) doesn't seem to be exhaustive...which is surprising because the compiler should complain about it?

src/commonTest/kotlin/fr/acinq/lightning/channel/states/OfflineTestsCommon.kt
```diff
@@ -662,8 +662,11 @@ class OfflineTestsCommon : LightningTestSuite() {
         val commitTx = alice1.commitments.latest.localCommit.publishableTxs.commitTx.tx
         val (alice2, actions2) = alice1.process(ChannelCommand.Close.ForceClose)
         assertIs<Closing>(alice2.state)
-        actions2.hasPublishTx(commitTx)
-        assertNull(actions2.findOutgoingMessageOpt<Error>()) // we're offline so we shouldn't try to send messages
+        assertTrue(actions2.contains(ChannelAction.Storage.StoreState(alice2.state)))
+        assertEquals(commitTx, actions2.hasPublishTx(ChannelAction.Blockchain.PublishTx.Type.CommitTx))
```
t-bast, line 666: Isn't `actions2.hasPublishTx(commitTx)` better?

src/commonMain/kotlin/fr/acinq/lightning/channel/states/Offline.kt
pm47: This may be due to the use of `when` without clause. Switch to `when` with clause in 6d057dca and made sure that it complains when pattern matching is not exhaustive.

src/commonTest/kotlin/fr/acinq/lightning/channel/states/OfflineTestsCommon.kt
pm47: Done in f6e4cec3.

PR #476, opened 2023-06-08
Use exhaustive pattern matching

t-bast:
When using `when` without a clause and with a `Unit` return type, the compiler does not check for exhaustiveness, which could lead to issues.

:warning: this PR should be reviewed using the `Hide whitespace` github option, otherwise it will be hard to review the diff (which is actually not that big)

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
@@ -27,12 +27,12 @@ data class Normal(
     override fun updateCommitments(input: Commitments): ChannelStateWithCommitments = this.copy(commitments = input)
 
     override fun ChannelContext.processInternal(cmd: ChannelCommand): Pair<ChannelState, List<ChannelAction>> {
-        return when {
-            cmd is ChannelCommand.ForbiddenDuringSplice && spliceStatus !is SpliceStatus.None -> {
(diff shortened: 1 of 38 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
```diff
@@ -664,7 +664,11 @@ data class Normal(
                 sigStash = emptyList()
                 Pair(Offline(this@Normal.copy(spliceStatus = spliceStatus1)), failedHtlcs)
             }
-            else -> unhandled(cmd)
```
t-bast, line 667: I'm a bit afraid of this `else` in channel states, we're never sure what events we're actually ignoring and if they're safe to ignore. I only updated the `Normal` state, should I do the same thing in other states? We will end up with quite a few `is XXX -> unhandled(cmd)`, but at least it's explicit and we can then detect that a command is incorrectly ignored. This could be particularly useful when adding new `ChannelCommand`s, to force us to think about each channel state and whether it should handle that new command or not. WDYT?

pm47: It makes adding new commands more involved, but is safer, and that doesn't happen often anyway. Let's try it, we can always revisit if needed.

pm47 (approved):
Not sure why I dropped the 🏀 on this.

PR #478, opened 2023-06-08
Forget unsigned channels on errors

t-bast:
We were previously updating the state to `Aborted`, but we weren't updating the DB, which means that a restart of the app would show the channel again.


pm47 (approved):
LGTM.

This is safe because in `WaitForFundingSigned` we have not yet sent our sigs.

PR #482, opened 2023-06-13
Derive per-user swap-in server keys

t-bast:
Swap-in servers provide an xpub, that can then be used to derive the public key they'll use for each peer deterministically (based on every peer's `node_id`).

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
@@ -117,7 +117,6 @@ interface KeyManager {
         val userPublicKey: PublicKey = userPrivateKey.publicKey()
 
         private val localServerExtendedPrivateKey: DeterministicWallet.ExtendedPrivateKey = DeterministicWallet.derivePrivateKey(master, swapInKeyBasePath(chain) / hardened(1))
-        val localServerExtendedPublicKey: DeterministicWallet.ExtendedPublicKey = DeterministicWallet.publicKey(localServerExtendedPrivateKey)
         fun localServerPrivateKey(remoteNodeId: PublicKey): PrivateKey = DeterministicWallet.derivePrivateKey(localServerExtendedPrivateKey, perUserPath(remoteNodeId)).privateKey
 
         val redeemScript: List<ScriptElt> = Scripts.swapIn2of2(userPublicKey, remoteServerPublicKey, refundDelay)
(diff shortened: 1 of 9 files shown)


src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
```diff
@@ -114,8 +116,9 @@ interface KeyManager {
         val userPrivateKey: PrivateKey = DeterministicWallet.derivePrivateKey(master, swapInKeyBasePath(chain) / hardened(0)).privateKey
         val userPublicKey: PublicKey = userPrivateKey.publicKey()
 
-        val localServerPrivateKey: PrivateKey = DeterministicWallet.derivePrivateKey(master, swapInKeyBasePath(chain) / hardened(1)).privateKey
-        val localServerPublicKey: PublicKey = localServerPrivateKey.publicKey()
+        private val localServerExtendedPrivateKey: DeterministicWallet.ExtendedPrivateKey = DeterministicWallet.derivePrivateKey(master, swapInKeyBasePath(chain) / hardened(1))
+        val localServerExtendedPublicKey: DeterministicWallet.ExtendedPublicKey = DeterministicWallet.publicKey(localServerExtendedPrivateKey)
```
sstone, line 120: `localServerExtendedPublicKey` does not seem to be used ?

src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
```diff
@@ -44,7 +44,11 @@ data class LocalKeyManager(val seed: ByteVector, val chain: Chain, val remoteSwa
     )
 
     override val finalOnChainWallet: KeyManager.Bip84OnChainKeys = KeyManager.Bip84OnChainKeys(chain, master, account = 0)
-    override val swapInOnChainWallet: KeyManager.SwapInOnChainKeys = KeyManager.SwapInOnChainKeys(chain, master, remoteSwapInServerKey)
+    override val swapInOnChainWallet: KeyManager.SwapInOnChainKeys = run {
+        val xpub = DeterministicWallet.ExtendedPublicKey.decode(remoteSwapInExtendedPublicKey).second
```
sstone, line 48: We should check that the prefix returned by  `ExtendedPublicKey.decode` matches our chain

sstone (commented):
LGTM, just a few nits

src/commonMain/kotlin/fr/acinq/lightning/crypto/LocalKeyManager.kt
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/482/commits/f5c599ba9fc7342a939de0110676e5d1828e7681

t-bast:
I'm not sure why one of the seemingly unrelated tests is failing without any stack trace on linux (but not on the JVM), will need to investigate...

sstone (approved):

pm47 (approved):

PR #485, opened 2023-06-21
Remove `Peer.bootChannels` map

pm47:
Right after starting a `Peer`, the `channels` map is empty because it is loaded asynchronously in a coroutine. Library integrators need a way to differentiate between "channels not yet loaded" and "no channels at all", otherwise they may e.g. displat a 0 sat balance for a short amount of time.

Instead of using a nullable map `bootChannels`, we add an explicit _initialized_ boolean.


dpad85:
Note: this PR is paused as there are side-effects with electrum (see https://github.com/ACINQ/lightning-kmp/pull/483#discussion_r1236735367). Removing the explicit `bootChannels` flow prevents us from distinguishing:
- channels not loaded yet
- channels loaded from local database
- channels loaded and reestablished with the peer (<- depends on the Electrum)

PR #486, opened 2023-06-22
Add optional feerate for swap-in

pm47:
Library integrator is responsible for providing a value. If set, Phoenix will use it to iniate a splice (which only happens in the swap-in case). Otherwise we don't start the swap-in coroutine.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -98,7 +98,8 @@ class Peer(
     socketBuilder: TcpSocket.Builder?,
     scope: CoroutineScope,
     private val isMigrationFromLegacyApp: Boolean = false,
-    private val initTlvStream: TlvStream<InitTlv> = TlvStream.empty()
+    private val initTlvStream: TlvStream<InitTlv> = TlvStream.empty(),
+    private val initialSwapInFeerate: FeeratePerKw? = null
 ) : CoroutineScope by scope {
     companion object {
         private const val prefix: Byte = 0x00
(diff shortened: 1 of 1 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -157,6 +157,7 @@ class Peer(
 
     val currentTipFlow = MutableStateFlow<Pair<Int, BlockHeader>?>(null)
     val onChainFeeratesFlow = MutableStateFlow<OnChainFeerates?>(null)
+    val swapInFeeratesFlow = MutableStateFlow<FeeratePerKw?>(null)
```
t-bast, line 160: Since this is a flow that is written to only by the application using `lightning-kmp`, why not make it an argument of the `Peer` or put it in `WalletParams` as an optional state flow (ie `StateFlow<FeeratePerKw>?`)? This way in `lightning-kmp`, if that flow is provided, we wait for it before `watchSwapInWallet` and it cannot contain `null`, and if the flow isn't provided we default to using `onChainFeeratesFlow`? That shouldn't be much harder and is conceptually cleaner IMO.

pm47: If we make the feerate non-nullable, then wherever we put the `StateFlow` it will effectively prevent initializing the peer before we know the feerate. With a nullable value we can start up the peer immediately, and only defer the swap-in wallet, which is asynchronous anyway. So I think we need to keep it a `StateFlow<FeeratePerKw?>`.

My rationale for declaring the flow as an attribute of `Peer` is that it sits logically next to `onChainFeeratesFlow`, and I don't see a case where it would make sense to use the default electrum feerate. Library users who are not Phoenix will not provide a feerate, and it will effectively not start the swap-in coroutine, it's pretty nice?

t-bast (approved):

PR #489, opened 2023-06-23
Retrieve feerates everytime we reconnect to electrum

pm47:
Before this change, we were only retrieving the feerates once during the lifetime of the app. If we happen to connect to an unhealthy Electrum server (connection succeeds, but then we get `ServerError(request=EstimateFees(confirmations=18), error=JsonRPCError(code=-102, message=server busy - request timed out))`), then the app will fail to start and actually crash.

This change is more subtle than it may look. Not only do we catch and ignore errors, but we also retrieve feerates at each reconnection. The goal is that in the scenario above we would try another electrum server and get back on our feet.

⚠️ It's very hard to test and I'm not sure of the side effects.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -196,8 +196,8 @@ class Peer(
         }
         launch {
             watcher.client.connectionState.filter { it == Connection.ESTABLISHED }.collect {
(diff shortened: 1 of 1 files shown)


sstone:
Since we want to retrieve onchain fee rate each time we connect to an electrum server, and that some may fail to return a valid response, another approach would be to add retrieving fee rates to the handshake we perform when we establish a new connection, see https://github.com/ACINQ/lightning-kmp/pull/490 ?

PR #490, opened 2023-06-27
Retrieve onchain feerate when we connect to an electrum server

sstone:
When we connect to an electrum sever we perform a "handshake" that includes exchanging protocol version messages and retrieving the server's current tip, and now we also retrieve onchain fees.


sstone:
I don't think so since we still ask the electrum server we're connected to for fee estimates and it may return an error ? The point here was do we add this to the handshake we perform when we connect to the server or do we try a different fix.

PR #492, opened 2023-06-28
Create invoice outside of the peer's message loop

pm47:
Invoices creation is independent from the peer, and it would remove the artificial dependency between creating an invoice and connecting to electrum.

This has impacts on Phoenix since the `ReceivePayment` message is removed and replaced by a suspend `createInvoice()` function. I think it's easier but let me know if it's a problem @dpad85 @robbiehanson.


t-bast (approved):

dpad85 (approved):
Successfully tested on Android.

PR #493, opened 2023-06-28
Make blockheight/fees flows non-nullable

pm47:
This guarantees that the peer will never block before processing a channel message, but implies that:
- updating these values is done outside of the peer (note: instantiating the electrum client is already done outside)
- the peer cannot be instantiated before a connection to electrum has been established

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -97,6 +97,9 @@ class Peer(
     val db: Databases,
     socketBuilder: TcpSocket.Builder?,
     scope: CoroutineScope,
+    val currentTipFlow: StateFlow<Pair<Int, BlockHeader>>,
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -188,20 +187,6 @@ class Peer(
     val swapInAddress: String = nodeParams.keyManager.swapInOnChainWallet.address.also { swapInWallet.addAddress(it) }
 
     init {
-        launch {
-            watcher.client.notifications.filterIsInstance<HeaderSubscriptionResponse>()
```
pm47, line 192: There was a race condition here between establishing the connection to electrum (which happens outside of the peer), and subscribing to `HeaderSubscriptionResponse`.

If the connection to electrum is very fast, tje peer will miss the initial `HeaderSubscriptionResponse` and then have to wait for the next block.

Same race condition applies for feerates.

PR #497, opened 2023-06-30
Provide a set of trusted txids for swap-in

pm47:
Instead of a `isMigrationFromLegacyApp` boolean, we provide a set of txids that can be considered safe for swap-in even if they are not confirmed.

This is not more complicated than relying on a boolean to implement zeroconf migration, and much more robust.


pm47:
Rebased.

dpad85 (approved):
Tested successfully on Phoenix.

t-bast (approved):

PR #509, opened 2023-07-10
Fix `skipAbsoluteFeeCheck` in liquidity policy

pm47:
There were two separate issues:
- the bypass wasn't working at all because we were using a max of 0.msat
- the bypass should only apply to off-chain payments


src/commonMain/kotlin/fr/acinq/lightning/payment/LiquidityPolicy.kt
```diff
@@ -25,7 +25,7 @@ sealed class LiquidityPolicy {
         return when (this) {
             is Disable -> LiquidityEvents.Rejected.Reason.PolicySetToDisabled
             is Auto -> {
-                val maxAbsoluteFee = if (skipAbsoluteFeeCheck) 0.msat else this.maxAbsoluteFee.toMilliSatoshi()
+                val maxAbsoluteFee = if (skipAbsoluteFeeCheck && source == LiquidityEvents.Source.OffChainPayment) Long.MAX_VALUE.msat else this.maxAbsoluteFee.toMilliSatoshi()
```
t-bast, line 28: How did we miss that one :man_facepalming:

t-bast (approved):

PR #513, opened 2023-07-19
Get transaction's confirmation information (including block timestamp)

robbiehanson:
The `OnChainOutgoingPayment` has a `confirmedAt` timestamp:

```kotlin
sealed class OnChainOutgoingPayment : OutgoingPayment() {
    abstract override val id: UUID
    abstract val miningFees: Satoshi
    abstract val channelId: ByteVector32
    abstract val txId: ByteVector32
    abstract override val createdAt: Long
    abstract val confirmedAt: Long?
    abstract val lockedAt: Long?
}
```

Currently this is being set (in Phoenix) by monitoring the blockchain for unconfirmed tx's, and setting `confirmedAt = currentTimestampMillis()`. However, the timestamp isn't ideal.

It's usually the case that a user sends an on-chain payment, and then backgrounds the app. (Which means the socket connections are closed on iOS.) Only when the user returns to the app (possibly days later) do we realize that the payment was mined, and we set `confirmedAt = currentTimestampMillis()`. Which, in this example, is days later than when the tx was actually mined.

I'm proposing we add a method like this:

```kotlin
data class ConfirmationStatus(
    val minedAtBlockHeight: Int,
    val currentConfirmationCount: Int,
    val firstBlock: BlockHeader // includes timestamp of when block was mined
)

suspend fun IElectrumClient.getConfirmationStatus(txId: ByteVector32): ConfirmationStatus?
```

Using this function, we'll be able to set the `confirmedAt` property to the timestamp of the block, which would be more accurate.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClientExtensions.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClientExtensions.kt
@@ -1,5 +1,6 @@
 package fr.acinq.lightning.blockchain.electrum
 
+import fr.acinq.bitcoin.BlockHeader
 import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.bitcoin.Satoshi
 import fr.acinq.bitcoin.Transaction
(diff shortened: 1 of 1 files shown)


src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumClientExtensions.kt
```diff
@@ -9,7 +10,48 @@ import fr.acinq.lightning.channel.LocalFundingStatus
 import fr.acinq.lightning.transactions.Transactions
 import fr.acinq.lightning.utils.MDCLogger
 import fr.acinq.lightning.utils.sat
+import kotlinx.coroutines.CompletableDeferred
 
+data class ConfirmationStatus(
+    val minedAtBlockHeight: Int,
+    val currentConfirmationCount: Int,
+    val firstBlock: BlockHeader // includes timestamp of when block was mined
```
t-bast, line 18: nit:
```suggestion
    val includedInBlock: BlockHeader // includes timestamp of when block was mined
```

t-bast (commented):
Concept ACK, but this will be cleaner to do on top of #512, which offers a direct `getHeader()` function on the `IElectrumClient` interface. After #512, we could also remove some duplication by having `getConfirmationStatus` internally call `getConfirmations` and simply follow up with `getHeader()`.

My goal is to finalize #512 soon, so we should probably wait for this?

t-bast:
@robbiehanson you can probably rebase that and do it more easily now!

PR #515, opened 2023-08-14
Check swap-in wallet at each block

pm47:
The `WalletState` class contains absolute block heights for utxos, and does not contain the current block height, so it is constant as soon as an utxo confirms. Even if `ElectrumMiniWallet` explicitly updated  `_walletStateFlow` at each new block (which it doesn't), it would be a no-op because `MutableStateFlow` deduplicates events.

We need to look for swap-ins when _either_ the wallet state or the current block height changes, otherwise we will only call `TrySwapIn` when the utxo reaches `weak` confirmation, and not when it graduates from `weak` to `deep` confirmation.


t-bast (approved):
Good catch!

PR #516, opened 2023-08-17
Make `Peer.trustedSwapInTxs` a callback

pm47:
It allows adding txids without restarting the Peer (useful for iOS channel migration/consolidation) and facilitates testing.

PR #517, opened 2023-08-18
Trust local mutual close for swap-in

pm47:
Instead of explicitly providing txids, we can look at local channel data to extract mutual close txids, because either the tx is unconfirmed and the channel data will still be in `Closing` state, or the tx is confirmed and the utxo can be used immediately anyway.

This is an alternative to #516.


t-bast (approved):
Since the min-confirmations for swaps matches the confirmations we put in `WatchConfirmed`, this should work :+1:

pm47:
Rebased.

PR #518, opened 2023-08-23
Add methods to start/stop swap-in wallet

pm47:
For the migration use case, it is useful to be able to wait for a certain number of UTXOs to be published before doing a swap-in. The reason is that:
- we want to make as few on-chain operations as possible to save fees, so it's better to group
- migration swap-in have special characteristics that are only implemented for channel creation, not for splices.

The new methods `startWatchSwapInWallet`/`stopWatchSwapInWallet` are footgunny and should only be used in the context of the migration, to be removed later. Note that the `Peer` calls `startWatchSwapInWallet` at initialization, so the migration process must first `stop...()`, then later `start...()`.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -185,7 +185,7 @@ class Peer(
     val swapInWallet = ElectrumMiniWallet(nodeParams.chainHash, watcher.client, scope, nodeParams.loggerFactory, name = "swap-in")
     val swapInAddress: String = nodeParams.keyManager.swapInOnChainWallet.address.also { swapInWallet.addAddress(it) }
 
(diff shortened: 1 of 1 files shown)


pm47:
With the latest commit, the swap-in watcher isn't automatically started by the `Peer` at initialization.

The goal is to be able to handle the case where something went wrong during the migration, and only a subset of the legacy channels have been closed:
- before this change, at the next restart the swap-in watcher would automatically process the closed channels and we would end up in an half-migrated state
- after this change, at the next restart the app can check if all legacy channels have been closed and then:
  - if there are no legacy channels, enable the swap-in watcher and start the app normally
  - if there are some legacy channels, do the migration again, and only then enable the swap-in watcher

@dpad85 this affects the Android app, `peer.startWatchSwapInWallet()` must be called after initialization.

dpad85 (approved):

PR #526, opened 2023-09-11
Fix race condition in connection to Peer

pm47:
The fix is in two parts:
- First we clean up the connection logic, introduce connection ids and make the main peer loop responsible for switching between connections (see [bafb425](https://github.com/ACINQ/lightning-kmp/pull/526/commits/bafb4253e243d6762c3a04e7b638a318749015d1))
- Then we fix `processActions` to make sure that we always use the connection that the `Channel` FSM was using when it asked to send messages (see [61446b4](https://github.com/ACINQ/lightning-kmp/pull/526/commits/61446b46cb2aa9d4c94a8ae5ec1292764b5df611))

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -342,7 +342,7 @@ class Peer(
             val ping = Ping(10, ByteVector("deadbeef"))
             while (isActive) {
                 delay(30.seconds)
-                sendToPeer(ping)
+                peerConnection.output.trySend(ping)
             }
         }
 
@@ -532,17 +532,16 @@ class Peer(
         return incomingPaymentHandler.createInvoice(paymentPreimage, amount, description, extraHops, expirySeconds)
     }
 
-    @OptIn(DelicateCoroutinesApi::class)
-    suspend fun sendToPeer(msg: LightningMessage) {
-        peerConnection?.output?.run { if (!isClosedForSend) send(msg) }
+    private fun sendToPeer(msg: LightningMessage) {
+        // We can safely use trySend because we use unlimited channel buffers.
+        peerConnection?.output?.run { trySend(msg) }
     }
 
     // The (node_id, fcm_token) tuple only needs to be registered once.
     // And after that, only if the tuple changes (e.g. different fcm_token).
     fun registerFcmToken(token: String?) {
-        launch {
-            sendToPeer(if (token == null) UnsetFCMToken else FCMToken(token))
-        }
+        val message = if (token == null) UnsetFCMToken else FCMToken(token)
+        sendToPeer(message)
     }
 
     private suspend fun processActions(channelId: ByteVector32, actions: List<ChannelAction>) {
@@ -743,7 +742,7 @@ class Peer(
     }
 
     // MUST ONLY BE SET BY processEvent()
-    var peerConnection: PeerConnection? = null
+    private var peerConnection: PeerConnection? = null
 
     private suspend fun processEvent(cmd: PeerCommand) {
         when (cmd) {
@@ -754,7 +753,7 @@ class Peer(
             }
             is MessageReceived -> {
                 if (cmd.connectionId != peerConnection?.id) {
-                    logger.warning { "ignoring ${cmd.msg} for connectionId=${cmd.connectionId}"}
+                    logger.warning { "ignoring ${cmd.msg} for connectionId=${cmd.connectionId}" }
                     return
                 }
                 val msg = cmd.msg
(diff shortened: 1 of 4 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -281,7 +279,7 @@ class Peer(
     fun disconnect() {
         if (this::socket.isInitialized) socket.close()
         _connectionState.value = Connection.CLOSED(null)
-        output.close()
+        // TODO in exception handler? output.close()
```
pm47, line 282: I'm not sure where to do this, I believe we should probably copy the clean up code from ElectrumClient?

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -334,17 +332,10 @@ class Peer(
...
-                session.send(message) { data, flush -> socket.send(data, flush) }
-            } catch (ex: TcpSocket.IOException) {
-                logger.warning { "TCP send: ${ex.message}" }
-                closeSocket(ex)
-            }
-        }
-
-        logger.info { "sending init $ourInit" }
-        send(LightningMessage.encode(ourInit))
+        // TODO use atomic counter instead
```
pm47, line 335: There doesn't seem to be a portable implementation `AtomicInt` unfortunately. I can fallback to `UUID`...

t-bast: This seems to be the library for multiplatform atomic operations, but I'm not sure if it's worth importing right now: https://github.com/Kotlin/kotlinx-atomicfu

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -49,7 +47,15 @@ data class OpenChannel(
     val channelType: ChannelType.SupportedChannelType
 ) : PeerCommand()
 
-data class BytesReceived(val data: ByteArray) : PeerCommand()
+data class PeerConnection(val id: Long, val output: Channel<LightningMessage>) {
+    fun send(msg: LightningMessage) {
+        // We can safely use trySend because we use unlimited channel buffers.
+        // If the connection was closed, the message will automatically be dropped.
+        output.run { trySend(msg) }
```
t-bast, line 54: nit: why the indirection through `run`, it's less readable that way?

```suggestion
        output.trySend(msg)
```

We could also look at the result and log something if the message was dropped, that could be useful when debugging? There are cases where we won't know that a message is dropped (if the TCP connection has been closed but the `output` channel hasn't been closed yet), but it should still be useful.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -360,30 +362,47 @@ class Peer(
...
+        suspend fun receiveLoop() {
             try {
                 while (isActive) {
                     val received = session.receive { size -> socket.receiveFully(size) }
-                    input.send(BytesReceived(received))
+                    try {
+                        val msg = LightningMessage.decode(received)
+                        input.send(MessageReceived(peerConnection.id, msg))
+                    } catch (e: Throwable) {
+                        logger.warning { "cannot deserialized message: ${received.byteVector().toHex()}" }
```
t-bast, line 373: nit:
```suggestion
                        logger.warning { "cannot deserialize message: ${received.byteVector().toHex()}" }
```

t-bast (commented):
I think your analysis on the `ChannelCommand` is correct, which means we should be fine without having to make `processEvent` fully synchronous. The most confusing command is `SendToSelf` which lets readers think there may be unbounded recursion, we should probably simply replace it by a `SignCommit` command instead since it's our only use-case (but probably in a separate PR)?

Let's see if that fixes the issue completely, otherwise we'll have more logs to help us troubleshoot.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
pm47: Yeah that seems a bit risky. I think we can stick with milliseconds. Maybe `UUID` to be safe?

t-bast: I'm not sure what is best between UUID and timestamp...we shouldn't have two connections established at the same millisecond, so timestamp is probably fine?

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
pm47: Done in fc4d5fae, with a nice dedicated MDC logger.

t-bast (approved):

PR #527, opened 2023-09-11
Add more improvements to PeerConnection

t-bast:
Builds on top of #526 to fully fix the race conditions in peer connections (:crossed_fingers:).

The most important and subtle commits are https://github.com/ACINQ/lightning-kmp/pull/527/commits/c75939296d56dfee65593ba84983af923b2e5298 and https://github.com/ACINQ/lightning-kmp/pull/527/commits/9eb5c43c8c4006696c9ea39a9ce4b13d73838c86

I wanted to refactor that code to make sure that event processing could not access the `input` channel, and provide an overall better encapsulation, but this was quite a rabbit hole so I chose to keep the number of changes minimal for this PR. A follow-up PR can work on pure refactoring to split the `Peer` into smaller classes.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -343,7 +343,7 @@ class Peer(
             val ping = Ping(10, ByteVector("deadbeef"))
             while (isActive) {
                 delay(30.seconds)
-                sendToPeer(ping)
+                peerConnection.output.trySend(ping)
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/commonTest/kotlin/fr/acinq/lightning/io/peer/PeerTest.kt
```diff
@@ -62,9 +62,9 @@ class PeerTest : LightningTestSuite() {
         val bob = buildPeer(this, TestConstants.Bob.nodeParams, TestConstants.Bob.walletParams)
 
         // start Init for Alice
-        alice.send(MessageReceived(Init(features = TestConstants.Bob.nodeParams.features)))
+        alice.send(MessageReceived(Init(features = TestConstants.Bob.nodeParams.features), connectionId = 0))
```
pm47, line 65: I intended to remove the default too, but also switch the order of args to put `connectionId` first.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -272,24 +271,30 @@ class Peer(
...
     }
 
     // Warning : lateinit vars have to be used AFTER their init to avoid any crashes
     //
-    // This shouldn't be used outside the establishedConnection() function
+    // This shouldn't be used outside the establishConnection() function
     // Except from the disconnect() one that check if the lateinit var has been initialized
     private lateinit var socket: TcpSocket
     private fun establishConnection() = launch {
+        // Clean up previous connection state: we do this here to ensure that it is handled before the Connected event for the new connection.
```
pm47, line 293: Don't you feel like replicating the `ListenJob` pattern of `ElectrumClient`? And the `CoroutineExceptionHandler` snippet? If only for symmetry? Why would it be appropriate for electrum but not for the peer?

PR #528, opened 2023-09-12
Don't send splice_locked before tx_signatures

t-bast:
When reconnecting in the middle of signing a splice, we must ensure that splice_locked is sent *after* tx_signatures. Otherwise when using 0-conf we may retransmit splice_locked before tx_signatures, which our peer will ignore because they don't have a corresponding fully signed commitment.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Syncing.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Syncing.kt
@@ -155,26 +155,13 @@ data class Syncing(val state: PersistedChannelState, val channelReestablishSent:
                                     // normal case, our data is up-to-date
                                     val actions = ArrayList<ChannelAction>()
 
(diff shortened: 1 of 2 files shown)


pm47 (approved):
Great catch!

Just to be sure, this happens if we get disconnected right after having gone over this line:
https://github.com/ACINQ/lightning-kmp/blob/fdba64a9d83b0da78c0ba413888223e1dc2e543e/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt#L729

t-bast:
> Just to be sure, this happens if we get disconnected right after having gone over this line:

Yes, that's exactly it! In that case on reconnection we will re-send `[commit_sig, tx_signatures, splice_locked]`. But we were actually sending `[splice_locked, commit_sig, tx_signatures]` so the `splice_locked` was ignored by the other side, which didn't yet have a matching commitment.

PR #529, opened 2023-09-14
Handle 'unknown segwit' case when checking amounts

pm47:
As per https://github.com/lightning/bolts/blob/master/03-transactions.md#dust-limits


t-bast (approved):
Damn, good catch!

PR #533, opened 2023-09-20
Make swap-in wait for peer ready

pm47:
There is a race at startup between the swap-in and the peer connection. If the former wins the race, we reach the following code and drop the message, causing the swap-in to be stuck.
https://github.com/ACINQ/lightning-kmp/blob/d2bc164b52fd2070af3514fa86e512ed981fb492/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt#L1026-L1027

This PR introduces a `waitForPeerReady` helper method which blocks until the peer is fully connected (meaning that existing channels are synced).

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -468,7 +468,7 @@ class Peer(
             channelsFlow.first { it.values.all { channel -> channel !is Offline && channel !is Syncing } }
         }
         if (result == null) {
-            logger.info { "timeout elapsed, not all channels are synced but proceeding anyway" }
+            logger.info { "peer ready timeout elapsed, not all channels are synced but proceeding anyway" }
         }
     }
 
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -420,7 +420,10 @@ class Peer(
         logger.info { "starting swap-in watch job" }
         if (swapInJob != null) return
         // wait to have a swap-in feerate available
+        logger.info { "waiting for feerates" }
         swapInFeeratesFlow.filterNotNull().first()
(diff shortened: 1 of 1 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -420,7 +420,10 @@ class Peer(
         logger.info { "starting swap-in watch job" }
         if (swapInJob != null) return
         // wait to have a swap-in feerate available
+        logger.info { "waiting for feerates" }
         swapInFeeratesFlow.filterNotNull().first()
+        logger.info { "waiting for peer to be ready" }
+        waitForPeerReady()
```
pm47, line 426: Moving this to the `processSwapInCommands` function, right before sending the `RequestChannelOpen`, would probably be more correct, but here it is simpler conceptually and consistent with waiting for feerates.

The rationale is to not try to reach a perfect solution, but a simple one which at worst guarantee that things fix themselves by simply restarting the app.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -452,6 +455,23 @@ class Peer(
...
 
+    /**
+     * This function blocks until the peer is connected and existing channels have been fully reestablished.
+     */
+    private suspend fun waitForPeerReady() {
+        // In theory we would only need to verify that no channel is in state Offline/Syncing, but there is a corner
+        // case where a channel permanently stays in Syncing, because it is only present locally, and the peer will
+        // never send a channel_reestablish (this happens e.g. due to an error at funding). That is why we consider
+        // the peer ready if "all channels are synced" OR "peer has been connected for 10s".
+        connectionState.first { it is Connection.ESTABLISHED }
```
pm47, line 466: We lack a metadata from the peer allowing us to tell well our channels are fully synced. In the meantime, this workaround should work.

t-bast (dismissed):
It seems to be a good enough workaround :+1:

t-bast (approved):

PR #536, opened 2023-09-22
Fix `invalid_commit_sig` issue when expecting `tx_signatures`

t-bast:
If we disconnect before receiving our peer's `tx_signatures`, we expect them to re-send it on reconnection (by filling the `next_funding_txid` field in `channel_reestablish`). In that case, our peer may also retransmit their corresponding `commit_sig` first (if they cannot be sure that we've received it). If we've already received that `commit_sig`, we should ignore it. But we must make sure we don't end up ignoring *other* `commit_sig` messages!

Eclair wasn't spec-compliant and wouldn't retransmit their `tx_signatures` if the transaction was already confirmed. We need to fix this on the eclair side (PR incoming). The issue is that `lightning-kmp` ended up ignoring unrelated `commit_sig` messages, which could eventually lead to an `invalid_commit_sig` error and a force-close.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
@@ -123,7 +123,7 @@ sealed class LocalFundingStatus {
         override val fee: Satoshi = sharedTx.tx.fees
     }
 
-    data class ConfirmedFundingTx(override val signedTx: Transaction, override val fee: Satoshi) : LocalFundingStatus() {
+    data class ConfirmedFundingTx(override val signedTx: Transaction, override val fee: Satoshi, val localSigs: TxSignatures) : LocalFundingStatus() {
         override val txId: ByteVector32 = signedTx.txid
     }
 }
@@ -832,7 +832,7 @@ data class Commitments(
                 when (c.localFundingStatus) {
                     is LocalFundingStatus.UnconfirmedFundingTx -> {
                         logger.debug { "setting localFundingStatus confirmed for fundingTxId=${fundingTx.txid}" }
-                        c.copy(localFundingStatus = LocalFundingStatus.ConfirmedFundingTx(fundingTx, c.localFundingStatus.sharedTx.tx.fees))
+                        c.copy(localFundingStatus = LocalFundingStatus.ConfirmedFundingTx(fundingTx, c.localFundingStatus.sharedTx.tx.fees, c.localFundingStatus.sharedTx.localSigs))
                     }
                     is LocalFundingStatus.ConfirmedFundingTx -> c
                 }
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
@@ -194,7 +194,7 @@ data class Normal(
                                 is InteractiveTxSigningSessionAction.SendTxSigs -> sendSpliceTxSigs(spliceStatus.origins, action, cmd.message.channelData)
                             }
                         }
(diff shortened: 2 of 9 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -123,7 +123,7 @@ sealed class LocalFundingStatus {
         override val fee: Satoshi = sharedTx.tx.fees
     }
 
-    data class ConfirmedFundingTx(override val signedTx: Transaction, override val fee: Satoshi) : LocalFundingStatus() {
+    data class ConfirmedFundingTx(override val signedTx: Transaction, override val fee: Satoshi, val localSigs: TxSignatures) : LocalFundingStatus() {
```
t-bast, line 126: It's a bit annoying that we have to store this here, but:

- it's annoying to recompute (because we could have wallet inputs)
- it's relatively small (a few hundreds of bytes at most)
- it makes the code trivially spec-compliant (always re-send `tx_signatures` if requested, regardless of confirmation state)

pm47: Why don't we extract our sigs from `signedTx`?

pm47: For the 2-of-2 it's easy, we just compare `local/remoteFundingPubkey`. However for our other inputs, I don't know...

t-bast: I think it's not worth the additional complexity to be honest, I don't see how we could simply and reliably figure out which inputs are ours.

pm47: I agree, let's drop the idea.

src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
```diff
@@ -192,11 +194,10 @@ data class Normal(
                                 is InteractiveTxSigningSessionAction.SendTxSigs -> sendSpliceTxSigs(spliceStatus.origins, action, cmd.message.channelData)
                             }
                         }
-                        commitments.params.channelFeatures.hasFeature(Feature.DualFunding) && commitments.latest.localFundingStatus.signedTx == null && cmd.message.batchSize == 1 -> {
-                            // The latest funding transaction is unconfirmed and we're missing our peer's tx_signatures: any commit_sig that we receive before that should be ignored,
-                            // it's either a retransmission of a commit_sig we've already received or a bug that will eventually lead to a force-close anyway.
-                            // NB: we check the dual-funding feature, because legacy single-funding unconfirmed channels will have `localFundingStatus=UnconfirmedFundingTx`,
-                            // so we cannot just check on the absence of `signedTx`.
+                        commitments.params.channelFeatures.hasFeature(Feature.DualFunding) && ignoreRetransmittedCommitSig(cmd.message) -> {
```
pm47, line 197: If you call the method `ignoreRetransmittedCommitSig`, then it means that we are in a dual-funding context and we should move the feature check to that method imo, makes things lighter on the calling site:

```suggestion
                        ignoreRetransmittedCommitSig(cmd.message) -> {
```

pm47 (dismissed):
Nice, that was a subtle 🐛 .

src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
t-bast: Agreed, done in https://github.com/ACINQ/lightning-kmp/pull/536/commits/d515fc598801e065728d4a2212db885ee2e23336

pm47 (approved):

PR #538, opened 2023-09-22
Include `paymentHash` in `LiquidityEvent.Source`

pm47:
This allows integrators to easily map events to payments.


dpad85:
Looks good. However this will require work on Phoenix side to account for the `Source` type change in the database.

t-bast:
I'm not sure this will be necessary once we have #649 where the `payment_hash` can be found in the liquidity purchase structure.

PR #540, opened 2023-09-26
Exposed swap-in utxos refund delay

### LINKED ISSUE #534: Expose deeply confirmed utxos grace period in `WalletWithConfirmations`
Phoenix needs to know the grace period for every confirmed utxos in the swap-in wallet's `WalletWithConfirmations`, in order to display accurate information to the user.

I think the code that computes whether these utxos are eligible to swap-in or not should be done in `lightning-kmp`, not in Phoenix. 

Ideally from Phoenix we would be able to use an API like this:
1. just like [we do with `swapInConfirmations`](https://github.com/ACINQ/phoenix/blob/88fddeb12041bf483baf4503fd384a1e0d993fbd/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/PeerManager.kt#L118), Phoenix could inject into `WalletParams` : 
  - a new `swapInConfirmationsMax` parameter: confirmations count after which an utxo is not eligible to swap-in.
  - a new `swapInConfirmationsExpiry` parameter:  confirmations count after which an utxo is released from the swap-in potentiam scheme
2. we add two new parameters in `WalletWithConfirmations`
  - `locked`: list of utxos with `confirmation > swapInConfirmationsMax`
  - `released`: list of utxos with `confirmation > swapInConfirmationsExpiry`

The naming is just an example.

Regarding 1., just like the `NodeParams` are now defined in `lightning-kmp`, these swap-in constants should also be defined in `lightning-kmp` and not in Phoenix. It would make sense as these values are static and tightly coupled with the remote peer's settings.

t-bast:
Utxos used for swap-in have an expiry: after a delay the user can spend them unilaterally, which makes them dangerous to accept by the swap provider when we get close to that expiry. We expose these limits so that applications can tell their users to finalize swaps before reaching the expiry, otherwise they'll have to wait before unlocking them.

Fixes #534

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
@@ -36,16 +36,37 @@ data class TrampolineFees(val feeBase: Satoshi, val feeProportional: Long, val c
 data class InvoiceDefaultRoutingFees(val feeBase: MilliSatoshi, val feeProportional: Long, val cltvExpiryDelta: CltvExpiryDelta)
 
 /**
+ * These parameters must match the parameters used by the wallet provider.
+ *
+ * @param minConfirmations number of confirmations needed on swap-in transactions, before importing those funds into a channel.
+ * @param maxConfirmations maximum number of confirmations for swap-in transactions: funds need to be imported into a channel before reaching that threshold.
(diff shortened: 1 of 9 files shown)


dpad85 (approved):
LGTM 🙏 
Made a prototype on Android and we have all the info needed.

PR #541, opened 2023-09-27
Refactor peer connection establishment

### LINKED ISSUE #531: Connection issues analysis
I've been able to reproduce a connection issue on my mainnet Android device where
- The device suddenly gets very bad Internet connectivity. All apps are affected, including Phoenix.
- connection to the `Peer` is lost
- connection to `Electrum` is not. We still receive the new blocks (and even receive them right on time, I was checking in real time), however pings do time-out.

I've edit the logs so they are very easy to read:
https://bin.idrix.fr/?365fbd76444eadc4#wcppPRpmo4YaVRze82nU4bCuYURoPeqvo6scuxbccSo

There are several issues:
1. The Peer connection should not stay in `ESTABLISHING` state so long. Note that when that the internet connection is fixed, the peer only gets disconnected 20s later (probably by the remote peer) and only then we try to reconnect. This makes for a frustrating UX.
2. Electrum connection soldiers on and remains `ESTABLISHED`, even when pings fail. Pings have no effect.
3. The two points above means that the user will think that there's an issue with the ACINQ node.

Another issue that will have to be fixed in the Phoenix repo:

4. The exponential backoff in the `AppConnectionDaemon` connection loop is not working as expected (the pause should be reset at some point, instead of incrementing mindlessly: in the end when all the issues above were fixed, this loop still waited for 2s which was not useful).

t-bast:
Use the same pattern as the `ElectrumClient`, with an explicit job and timeouts. The `connect` function now suspends, which makes it easier to handle from the caller's point of view.

We don't have an easy way of unit testing that in `lightning-kmp`, so this will deserve some e2e tests.

Fixes #531

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -25,6 +25,7 @@ import kotlinx.coroutines.channels.Channel.Factory.UNLIMITED
 import kotlinx.coroutines.channels.onFailure
 import kotlinx.coroutines.flow.*
 import org.kodein.log.newLogger
+import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
 sealed class PeerCommand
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -275,137 +276,159 @@ class Peer(
...
+            // Clean up previous connection state: we do this here to ensure that it is handled before the Connected event for the new connection.
+            // That means we're not sending this event if we don't reconnect. It's ok, since that has the same effect as not detecting a disconnection and closing the app.
+            input.send(Disconnected)
             _connectionState.value = Connection.ESTABLISHING
-            establishConnection()
+
+            val connectionId = currentTimestampMillis()
+            val logger = MDCLogger(nodeParams.loggerFactory.newLogger(this::class), staticMdc = mapOf("remoteNodeId" to remoteNodeId, "connectionId" to connectionId))
+            logger.info { "connecting to ${walletParams.trampolineNode.host}" }
+            val socket = openSocket(connectTimeout) ?: return false
```
dpad85, line 298: If `openSocket` returns null and `connectionState` is NOT set to `CLOSED`, the `connect` method would return and the connection state would remain `CONNECTING`, i.e. we'd be deadlocked.

Looking at `openSocket` it seems that the only scenario where it can return null will be `caught` first -- and that catch does update the connection state to `CLOSED`. So we are probably OK and the deadlock scenario ^ is not possible. Can you confirm?

dpad85 (commented):
LGTM, just a nit

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
t-bast: Yes indeed: since `socketBuilder.connect` always returns a non-null `TcpSocket`, the only case where we can end up with `null` is if we went through the `catch` handler that resets the `_connectionState` to `CLOSED`, so we should be fine!

dpad85 (approved):

PR #542, opened 2023-09-27
Strip witnesses of interactive tx parents

pm47:
We can safely remove the witness data, it doesn't affect txid and saves space. There is a max size on transactions due to lightning message limits.


t-bast (approved):
LGTM! We should add the same improvement to `eclair` :+1:

PR #548, opened 2023-10-02
Try swap-in on liquidity policy changes

t-bast:
We previously tried swapping funds in:

- at start-up
- when our wallet state was updated
- when a new block was found

We now also check for swaps when the liquidity policy changes or the swap feerate changes.

Since this all happens inside the `Peer`, we don't have good ways of unit testing it. Should we start refactoring the `Peer` to allow more unit testing? This will be a bit painful because of all the internal channels and flows, but may be beneficial?

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -771,6 +771,7 @@ class Peer(
     // MUST ONLY BE SET BY processEvent()
     private var peerConnection: PeerConnection? = null
 
+    @OptIn(ExperimentalCoroutinesApi::class)
     private suspend fun processEvent(cmd: PeerCommand, logger: MDCLogger) {
         when (cmd) {
             is Connected -> {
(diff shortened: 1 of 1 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -1008,6 +1011,11 @@ class Peer(
                         val (feerate, fee) = watcher.client.computeSpliceCpfpFeerate(channel.commitments, targetFeerate, spliceWeight = weight, logger)
 
                         logger.info { "requesting splice-in using balance=${cmd.walletInputs.balance} feerate=$feerate fee=$fee" }
+                        if (channel.spliceStatus != SpliceStatus.None) {
+                            logger.info { "deferring splice, another splice is already in progress" }
+                            swapInCommands.send(SwapInCommand.UnlockWalletInputs(cmd.walletInputs.map { it.outPoint }.toSet()))
+                            return
+                        }
```
pm47, line 1018: This is a bit parallel to the main topic of the PR, but definitely needed too. I've seen conflicts between `UpdateFee` (from peer) and `SpliceInit` happen in the wild. In that case we stay stuck.

However I think we should do it at a different place if we want to really handle races properly, and instead define a `replyTo` here, that unlocks the inputs: https://github.com/ACINQ/lightning-kmp/blob/88b436644b731d194d7228c78a374d662b9c1557/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt#L1019-L1024

t-bast: Good idea, it's much better to handle all failure cases with the `replyTo` field :+1: 
Done in https://github.com/ACINQ/lightning-kmp/pull/548/commits/9d6fdce2301514c022753466dbfa9106c972e34f

PR #550, opened 2023-10-09
Add a keep-alive mechanism

pm47:
The goal is to make mobile apps (especially on iOS) more robust wrt to disconnections/app switching/background, etc.  by signalling when they need a little bit more uptime to complete some tasks.

~~This is just a proof of concept. In particular:
(a) The callback could take an argument providing more context on the task, for logging purposes. It could also be a `Flow` which maybe would help manage concurrency by integrators. There is no way to indicate when a given task has been completed, the call is stateless and just means "don't kill me just yet". I'm afraid it would be very difficult/impossible to have that level of precision in the general case.
(b) There is only one call to the callback and the location is probably not optimal, it's just an example.~~

~~Pay-to-open is a prime candidate for experimenting, because a disconnection _after_ the parts have been received by Phoenix and _before_ the peer has received the preimage would cause the payment to be aborted by the peer, and stay pending forever on phoenix.~~

@robbiehanson Please advise on a). Is there any limitation or best practice that we should be aware of (should we make frequent requests? or as few as possible?)? That will help addressing b) in an optimal way.~

edit: Following feeback, implementation now uses `NodeEvent`s. Support is limited to interactive-tx in a splice as a proof of concept.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Channel.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Channel.kt
@@ -78,10 +78,26 @@ sealed class ChannelState {
             is Syncing -> this@ChannelState.state
             else -> this@ChannelState
         }
+        maybeSignalSensitiveTask(oldState, newState)
         return when {
             // we only want to fire the PaymentSent event when we transition to Closing for the first time
             oldState is ChannelStateWithCommitments && oldState !is Closing && newState is Closing -> emitClosingEvents(oldState, newState)
             else -> emptyList()
+        }
+    }
+
+    /** Some transitions imply that we are in the middle of tasks that may require some time. */
+    private fun ChannelContext.maybeSignalSensitiveTask(oldState: ChannelState, newState: ChannelState) {
+        val spliceStatusBefore = (oldState as? Normal)?.spliceStatus
+        val spliceStatusAfter = (newState as? Normal)?.spliceStatus
+        when {
+            spliceStatusBefore !is SpliceStatus.InProgress && spliceStatusAfter is SpliceStatus.InProgress -> // splice initiated
+                staticParams.nodeParams._nodeEvents.tryEmit(SensitiveTaskEvents.TaskStarted(SensitiveTaskEvents.TaskIdentifier.InteractiveTx(spliceStatusAfter.spliceSession.fundingParams)))
+            spliceStatusBefore is SpliceStatus.InProgress && spliceStatusAfter !is SpliceStatus.WaitingForSigs -> // splice aborted before reaching signing phase
+                staticParams.nodeParams._nodeEvents.tryEmit(SensitiveTaskEvents.TaskEnded(SensitiveTaskEvents.TaskIdentifier.InteractiveTx(spliceStatusBefore.spliceSession.fundingParams)))
+            spliceStatusBefore is SpliceStatus.WaitingForSigs && spliceStatusAfter !is SpliceStatus.WaitingForSigs -> // splice leaving signing phase (successfully or not)
+                staticParams.nodeParams._nodeEvents.tryEmit(SensitiveTaskEvents.TaskEnded(SensitiveTaskEvents.TaskIdentifier.InteractiveTx(spliceStatusBefore.session.fundingParams)))
+            else -> {}
         }
     }
 
--- a/src/commonMain/kotlin/fr/acinq/lightning/NodeEvents.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/NodeEvents.kt
@@ -2,6 +2,8 @@ package fr.acinq.lightning
 
 import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.bitcoin.Satoshi
+import fr.acinq.lightning.channel.InteractiveTxParams
+import fr.acinq.lightning.channel.SharedFundingInput
 import fr.acinq.lightning.channel.states.ChannelStateWithCommitments
 import fr.acinq.lightning.channel.states.Normal
 import fr.acinq.lightning.channel.states.WaitForFundingCreated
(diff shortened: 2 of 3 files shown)


robbiehanson:
The API that Apple gives us is [beginBackgroundTask(expirationHandler:)](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask).

You call this to "mark the start of a task that should continue if the app enters the background."

You can call this method as many times as you want. If the app is performing 101 "sensitive" tasks, then you can call this method 101 times.

Each time you call the method it returns a UUID. And you MUST balance each call with a matching call to `endBackgroundTask(theUUID)`.

When the user moves the app into the background, iOS checks to see if the app's `activeBackgroundTasks: Set<UUID>` is non-empty. If so, it will give the app up to 3 minutes to finish it's work. (That is, to call `endBackgroundTask(:)` for each outstanding UUID). If the app fails to do this, then it will be killed.

There are also penalties for abusing this system. If your app calls this all the time, then Apple will start giving you less and less time.

We currently use the `beginBackgroundTask(:)` API to keep the app alive when an outgoing payment is in-flight. For this we make use of the PeerEvents `PaymentProgress`, `PaymentSent`, `PaymentNotSent`. This works really well because each event includes a unique paymentID that we can track.

So this might work better if it was converted into various PeerEvent's. One when the `PayToOpenRequest` starts, and another when it completes (either Failure or Success).

pm47:
I've only implemented the mechanism for splices (this is the main/unique case we see in the wild at the moment), but I'm not really satisfied: this is a bit invasive and all over the place. Let me know if you have a better idea @t-bast

t-bast:
I agree, it is error-prone to have to insert this in various places in the flow, it's likely that we will miss one. I think that it would be simpler and less error-prone to insert this when introspecting outgoing messages, like the way we insert the channel backup in `maybeAddBackupToMessages`.

When processing `ChannelAction.Message.Send`, we could do the following:

- if the message is `splice_init` or `splice_ack`, emit `TaskStarted`
- if the message is `tx_abort`, emit `TaskEnded`
- if the message is `tx_signatures`, emit `TaskEnded`

There is a subtlety for the `TaskEnded` case: we may retransmit `tx_signatures`/`tx_abort` on reconnection, which means we may emit `TaskEnded` multiple times for the same task. Is that an issue? I'd expect iOS to simply ignore redundant requests to end a task, so it shouldn't be an issue, but I may be wrong.

src/commonMain/kotlin/fr/acinq/lightning/channel/states/Channel.kt
```diff
@@ -78,13 +78,29 @@ sealed class ChannelState {
...
+    private fun ChannelContext.maybeSignalSensitiveTask(oldState: ChannelState, newState: ChannelState) {
+        val spliceStatusBefore = (oldState as? Normal)?.spliceStatus
+        val spliceStatusAfter = (newState as? Normal)?.spliceStatus
+        when {
+            spliceStatusBefore !is SpliceStatus.InProgress && spliceStatusAfter is SpliceStatus.InProgress -> // splice initiated
+                staticParams.nodeParams._nodeEvents.tryEmit(SensitiveTaskEvents.TaskStarted(SensitiveTaskEvents.TaskIdentifier.InteractiveTx(spliceStatusAfter.spliceSession.fundingParams)))
+            spliceStatusBefore is SpliceStatus.InProgress && spliceStatusAfter !is SpliceStatus.WaitingForSigs -> // splice aborted before reaching signing phase
+                staticParams.nodeParams._nodeEvents.tryEmit(SensitiveTaskEvents.TaskEnded(SensitiveTaskEvents.TaskIdentifier.InteractiveTx(spliceStatusBefore.spliceSession.fundingParams)))
+            spliceStatusBefore is SpliceStatus.WaitingForSigs && spliceStatusAfter !is SpliceStatus.WaitingForSigs -> // splice leaving signing phase (successfully or not)
```
pm47, line 98: The argument for considering the process done in the success case when leaving `SpliceStatus.WaitingForSigs` is that we have already either sent or received `tx_signatures` (not both). The splice could in theory be failed after restart, but the implementation on lightning-kmp and eclair guarantees that it will not be the case.

pm47:
> I think that it would be simpler and less error-prone to insert this when introspecting outgoing messages

I did a variant that looks at state transition in [73dc969](https://github.com/ACINQ/lightning-kmp/pull/550/commits/73dc969ee1c81fce29819381060eecae016621df). It looks much cleaner, hope it's correct 😅

src/commonMain/kotlin/fr/acinq/lightning/channel/states/Channel.kt
```diff
@@ -78,13 +78,29 @@ sealed class ChannelState {
...
+        maybeSignalSensitiveTask(oldState, newState)
         return when {
             // we only want to fire the PaymentSent event when we transition to Closing for the first time
             oldState is ChannelStateWithCommitments && oldState !is Closing && newState is Closing -> emitClosingEvents(oldState, newState)
             else -> emptyList()
         }
     }
 
+    /** Some transitions imply that we are in the middle of tasks that may require some time. */
+    private fun ChannelContext.maybeSignalSensitiveTask(oldState: ChannelState, newState: ChannelState) {
```
t-bast, line 90: Don't you want to also handle initial channel creation here? Or maybe we'll do that in a separate, follow-up PR? I think this mechanism would still work, we should be able to handle those transitions here.

t-bast (approved):

src/commonMain/kotlin/fr/acinq/lightning/channel/states/Channel.kt
t-bast: That looks correct! Using the transitions on the `spliceStatus` really makes a lot of sense here :+1:

t-bast: > We have already either sent or received-then-sent tx_signatures (same for our peer). 

That even guarantees that we won't forget this splice, because either us or our peer has enough signatures to broadcast it. The only way we could forget it in that case is if it was double-spent, but 1) we don't detect that in lightning-kmp and 2) we explicitly avoid double-spending those in eclair

PR #551, opened 2023-10-16
Fix `update_fee` reserve requirement mismatch

t-bast:
The splicing update removed the `channelReserve` field from `LocalParams` and always uses a value of 1% of the channel capacity. But that is incorrect for channels that were created before that update and haven't been migrated.

For such channels, we will use an incorrect channel reserve which can end up force-closing channels when receiving otherwise valid `update_fee` messages. We temporarily ignore the reserve in this computation and will re-enable it once all users have been migrated to use splicing.

We also fix a mismatch between the local and remote dust limit used when computing commit tx fees.


pm47 (approved):
Nice catch, the temporary fix makes sense.

PR #553, opened 2023-10-16
Send commit sigs for alternative feerates

t-bast:
When the commitment transaction doesn't contain any HTLC, we send additional signatures for alternative feerates. This lets the wallet provider force-close with a more interesting feerate when the wallet user disappears without closing their channels.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
@@ -5,6 +5,7 @@ import fr.acinq.bitcoin.Crypto.sha256
 import fr.acinq.lightning.CltvExpiryDelta
 import fr.acinq.lightning.Feature
 import fr.acinq.lightning.MilliSatoshi
+import fr.acinq.lightning.blockchain.fee.FeeratePerByte
 import fr.acinq.lightning.blockchain.fee.FeeratePerKw
 import fr.acinq.lightning.blockchain.fee.FeerateTolerance
 import fr.acinq.lightning.channel.states.Channel
@@ -421,7 +422,18 @@ data class Commitment(
             "built remote commit number=${remoteCommit.index + 1} toLocalMsat=${spec.toLocal.toLong()} toRemoteMsat=${spec.toRemote.toLong()} htlc_in=$htlcsIn htlc_out=$htlcsOut feeratePerKw=${spec.feerate} txId=${remoteCommitTx.tx.txid} fundingTxId=$fundingTxId"
         }
 
-        val commitSig = CommitSig(params.channelId, sig, htlcSigs.toList())
+        val commitSig = when (spec.htlcs.size) {
+            0 -> {
+                val alternativeSigs = Commitments.alternativeFeerates.map { feerate ->
+                    val alternativeSpec = spec.copy(feerate = feerate)
+                    val (alternativeRemoteCommitTx, _) = Commitments.makeRemoteTxs(channelKeys, commitTxNumber = remoteCommit.index + 1, params.localParams, params.remoteParams, fundingTxIndex = fundingTxIndex, remoteFundingPubKey = remoteFundingPubkey, commitInput, remotePerCommitmentPoint = remoteNextPerCommitmentPoint, alternativeSpec)
+                    val alternativeSig = Transactions.sign(alternativeRemoteCommitTx, channelKeys.fundingKey(fundingTxIndex))
+                    CommitSigTlv.AlternativeFeerateSig(feerate, alternativeSig)
+                }
+                CommitSig(params.channelId, sig, listOf(), TlvStream(CommitSigTlv.AlternativeFeerateSigs(alternativeSigs)))
+            }
+            else -> CommitSig(params.channelId, sig, htlcSigs.toList())
+        }
         val commitment1 = copy(nextRemoteCommit = NextRemoteCommit(commitSig, RemoteCommit(remoteCommit.index + 1, spec, remoteCommitTx.tx.txid, remoteNextPerCommitmentPoint)))
         return Pair(commitment1, commitSig)
     }
@@ -932,6 +944,12 @@ data class Commitments(
         const val HTLC_OUTPUT_WEIGHT = 172
         const val HTLC_TIMEOUT_WEIGHT = 666
         const val HTLC_SUCCESS_WEIGHT = 706
+
+        /**
+         * Alternative feerates at which we will sign commitment transactions that have no pending HTLCs.
+         * WARNING: never remove a feerate from this list, we can only add more, otherwise we will not be able to detect when our peer broadcasts the commit tx at the removed feerate.
+         */
+        val alternativeFeerates = listOf(1.sat, 2.sat, 5.sat, 10.sat).map { FeeratePerKw(FeeratePerByte(it)) }
 
         fun makeLocalTxs(
             channelKeys: KeyManager.ChannelKeys,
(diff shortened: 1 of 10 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/states/Channel.kt
```diff
@@ -501,18 +506,48 @@ sealed class ChannelStateWithCommitments : PersistedChannelState() {
...
+                        commitments.latest.nextRemoteCommit?.let { add(it.commit) }
+                    }.filter { remoteCommit ->
+                        remoteCommit.spec.htlcs.isEmpty()
+                    }.flatMap { remoteCommit ->
+                        Commitments.alternativeFeerates.map { feerate ->
+                            val alternativeSpec = remoteCommit.spec.copy(feerate = feerate)
+                            val (alternativeRemoteCommitTx, _) = Commitments.makeRemoteTxs(channelKeys(), remoteCommit.index, commitments.params.localParams, commitments.params.remoteParams, commitments.latest.fundingTxIndex, commitments.latest.remoteFundingPubkey, commitments.latest.commitInput, remoteCommit.remotePerCommitmentPoint, alternativeSpec)
+                            RemoteCommit(remoteCommit.index, alternativeSpec, alternativeRemoteCommitTx.tx.txid, remoteCommit.remotePerCommitmentPoint)
+                        }
+                    }
```
pm47, line 521: Could we move this to `Commitments.alternativeFeerateCommits()` right under `Commitments.alternativeFeerates` (which could be made private)?

This would simply a bit this pretty scary `handleRemoteSpentOther()` function and contain the alternate sigs stuff in one place.

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -421,7 +422,18 @@ data class Commitment(
...
+                val alternativeSigs = Commitments.alternativeFeerates.map { feerate ->
+                    val alternativeSpec = spec.copy(feerate = feerate)
+                    val (alternativeRemoteCommitTx, _) = Commitments.makeRemoteTxs(channelKeys, commitTxNumber = remoteCommit.index + 1, params.localParams, params.remoteParams, fundingTxIndex = fundingTxIndex, remoteFundingPubKey = remoteFundingPubkey, commitInput, remotePerCommitmentPoint = remoteNextPerCommitmentPoint, alternativeSpec)
+                    val alternativeSig = Transactions.sign(alternativeRemoteCommitTx, channelKeys.fundingKey(fundingTxIndex))
+                    CommitSigTlv.AlternativeFeerateSig(feerate, alternativeSig)
+                }
+                CommitSig(params.channelId, sig, listOf(), TlvStream(CommitSigTlv.AlternativeFeerateSigs(alternativeSigs)))
+            }
+            else -> CommitSig(params.channelId, sig, htlcSigs.toList())
+        }
```
pm47, line 436: Nit: I would factor the declaration of `CommitSig()`:
```suggestion
        val tlvs = buildSet {
            if (spec.htlcs.isEmpty()) {
                val alternativeSigs = Commitments.alternativeFeerates.map { feerate ->
                    val alternativeSpec = spec.copy(feerate = feerate)
                    val (alternativeRemoteCommitTx, _) = Commitments.makeRemoteTxs(channelKeys, commitTxNumber = remoteCommit.index + 1, params.localParams, params.remoteParams, fundingTxIndex = fundingTxIndex, remoteFundingPubKey = remoteFundingPubkey, commitInput, remotePerCommitmentPoint = remoteNextPerCommitmentPoint, alternativeSpec)
                    val alternativeSig = Transactions.sign(alternativeRemoteCommitTx, channelKeys.fundingKey(fundingTxIndex))
                    CommitSigTlv.AlternativeFeerateSig(feerate, alternativeSig)
                }
                add(CommitSigTlv.AlternativeFeerateSigs(alternativeSigs))
            }
        }
        val commitSig = CommitSig(params.channelId, sig, htlcSigs.toList(), TlvStream(tlvs)) 
```

pm47 (commented):
Only nits, otherwise LGTM.

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/553/commits/abee6c6eb1fd120d49c653829380478973bf2ce7

src/commonMain/kotlin/fr/acinq/lightning/channel/states/Channel.kt
t-bast: Agreed, done in https://github.com/ACINQ/lightning-kmp/pull/553/commits/abee6c6eb1fd120d49c653829380478973bf2ce7
I didn't make `alternativeFeerates` private though, because that would require moving the alternative sig generation in a helper function, and I think it's clearer to have it inline in `sendCommit` right below the "normal" signature.

pm47 (approved):
LGTM, please let me do a real test with Phoenix on testnet before merging, as it's not entirely obvious to me if it works seamlessly on the eclair side.

pm47:
> please let me do a real test with Phoenix on testnet before merging

Test successful, I even RBFed the commit tx with higher custom commit sigs.

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -421,7 +422,18 @@ data class Commitment(
...
+        val tlvs = buildSet {
+            if (spec.htlcs.isEmpty()) {
+                val alternativeSigs = Commitments.alternativeFeerates.map { feerate ->
+                    val alternativeSpec = spec.copy(feerate = feerate)
+                    val (alternativeRemoteCommitTx, _) = Commitments.makeRemoteTxs(channelKeys, commitTxNumber = remoteCommit.index + 1, params.localParams, params.remoteParams, fundingTxIndex = fundingTxIndex, remoteFundingPubKey = remoteFundingPubkey, commitInput, remotePerCommitmentPoint = remoteNextPerCommitmentPoint, alternativeSpec)
+                    val alternativeSig = Transactions.sign(alternativeRemoteCommitTx, channelKeys.fundingKey(fundingTxIndex))
+                    CommitSigTlv.AlternativeFeerateSig(feerate, alternativeSig)
+                }
+                add(CommitSigTlv.AlternativeFeerateSigs(alternativeSigs))
+            }
```
pm47, line 434: This code should be called by InteractiveTx.create(), otherwise we are not providing additional sigs at channel creation and for splice txs.

pm47 (changes requested):

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
t-bast: Good catch, I'll add that.

t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/553/commits/8032630c6533153d7f226be5069b5bbebb4111d4 with additional tests.

t-bast:
Rebased to fix import issue.

pm47 (approved):

remyers:
Do we want to extend this to also support sending htlc sigs when splices contain htlcs? or just skip sending these tlvs when htlc's are present?

t-bast:
We're only sending those additional sigs when the commitment is empty (no HTLCs), so there's nothing to do in the case where it's a splice with HTLCs.

PR #554, opened 2023-10-17
Fix `tx_signatures` ordering for splices

t-bast:
We were incorrectly splitting the shared input amount based on each peer's balance, but for the signing order the peer that adds the shared input takes credit for the whole amount of that input.


t-bast:
We can merge this PR now, our node contains the fix and a work-around to follow both the old and new way of choosing the signature order :+1:

pm47 (approved):

PR #556, opened 2023-10-23
Apply 0-reserve to both peers

### LINKED ISSUE #552: Restore channel reserve computation when receiving `update_fee`
See https://github.com/ACINQ/lightning-kmp/pull/551

t-bast:
We previously only applied 0-reserve to the non-initiator when the feature bit was active. We should apply it to both sides of the channel.

This reduces the liquidity requirements on the initiator side, and allows receiving more funds over lightning before requiring a splice to get more inbound liquidity.

This doesn't create any backwards-compatibility issue on the `lightning-kmp` side, because if we release with that change, we simply allow our peer to dip into their reserve, which they may or may not do (so we can merge this PR whenever we want). On the eclair side, we can "activate" this when the Phoenix node also supports `quiescence`, which implicitly means that it contains this change. In the longer term, we can clean-up those requirements as being simply dependent on the 0-reserve feature bit.

Fixes #552

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
@@ -146,12 +146,12 @@ data class Commitment(
     val fundingAmount: Satoshi = commitInput.txOut.amount
 
     fun localChannelReserve(params: ChannelParams): Satoshi = when {
-        params.channelFeatures.hasFeature(Feature.ZeroReserveChannels) && !params.localParams.isInitiator -> 0.sat
+        params.channelFeatures.hasFeature(Feature.ZeroReserveChannels) -> 0.sat
         else -> (fundingAmount / 100).max(params.remoteParams.dustLimit)
     }
 
     fun remoteChannelReserve(params: ChannelParams): Satoshi = when {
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -515,11 +508,11 @@ data class FullCommitment(
     val fundingTxId: ByteVector32 = commitInput.outPoint.txid
     val fundingAmount = commitInput.txOut.amount
     val localChannelReserve = when {
-        params.channelFeatures.hasFeature(Feature.ZeroReserveChannels) && !params.localParams.isInitiator -> 0.sat
+        params.channelFeatures.hasFeature(Feature.ZeroReserveChannels) -> 0.sat
```
pm47, line 511: The duplication between `Commitments` and `FullCommitments` shows that there is some cleanup to do

pm47 (approved):

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
t-bast: True, but I'm not sure how it could be cleaned up though, so I'm leaving it as-is for now.

PR #559, opened 2023-11-20
Don't double-spend unspendable inputs

t-bast:
When filtering inputs that can be used for swaps, we filter out all the inputs that are already used in pending funding transactions. However, we ignored previously confirmed funding transactions, assuming that our electrum server would have already filtered them from our wallet state.

Unfortunately, we cannot assume that our electrum server is up-to-date (especially if we connect to a different one after a restart) and behaving correctly, so we must also exclude inputs use in the latest confirmed funding transaction.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/SwapInManager.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/SwapInManager.kt
@@ -57,31 +57,34 @@ class SwapInManager(private var reservedUtxos: Set<OutPoint>, private val logger
 
     companion object {
         /**
-         * Return the list of wallet inputs used in pending unconfirmed channel funding attempts.
+         * Return the list of wallet inputs already used in channel funding attempts.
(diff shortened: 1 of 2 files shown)


pm47 (commented):
LGTM, but why did you remove the factorization on `fundingTx.tx.localInputs.map { it.outPoint }`?

t-bast:
Because this isn't using the same type anymore, we cannot factor `ConfirmedFundingTx` which only has a `Transaction`, not a `SharedTransaction`.

pm47 (approved):
> Because this isn't using the same type anymore, we cannot factor ConfirmedFundingTx which only has a Transaction, not a SharedTransaction.

You could stil factor on `Transaction`, but I suppose you didn't want to call `sharedTx.tx.buildUnsignedTx()`? Anyway that'd fine.

t-bast:
> You could stil factor on Transaction, but I suppose you didn't want to call sharedTx.tx.buildUnsignedTx()?

Exactly, we'd be losing information if we did that, and would be including remote inputs as well.
When we only have a `Transaction` we can't avoid it, but when we can, I think we should, it's more correct that way.

PR #560, opened 2023-11-23
Improve swap-in protocol with taproot and musig2

sstone:
This PR implements an evolution of our p2swh-based swap-in that uses Musig2 and taproot for better privacy and lower on-chain fees.

Our swap-in protocol becomes `user key + server key OR user refund key + delay`, and is implemented with taproot which provides a clean separation between the mutual use case (handled by the internal key, which is a Musig2 aggregation of the user's and server's key) and the refund case (handled by a taproot script with a single leaf: the refund script).

Musig2 requires both parties to generate secret nonces and exchange public nonces before they can compute partial Musig2 signatures, which can then be combined into a single Schnorr signature:

- A new optional TLV is added to `TxComplete`, which includes public nonces for all swap-in inputs
- A new optional TLV is added to `TxSignatures`, which includes partial Musig2 signatures for all swap-in inputs

Nonces are never persisted and their lifecycle is tied to the interactive tx session lifecycle: while the session is alive, it remembers nonces that it sent before i.e if `TxComplete` is sent several times, nonces for the same input id will be re-used. I believe that it is safe since they will be used only once for signing.

A specific user refund key is used in the refund script: this allows us to easily rotate swap-in addresses (each time we change th user refund key) and generate a single taproot descriptor that will allow bitcoin core to find and spend (after the refund delay) all swap-in transactions for a given user (i.e for a given phoenix wallet). 

This is a first step towards fixing https://github.com/ACINQ/phoenix/issues/403: this new scheme makes swap-in address rotation easy to implement but this is out of the scope of this PR.

Swap-in addresses are now regular `p2tr` addresses, and swap-in transactions become indistinguishable from other `p2tr` transactions.
And since spending swap-in inputs requires a single 64 bytes Schnorr transaction (instead of 2 73 bytes DER transaction + a 77 bytes redeem script), the weight of a swap-in input is reduced by 40% for cheaper on-chain fees.


sstone:
superseded by https://github.com/ACINQ/lightning-kmp/pull/563

PR #561, opened 2023-11-29
Liquidity Ads

t-bast:
Implement a prototype for liquidity ads, compatible with https://github.com/ACINQ/eclair/pull/2550
Note that we only implement the buyer side, which limits testing.
The specification is available here: https://github.com/lightning/bolts/pull/878

We currently don't add CLTV locks to the commitment transactions, for simplicity's sake.
We thus don't store yet the lease in the commitments (to be defined where we should store it, and what data we need).


src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
```diff
@@ -0,0 +1,153 @@
...
+                leaseFeeProportional = LightningCodecs.u16(input),
+                maxRelayFeeProportional = LightningCodecs.u16(input),
+                leaseFeeBase = LightningCodecs.u32(input).sat,
+                maxRelayFeeBase = LightningCodecs.tu32(input).msat,
+            )
+        }
+    }
+
+    /** Request inbound liquidity from a remote peer that supports liquidity ads. */
+    data class RequestRemoteFunding(val fundingAmount: Satoshi, val maxFee: Satoshi, val leaseStart: Int, val leaseDuration: Int) {
```
pm47, line 76: I think it would make more sense to provide the actual liquidity parameters here, (that we would get either from conf, our direct peer, network announcements...).

```suggestion
    data class RequestRemoteFunding(val fundingAmount: Satoshi, val leaseRates: LeaseRates) {
```

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -561,6 +562,25 @@ class Peer(
...
                     spliceOut = null,
+                    requestRemoteFunding = null,
+                    feerate = feerate
+                )
+                send(WrappedChannelCommand(channel.channelId, spliceCommand))
+                spliceCommand.replyTo.await()
+            }
+    }
+
+    suspend fun purchaseInboundLiquidity(amount: Satoshi, feerate: FeeratePerKw, maxFee: Satoshi, leaseDuration: Int): ChannelCommand.Commitment.Splice.Response? {
```
pm47, line 573: For ease of integration:

```suggestion
    suspend fun purchaseInboundLiquidity(amount: Satoshi, feerate: FeeratePerKw, leaseRates: LeaseRates): ChannelCommand.Commitment.Splice.Response? {
```

Assuming that lease rates are an attribute of the `Peer`, or of `NodeParams`, this could be simplified to:

```suggestion
    suspend fun purchaseInboundLiquidity(amount: Satoshi, feerate: FeeratePerKw): ChannelCommand.Commitment.Splice.Response? {
```

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
```diff
@@ -0,0 +1,153 @@
...
+                    val witness = LeaseWitness(remoteFundingPubKey, leaseExpiry, leaseDuration, willFund.leaseRates.maxRelayFeeProportional, willFund.leaseRates.maxRelayFeeBase)
+                    val fees = willFund.leaseRates.fees(fundingFeerate, fundingAmount, remoteFundingAmount)
+                    return if (!LeaseWitness.verify(remoteNodeId, willFund.sig, witness)) {
+                        Either.Left(InvalidLiquidityAdsSig(channelId))
+                    } else if (remoteFundingAmount <= 0.sat) {
+                        Either.Left(LiquidityRatesRejected(channelId))
+                    } else if (maxFee < fees) {
+                        Either.Left(LiquidityRatesRejected(channelId))
+                    } else {
+                        val leaseAmount = fundingAmount.min(remoteFundingAmount)
```
pm47, line 102: I think we should check here that the remote amount is exactly what was requested, as users will probably reason in terms of lease_fee/mining_fee ratio.

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
```diff
@@ -0,0 +1,153 @@
...
+     * that they misbehaved.
+     */
+    data class LeaseRates(val fundingWeight: Int, val leaseFeeProportional: Int, val maxRelayFeeProportional: Int, val leaseFeeBase: Satoshi, val maxRelayFeeBase: MilliSatoshi) {
+        val maxRelayFeeProportionalMillionths: Long = maxRelayFeeProportional.toLong() * 100
+
+        /**
+         * Fees paid by the liquidity buyer: the resulting amount must be added to the seller's output in the corresponding
+         * commitment transaction.
+         */
+        fun fees(feerate: FeeratePerKw, requestedAmount: Satoshi, contributedAmount: Satoshi): Satoshi {
```
pm47, line 43: Integrators will find it useful to have a split between mining fees and "service" fees:

```suggestion
        data class LeaseFees(miningFee: Satoshi, serviceFee: Satoshi) {
          val fees: Satoshi = miningFee + serviceFee
        }

        fun fees(feerate: FeeratePerKw, requestedAmount: Satoshi, contributedAmount: Satoshi): LeaseFees {
```

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
```diff
@@ -0,0 +1,153 @@
...
+        val expiry: Int = witness.leaseEnd
+    }
+
+    /** The seller signs the lease parameters: if they cheat, the buyer can use that signature to prove they cheated. */
+    data class LeaseWitness(val fundingPubKey: PublicKey, val leaseEnd: Int, val leaseDuration: Int, val maxRelayFeeProportional: Int, val maxRelayFeeBase: MilliSatoshi) {
+        fun sign(nodeKey: PrivateKey): ByteVector64 = Crypto.sign(Crypto.sha256(encode()), nodeKey)
+
+        fun encode(): ByteArray {
+            val out = ByteArrayOutput()
+            LightningCodecs.writeBytes("option_will_fund".encodeToByteArray(), out)
```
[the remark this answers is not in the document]
t-bast: I can't find that in the classpath...are you sure this doesn't require a new dependency?

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
t-bast: Good idea, done in https://github.com/ACINQ/lightning-kmp/pull/561/commits/c65c4cf919c3572095bc997868e8a01dbabb8ede

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
t-bast: I changed that in https://github.com/ACINQ/lightning-kmp/pull/561/commits/c65c4cf919c3572095bc997868e8a01dbabb8ede, we now provide the rates as input and only accept the splice attempt if the rates haven't changed and the LSP adds at least the requested amount.

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/561/commits/c65c4cf919c3572095bc997868e8a01dbabb8ede

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
```diff
@@ -0,0 +1,153 @@
...
+            )
+        }
+    }
+
+    /** Request inbound liquidity from a remote peer that supports liquidity ads. */
+    data class RequestRemoteFunding(val fundingAmount: Satoshi, val maxFee: Satoshi, val leaseStart: Int, val leaseDuration: Int) {
+        private val leaseExpiry: Int = leaseStart + leaseDuration
+        val requestFunds: ChannelTlv.RequestFunds = ChannelTlv.RequestFunds(fundingAmount, leaseExpiry, leaseDuration)
+
+        fun validateLeaseRates(
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/561/commits/c65c4cf919c3572095bc997868e8a01dbabb8ede

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
```diff
@@ -0,0 +1,153 @@
...
+object LiquidityAds {
+
+    /**
+     * Liquidity is leased using the following rates:
+     *
+     *  - the buyer pays [[leaseFeeBase]] regardless of the amount contributed by the seller
+     *  - the buyer pays [[leaseFeeProportional]] (expressed in basis points) of the amount contributed by the seller
+     *  - the buyer refunds the on-chain fees for up to [[fundingWeight]] of the utxos contributed by the seller
+     *
+     * The seller promises that their relay fees towards the buyer will never exceed [[maxRelayFeeBase]] and [[maxRelayFeeProportional]].
```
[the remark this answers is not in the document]
t-bast: I updated the comment in https://github.com/ACINQ/lightning-kmp/pull/561/commits/c65c4cf919c3572095bc997868e8a01dbabb8ede

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
```diff
@@ -0,0 +1,158 @@
...
+     *  - the buyer pays [leaseFeeBase] regardless of the amount contributed by the seller
+     *  - the buyer pays [leaseFeeProportional] (expressed in basis points) of the amount contributed by the seller
+     *  - the seller will have to add inputs/outputs to the transaction and pay on-chain fees for them, but the buyer
+     *    refunds on-chain fees for [fundingWeight] vbytes
+     *
+     * The seller promises that their relay fees towards the buyer will never exceed [maxRelayFeeBase] and [maxRelayFeeProportional].
+     * This cannot be enforced, but if the buyer notices the seller cheating, they should blacklist them and can prove
+     * that they misbehaved using the seller's signature of the [LeaseWitness].
+     */
+    data class LeaseRate(val leaseDuration: Int, val fundingWeight: Int, val leaseFeeProportional: Int, val leaseFeeBase: Satoshi, val maxRelayFeeProportional: Int, val maxRelayFeeBase: MilliSatoshi) {
```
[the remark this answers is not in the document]
t-bast: > Nit: traditionally, base is before prop:

Lisa chose to include the proportional fees first in the spec messages, that's why I kept it that way (because in eclair we'll have to follow the encoding order, and it's nice to be consistent with eclair).

> Also, prop units are misleading: leaseFeeProportional is in per-10,000, maxRelayFeeProportional is in per-1,000,000.

Here, they are both basis points for consistency (which I agree is confusing, but there has to be confusion somewhere while we're moving to basis points instead of ppm). I can rename them to `ProportionalBasisPoint` to clarify though!

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -168,6 +168,7 @@ class Peer(
     val currentTipFlow = MutableStateFlow<Pair<Int, BlockHeader>?>(null)
     val onChainFeeratesFlow = MutableStateFlow<OnChainFeerates?>(null)
     val swapInFeeratesFlow = MutableStateFlow<FeeratePerKw?>(null)
+    val liquidityRatesFlow = MutableStateFlow<LiquidityAds.LeaseRate?>(null)
```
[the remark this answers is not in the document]
t-bast: As discussed offline, I'll remove this from `init` while this isn't the final spec version, it will instead be provided out of band.

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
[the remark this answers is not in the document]
pm47: Noted thanks!

Okay, so the convention is that "proportional" is bips by default, I think that's fine.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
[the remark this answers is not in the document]
pm47: It's independent I think? In my tests I'm setting the flow locally, it works fine. We can probably figure this out later as for now we'll only have one rate.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Serialization.kt
```diff
@@ -356,31 +357,68 @@ object Serialization {
...
+        writeNumber(lease.witness.maxRelayFeeProportional)
+        writeNumber(lease.witness.maxRelayFeeBase.toLong())
+    }
+
     private fun Output.writeInteractiveTxSigningSession(s: InteractiveTxSigningSession) = s.run {
         writeInteractiveTxParams(fundingParams)
         writeNumber(s.fundingTxIndex)
         writeSignedSharedTransaction(fundingTx)
         // We don't bother removing the duplication across HTLCs: this is a short-lived state during which the channel cannot be used for payments.
-        writeEither(localCommit,
```
pm47, line 364: I think this conveys the intent better (NB: `writeEither` uses `write`, not `writeNumber`, for discriminators)

```kotlin
    private fun Output.writeInteractiveTxSigningSession(s: InteractiveTxSigningSession) = s.run {
        writeInteractiveTxParams(fundingParams)
        writeNumber(s.fundingTxIndex)
        writeSignedSharedTransaction(fundingTx)
        // We use 2/3 instead of the standard 0/1 as discriminator for the option to handle backward compatibility (this field was added later)
        when (liquidityPurchased) {
            null -> write(2)
            else -> {
                write(3)
                writeLiquidityPurchase(liquidityPurchased)
            }
        }
        // We don't bother removing the duplication across HTLCs: this is a short-lived state during which the channel cannot be used for payments.
        writeEither(localCommit,
            writeLeft = { writeUnsignedLocalCommitWithHtlcs(it) },
            writeRight = { writeLocalCommitWithHtlcs(it) }
        )
        remoteCommit.run {
            writeNumber(index)
            writeCommitmentSpecWithHtlcs(spec)
            writeTxId(txid)
            writePublicKey(remotePerCommitmentPoint)
        }
    }
```

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
```diff
@@ -307,43 +308,60 @@ object Deserialization {
...
+            maxRelayFeeBase = readNumber().msat,
         ),
-        remoteCommit = RemoteCommit(
+    )
+
+    private fun Input.readInteractiveTxSigningSession(): InteractiveTxSigningSession {
+        val fundingParams = readInteractiveTxParams()
+        val fundingTxIndex = readNumber()
+        val fundingTx = readSignedSharedTransaction() as PartiallySignedSharedTransaction
+        val (liquidityPurchase, localCommit) = when (val discriminator = read()) {
```
pm47, line 350:
```suggestion
        // liquidityPurchase and localCommit are logically independent, this is just a serialization trick for backward compatibility
        val (liquidityPurchase, localCommit) = when (val discriminator = read()) {
```

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -566,6 +568,31 @@ class Peer(
...
                     spliceOut = null,
+                    requestRemoteFunding = null,
+                    feerate = feerate
+                )
+                send(WrappedChannelCommand(channel.channelId, spliceCommand))
+                spliceCommand.replyTo.await()
+            }
+    }
+
+    suspend fun estimateFeeForInboundLiquidity(amount: Satoshi, feerate: FeeratePerKw): LiquidityAds.LeaseFees {
```
pm47, line 579: This should take a `targetFeerate` and return the corrected feerate (taking into account unconfirmed parents), like we do for splice-out and cpfp. (also the ordering of methods is not consistent)

```suggestion
    suspend fun estimateFeeForInboundLiquidity(amount: Satoshi, targetFeerate: FeeratePerKw): Pair<FeeratePerKw, LiquidityAds.LeaseFees> {
```

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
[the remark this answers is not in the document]
t-bast: Right, let's start with a single rate, and we can later use a list when we'll need to handle backwards-compatibility if we change our rates.

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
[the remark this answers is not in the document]
pm47: Okay, this is JVM only.

There is a `.toByteArray(Charsets.UTF_8)` used in `ElectrumClientTest`, but it requires `import io.ktor.utils.io.core.*`. Let's forget about this...

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
t-bast: Added in https://github.com/ACINQ/lightning-kmp/pull/561/commits/3541777350fdae4b0ac92cbf55d4e52ae1c200c5

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
t-bast: Good idea, done in https://github.com/ACINQ/lightning-kmp/pull/561/commits/cb0c0f80900c95f03f4be9c335decc046229958b

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Serialization.kt
pm47: Right, nevermind!

src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
```diff
@@ -24,7 +24,8 @@ data class Normal(
     val localShutdown: Shutdown?,
     val remoteShutdown: Shutdown?,
     val closingFeerates: ClosingFeerates?,
-    val spliceStatus: SpliceStatus
+    val spliceStatus: SpliceStatus,
+    val liquidityPurchases: List<LiquidityAds.Lease>,
```
pm47, line 28: Nit:
```suggestion
    val liquidityLeases: List<LiquidityAds.Lease>,
```

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Serialization.kt
```diff
@@ -144,6 +145,11 @@ object Serialization {
                 write(0x00)
             }
         }
+        if (liquidityPurchases.isNotEmpty()) {
+            write(0x01)
+            writeNumber(liquidityPurchases.size)
+            liquidityPurchases.forEach { writeLiquidityPurchase(it) }
```
pm47, line 151:
```suggestion
            writeCollection(liquidityPurchases) { writeLiquidityPurchase(it) }
```

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
```diff
@@ -99,6 +100,16 @@ object Deserialization {
...
             )
             else -> error("unknown discriminator $discriminator for class ${SpliceStatus::class}")
+        },
+        liquidityPurchases = when {
+            availableBytes == 0 -> listOf()
+            else -> when (val discriminator = read()) {
+                0x01 -> {
+                    val liquidityPurchaseCount = readNumber()
+                    (0 until liquidityPurchaseCount).map { readLiquidityPurchase() }
+                }
```
pm47, line 110:
```suggestion
                0x01 -> readCollection { readLiquidityPurchase() }.toList()
```

pm47 (commented):
Nice! Some nits

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Serialization.kt
pm47: On 2nd thought I think we should also remove the `if`, and always write the `0x01` and the leases, even if the list is empty. Otherwise, if we want to add a 2nd optional field later we will have to do the same kind of tricks as for `liquidityPurchase/localCommit` to handle cases where there the middle object is not defined (instead of simply reading objects one after the other).

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/561/commits/f735f45eee0b765ed04563ecfad0f5a0b061a774

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Serialization.kt
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/561/commits/f735f45eee0b765ed04563ecfad0f5a0b061a774

src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
```diff
@@ -387,6 +388,21 @@ data class SpliceCpfpOutgoingPayment(
     override val completedAt: Long? = confirmedAt
 }
 
+data class InboundLiquidityOutgoingPayment(
+    override val id: UUID,
+    override val channelId: ByteVector32,
```
pm47, line 393: The mining fees should be provided explicitly and set the same way as for `SpliceOutgoingPayment`, currently we only count the mining fees for remote funding.

```suggestion
    override val miningFees: Satoshi,
    override val channelId: ByteVector32,
```

pm47: Proposed fix in #569.

t-bast: Good catch!

There is something I'm not 100% sure about. For every other splice transaction, the wallet user pays for *all* the mining fees. But for liquidity ads, this may be different: what the user pays in mining fees may be less than the actual transaction's mining fees, or more, depending on how many inputs/outputs are added by the LSP and what was set in the lease's `funding_weight`. Should the `miningFees` field match the transaction's mining fees, or should it match what was paid by the user? I chose the second option in https://github.com/ACINQ/lightning-kmp/pull/561/commits/df15490d105fc02d4e2fc9fa3ace49cfa409469e, while your proposal chose the first one.

t-bast: > I agree with option 2, however I would have defined miningFees as a normal attribute and set them to miningFees = action.fundingTx.sharedTx.tx.localFees.truncateToSatoshi().

If we do that, we should set `miningFees = action.fundingTx.sharedTx.tx.localFees.truncateToSatoshi() + lease.miningFee`, right? We want that field to include all fees that are for on-chain stuff, don't we? The goal is to cleanly separate them from the service fees?

t-bast: I think https://github.com/ACINQ/lightning-kmp/pull/561/commits/b77df8cfcc9a7446eb0c203ceffef622c9648c1f should remove the confusion, and fix other types of splices (even though in their case, `localFees` currently equals `fees`, but it's more future-proof to only use `localFees` in case we later include unrelated inputs/outputs on the LSP side).

dpad85 (approved):
LGTM 🥳 Successfully tested on Phoenix.

PR #563, opened 2023-12-05
Improve swap-in protocol with taproot and musig2

sstone:
This PR implements an evolution of our p2swh-based swap-in that uses Musig2 and taproot for better privacy and lower on-chain fees.

Our swap-in protocol becomes `user key + server key OR user refund key + delay`, and is implemented with taproot which provides a clean separation between the mutual use case (handled by the internal key, which is a Musig2 aggregation of the user's and server's key) and the refund case (handled by a taproot script with a single leaf: the refund script).

Musig2 requires both parties to generate secret nonces and exchange public nonces before they can compute partial Musig2 signatures, which can then be combined into a single Schnorr signature:

- A new optional TLV is added to `TxComplete`, which includes public nonces for all swap-in inputs
- A new optional TLV is added to `TxSignatures`, which includes partial Musig2 signatures for all swap-in inputs

Nonces are never persisted and their lifecycle is tied to the interactive tx session lifecycle: while the session is alive, it remembers nonces that it sent before i.e if `TxComplete` is sent several times, nonces for the same input id will be re-used. I believe that it is safe since they will be used only once for signing.

A specific user refund key is used in the refund script: this allows us to easily rotate swap-in addresses (each time we change th user refund key) and generate a single taproot descriptor that will allow bitcoin core to find and spend (after the refund delay) all swap-in transactions for a given user (i.e for a given phoenix wallet). 

This is a first step towards fixing https://github.com/ACINQ/phoenix/issues/403: this new scheme makes swap-in address rotation easy to implement but this is out of the scope of this PR.

Swap-in addresses are now regular `p2tr` addresses, and swap-in transactions become indistinguishable from other `p2tr` transactions.
And since spending swap-in inputs requires a single 64 bytes Schnorr transaction (instead of 2 73 bytes DER transaction + a 77 bytes redeem script), the weight of a swap-in input is reduced by 40% for cheaper on-chain fees.

This PR is built on top of https://github.com/ACINQ/lightning-kmp/pull/565 because of https://github.com/ACINQ/secp256k1-kmp/pull/93 but if needed changes to `secp256k1-kmp` and `bitcoin-kmp` could be back-ported to work with kotlin 1.8.


src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -341,23 +396,80 @@ data class SharedTransaction(
...
     }
 
-    fun sign(keyManager: KeyManager, fundingParams: InteractiveTxParams, localParams: LocalParams, remoteNodeId: PublicKey): PartiallySignedSharedTransaction {
+    fun sign(session: InteractiveTxSession, keyManager: KeyManager, fundingParams: InteractiveTxParams, localParams: LocalParams, remoteNodeId: PublicKey): PartiallySignedSharedTransaction {
         val unsignedTx = buildUnsignedTx()
         val sharedSig = fundingParams.sharedInput?.sign(keyManager.channelKeys(localParams.fundingKeyPath), unsignedTx)
+        val sharedOutput = fundingParams.sharedInput?.let { i -> mapOf(i.info.outPoint to i.info.txOut) } ?: mapOf()
+        val localOutputs = localInputs.associate { i -> i.outPoint to i.txOut }
+        val remoteOutputs = remoteInputs.associate { i -> i.outPoint to i.txOut }
+        val previousOutputsMap = sharedOutput + localOutputs + remoteOutputs
```
t-bast, line 405: Why are you re-creating this? This is just the `spentOutputs` field you added above, you can directly use it?

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -341,23 +396,80 @@ data class SharedTransaction(
...
+        val previousOutputs = unsignedTx.txIn.map { previousOutputsMap[it.outPoint]!! }.toList()
```
t-bast, line 406: As discussed in the PR for the eclair implementation, it's worth adding a comment to explain that this list must be correctly ordered to match the transaction inputs, otherwise signing will fail.

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -94,44 +97,62 @@ sealed class InteractiveTxInput {
...
+        override val serialId: Long,
+        override val outPoint: OutPoint,
+        override val txOut: TxOut,
+        override val sequence: UInt,
+        val swapInParams: TxAddInputTlv.SwapInParams
+    ) : Remote()
 
     /** The shared input can be added by us or by our peer, depending on who initiated the protocol. */
-    data class Shared(override val serialId: Long, override val outPoint: OutPoint, override val sequence: UInt, val localAmount: MilliSatoshi, val remoteAmount: MilliSatoshi) : InteractiveTxInput(), Incoming, Outgoing
+    data class Shared(override val serialId: Long, override val outPoint: OutPoint, override val txOut: TxOut, override val sequence: UInt, val localAmount: MilliSatoshi, val remoteAmount: MilliSatoshi) : InteractiveTxInput(), Incoming,
```
t-bast, line 154: We don't need to add the whole `txOut` field, just its `publicKeyScript`, like we did for `eclair`.

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -341,23 +396,80 @@ data class SharedTransaction(
...
         val unsignedTx = buildUnsignedTx()
         val sharedSig = fundingParams.sharedInput?.sign(keyManager.channelKeys(localParams.fundingKeyPath), unsignedTx)
+        val sharedOutput = fundingParams.sharedInput?.let { i -> mapOf(i.info.outPoint to i.info.txOut) } ?: mapOf()
+        val localOutputs = localInputs.associate { i -> i.outPoint to i.txOut }
+        val remoteOutputs = remoteInputs.associate { i -> i.outPoint to i.txOut }
+        val previousOutputsMap = sharedOutput + localOutputs + remoteOutputs
+        val previousOutputs = unsignedTx.txIn.map { previousOutputsMap[it.outPoint]!! }.toList()
+
+        // nonces that we've received for all musig2 swap-in
+        val receivedNonces: Map<Long, IndividualNonce> = when (session.txCompleteReceived) {
```
t-bast, line 409: This can be simplified to use the same pattern as eclair:

- validate (once) that `tx_complete` contains the right number of nonces
- zip them with our sorted musig2 remote inputs (without redundant validation)

See the corresponding comment on the eclair PR.

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -405,18 +516,43 @@ data class FullySignedSharedTransaction(override val tx: SharedTransaction, over
...
+            val (userSig, serverSig) = sigs
+            val swapInProtocol = SwapInProtocolLegacy(i.userKey, i.serverKey, i.refundDelay)
+            val witness = swapInProtocol.witness(userSig, serverSig)
+            Pair(i.serialId, TxIn(OutPoint(i.previousTx, i.previousTxOutput), ByteVector.empty, i.sequence.toLong(), witness))
+        }
+        val localSwapTxInMusig2 = tx.localInputs.filterIsInstance<InteractiveTxInput.LocalSwapIn>().sortedBy { i -> i.serialId }.zip(localSigs.swapInUserPartialSigs.zip(remoteSigs.swapInServerPartialSigs)).map { (i, sigs) ->
             val (userSig, serverSig) = sigs
-            val witness = Scripts.witnessSwapIn2of2(userSig, i.userKey, serverSig, i.serverKey, i.refundDelay)
+            val swapInProtocol = SwapInProtocol(i.swapInParams)
+            require(userSig.aggregatedPublicNonce == serverSig.aggregatedPublicNonce) { "aggregated public nonces mismatch for local input ${i.serialId}" }
```
t-bast, line 528: In which cases do you expect that to happen? A `require` will make the app crash, while we verify below that signatures are correct and cleanly fail the session if signatures don't check out.

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -405,18 +516,43 @@ data class FullySignedSharedTransaction(override val tx: SharedTransaction, over
...
+            val (userSig, serverSig) = sigs
+            val swapInProtocol = SwapInProtocolLegacy(i.userKey, i.serverKey, i.refundDelay)
+            val witness = swapInProtocol.witness(userSig, serverSig)
+            Pair(i.serialId, TxIn(i.outPoint, ByteVector.empty, i.sequence.toLong(), witness))
+        }
+        val remoteSwapTxInMusig2 = tx.remoteInputs.filterIsInstance<InteractiveTxInput.RemoteSwapIn>().sortedBy { i -> i.serialId }.zip(remoteSigs.swapInUserPartialSigs.zip(localSigs.swapInServerPartialSigs)).map { (i, sigs) ->
             val (userSig, serverSig) = sigs
-            val witness = Scripts.witnessSwapIn2of2(userSig, i.userKey, serverSig, i.serverKey, i.refundDelay)
+            val swapInProtocol = SwapInProtocol(i.swapInParams)
+            require(userSig.aggregatedPublicNonce == serverSig.aggregatedPublicNonce) { "aggregated public nonces mismatch for remote input ${i.serialId}" }
```
t-bast, line 547: Same here.

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -503,31 +641,52 @@ data class InteractiveTxSession(
...
 
-    val isComplete: Boolean = txCompleteSent && txCompleteReceived
+    val isComplete: Boolean = txCompleteSent != null && txCompleteReceived != null
 
     fun send(): Pair<InteractiveTxSession, InteractiveTxSessionAction> {
         return when (val msg = toSend.firstOrNull()) {
             null -> {
-                val txComplete = TxComplete(fundingParams.channelId)
-                val next = copy(txCompleteSent = true)
+                // generate a new secret nonce for each musig2 new swapin every time we send TxComplete
```
t-bast, line 649: Why do we do that? It is much simpler to generate one secret nonce whenever we send `tx_add_input`, and here to just send the public part, isn't it?

It would also be more consistent since that's what we do for remote inputs: we generate our secret nonce as soon as we receive their swap-in input.

Then you can trivially filter the local secret nonces from the remote secret nonces by just looking at the parity of the `serial_id`.

RECOVERY.md
```diff
@@ -19,24 +19,29 @@ For example, when using [electrum](https://electrum.org/):
...
 This process will become simpler once popular on-chain wallets (such as [electrum](https://electrum.org/)) add supports for output script descriptors.
 
-### Extract master keys
+### Get your wallet descriptor
 
-We don't directly export your extended master private key for security reasons, so you will need to manually insert it in the descriptor.
-You can obtain your extended master private key in [electrum](https://electrum.org/). After restoring your seed, type `wallet.keystore.xprv` in the console to obtain your master `xprv`.
+lighting-kmp provides both a public descriptor and private descriptor for your swap-in wallet.
+The public descriptor can be used to create a watch-only wallet for your swap-in funds.
+The private descriptor can be used to recover your swap-in funds, after the refund delay has passed.
```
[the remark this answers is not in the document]
sstone: With the public descriptors, users can track their on-chain funds but bot spend them. With the private descriptors bitcoin core will also be able to compute the correct private key.
I don't think it is a security issue as this descriptor can only generate swap-in refunds keys (i.e it cannot be used to go up the BIP32 derivation chain)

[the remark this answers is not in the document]
sstone: AFAICT if you want bitcoin core to generate private keys you must import a private descriptors (importing a public key descriptor into a wallet with private keys enable will fail with ` Cannot import descriptor without private keys to a wallet with private keys enabled`)

RECOVERY.md
```diff
@@ -19,24 +19,29 @@ For example, when using [electrum](https://electrum.org/):
...
 - we wait for that transaction to have enough confirmations
 - then, if the fees don't exceed the user's liquidity policy, these funds are moved into a lightning channel
 
+We use musig2 to aggregate user keys (user being the wallet) and server keys (server being the LSP: the ACINQ node): swap-in addresses are standard p2tr addresses, and
+swap-in transactions to your wallet are indistinguishable from other p2tr transactions.
+
 The swap transaction's output can be spent using either:
 
-1. A signature from the user's wallet and a signature from the remote node
+1. A aggregated musig2 signature built from a partial signature from the user's wallet and a partial signature from the remote node
```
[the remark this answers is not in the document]
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/563/commits/4237120872025646386ada61fe5d2f7f9d359538

RECOVERY.md
```diff
@@ -19,24 +19,29 @@ For example, when using [electrum](https://electrum.org/):
...
+This process needs at least Bitcoin Core 26.0.
 
 This process will become simpler once popular on-chain wallets (such as [electrum](https://electrum.org/)) add supports for output script descriptors.
 
-### Extract master keys
+### Get your wallet descriptor
 
-We don't directly export your extended master private key for security reasons, so you will need to manually insert it in the descriptor.
-You can obtain your extended master private key in [electrum](https://electrum.org/). After restoring your seed, type `wallet.keystore.xprv` in the console to obtain your master `xprv`.
+lighting-kmp provides both a public descriptor and private descriptor for your swap-in wallet.
```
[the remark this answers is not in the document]
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/563/commits/4237120872025646386ada61fe5d2f7f9d359538

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/563/commits/4237120872025646386ada61fe5d2f7f9d359538

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/563/commits/4237120872025646386ada61fe5d2f7f9d359538 (it was wrong to throw an exception here and nonce mismatch is fairly unlikely to happen)

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/563/commits/4237120872025646386ada61fe5d2f7f9d359538

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/563/commits/4237120872025646386ada61fe5d2f7f9d359538

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
t-bast: I see what you mean, it becomes messy to untangle the local and remote swap inputs, this map is the simplest way of writing that code, let's forget about this.

t-bast (commented):
I've opened #592 to leverage the musig2 helpers and fix my remaining open comments. Once we converge on it and merge it into this PR, I think that the code will be in pretty good shape, the only thing that will remain will be to write unit tests that exercise the musig2 swaps and fix the `SwapInManager` tests!

t-bast (commented):
LGTM, we now just need to add tests for the interactive-tx signing part using musig2 and fix the swap-in manager tests, and I think we'll be ready :rocket:

sstone:
The swap-in manager tests were failing when we experimented with not sending the full tx (I was creating wallet UTXOs manually), they've been restored now.

t-bast:
Right, that was why! That makes sense, then we just need to add tests to `InteractiveTxTestsCommon` to use the musig2 path and we should have everything.

src/commonTest/kotlin/fr/acinq/lightning/channel/InteractiveTxTestsCommon.kt
```diff
@@ -846,39 +851,31 @@ class InteractiveTxTestsCommon : LightningTestSuite() {
     }
 
     @Test
-    fun `swap-in input missing user key`() {
-        val f = createFixture(100_000.sat, listOf(150_000.sat), 0.sat, listOf(), FeeratePerKw(2500.sat), 330.sat, 0)
-        val alice0 = InteractiveTxSession(f.channelKeysA, f.keyManagerA.swapInOnChainWallet, f.fundingParamsA, 0.msat, 0.msat, f.fundingContributionsA)
-        val bob0 = InteractiveTxSession(f.channelKeysB, f.keyManagerB.swapInOnChainWallet, f.fundingParamsB, 0.msat, 0.msat, f.fundingContributionsB)
+    fun `swap-in input missing user nonce`() {
```
t-bast, line 854: This is a nice test, but it shouldn't replace the previous test, it should just be a new one? We're still supporting legacy swap-ins, so we want to keep the `swap-in input missing user key` test.

src/commonTest/kotlin/fr/acinq/lightning/channel/InteractiveTxTestsCommon.kt
```diff
@@ -1213,10 +1215,15 @@ class InteractiveTxTestsCommon : LightningTestSuite() {
             return action1
         }
 
-        private fun createWallet(onChainKeys: KeyManager.SwapInOnChainKeys, amounts: List<Satoshi>): List<WalletState.Utxo> {
+        private fun createWallet(onChainKeys: KeyManager.SwapInOnChainKeys, amounts: List<Satoshi>, legacyAmounts: List<Satoshi> = listOf()): List<WalletState.Utxo> {
```
t-bast, line 1218: You're introducing these two new parameters but you don't even use them once...you should modify at least one of the existing tests to provide both taproot inputs and legacy inputs, otherwise we don't have any test coverage of the legacy swap protocol anymore.

src/commonTest/kotlin/fr/acinq/lightning/channel/InteractiveTxTestsCommon.kt
```diff
@@ -591,13 +592,18 @@ class InteractiveTxTestsCommon : LightningTestSuite() {
         assertTrue(targetFeerate <= feerate && feerate <= targetFeerate * 1.25, "unexpected feerate (target=$targetFeerate actual=$feerate)")
     }
 
+
```
t-bast, line 595: nit: unnecessary new line

src/commonTest/kotlin/fr/acinq/lightning/channel/InteractiveTxTestsCommon.kt
sstone: Yes I'm adding tests with mixed legacy/new swap-in outputs.

sstone: done in https://github.com/ACINQ/lightning-kmp/pull/563/commits/02d89fb252f52edd27bbbcc42292440b503214b7

t-bast: It still looks like you deleted the previous test? I think it should be restored, it is useful to test the scenario where one side "forgets" to send the right TLV to advertise that an input is using swap-in-potentiam.

sstone: Sorry I missed this, restored in https://github.com/ACINQ/lightning-kmp/pull/563/commits/962efbb24c9e739750ac5b31d31f642cd454abc9

t-bast (commented):
ACK https://github.com/ACINQ/lightning-kmp/pull/563/commits/962efbb24c9e739750ac5b31d31f642cd454abc9

RECOVERY.md
```diff
@@ -19,24 +19,29 @@ For example, when using [electrum](https://electrum.org/):
...
 2. A signature from the user's wallet after a refund delay
 
 Funds can be recovered using the second option and [Bitcoin Core](https://github.com/bitcoin/bitcoin).
-This process needs at least Bitcoin Core 25.0.
+This process needs at least Bitcoin Core 26.0.
 
 This process will become simpler once popular on-chain wallets (such as [electrum](https://electrum.org/)) add supports for output script descriptors.
 
-### Extract master keys
+### Get your wallet descriptor
```
pm47, line 39: We need to keep documenting the legacy p2wsh case for the foreseeable future. I'm unclear on the differences between the p2wsh/p2tr cases and what can be factored. Also the address rotation adds another layer of complexity and makes factoring the doc even more necessary. Is the difference just around descriptors?

sstone: I added a link to the recovery documentation for legacy swap-ins. 

The new recovery process is simpler, and just requires to import a single descriptor into bitcoin core (26 or later). This process will remain the same once address rotation has been added (the current descriptor will match all generated addresses).

The biggest difference is that the legacy descriptor's path included an xpub with a path that started at the master key, which required an intermediary step to get the xpriv first with Electrum. Here the descriptor includes an xpriv for the swap-in refund keys (it cannot be used to go "up").

src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
```diff
@@ -118,28 +122,30 @@ interface KeyManager {
...
+        val userRefundPrivateKey: PrivateKey = DeterministicWallet.derivePrivateKey(userRefundExtendedPrivateKey, 0).privateKey
+        val userRefundPublicKey: PublicKey = userRefundPrivateKey.publicKey()
+
         private val localServerExtendedPrivateKey: DeterministicWallet.ExtendedPrivateKey = DeterministicWallet.derivePrivateKey(master, swapInLocalServerKeyPath(chain))
         fun localServerPrivateKey(remoteNodeId: PublicKey): PrivateKey = DeterministicWallet.derivePrivateKey(localServerExtendedPrivateKey, perUserPath(remoteNodeId)).privateKey
 
-        val redeemScript: List<ScriptElt> = Scripts.swapIn2of2(userPublicKey, remoteServerPublicKey, refundDelay)
-        val pubkeyScript: List<ScriptElt> = Script.pay2wsh(redeemScript)
-        val address: String = Bitcoin.addressFromPublicKeyScript(chain.chainHash, pubkeyScript).result!!
+        val swapInProtocol = SwapInProtocol(userPublicKey, remoteServerPublicKey, userRefundPublicKey, refundDelay)
```
pm47, line 136: `SwapInOnChainKeys` has a `chain` attribute, but when we do `keyManager.swapInProtocol` we get an object that is chain-agnostic, and need to provide the chain to `swapInProtocol.address()` and `swapInProtocol.descriptor()`. Why not add `chain` to `SwapInProtocol`?

sstone: I've reworked the generation of descriptors and our recovery procedure (which had problems, see the commit message). There are now 2 different methods, to generate either a private descriptor or a public descriptor. Both are valid, compatible with refund key rotation, and can be imported directly into bitcoin core.

For the recovery process, the user's seed is not enough (it's true for the current swaps too): users also need their swap-in public key, the server's swap-in public key, and the refund delay.

sstone: see https://github.com/ACINQ/lightning-kmp/pull/563/commits/689949a9a2c6bf9e86de9d2db34f1c686a85e66a for more info

t-bast: An example of that is that the `compute descriptors to recover swap-in funds` test isn't a very good test: it uses custom code in the test itself instead of testing the actual code paths used by a wallet user. It would be better to use the seeds for the standard test users (`TestConstants.Alice` and `TestConstants.Bob`) and the corresponding `keyManager` instead to derive the descriptors?

sstone: > I don't understand why you moved the `descriptor` functions to the `companion object`: you need to provide the same parameters as the `(Legacy)SwapInProtocol` instance, so they should just be methods? Otherwise you're risking that we mess up the parameters and don't provide the right ones (or at least it's just annoying because we have to provide a list of parameters we already provided when creating the `(Legacy)SwapInProtocol` instances)...

in the legacy case `SwapInProtocolLegacy` has all that is needed to generate a descriptor. 
But for the new swap-ins, you need to provide the refund master xpriv, and the descriptor will be valid only if refund keys passed to create instances of `SwapInProtocol` are derived directly from this xpriv, so I think that having a static method that takes all required parameters is a bit cleaner.

t-bast: I'm not really convinced to be honest, making it a method lets us provide less arguments and makes it easy to check correctness of the arguments provided inside the function with a `require`, but that's not very important, you can ignore that.

RECOVERY.md
```diff
@@ -19,89 +21,99 @@ For example, when using [electrum](https://electrum.org/):
...
 
-1. A signature from the user's wallet and a signature from the remote node
+1. An aggregated musig2 signature built from a partial signature from the user's wallet and a partial signature from the remote node
 2. A signature from the user's wallet after a refund delay
 
 Funds can be recovered using the second option and [Bitcoin Core](https://github.com/bitcoin/bitcoin).
-This process needs at least Bitcoin Core 25.0.
+This process needs at least Bitcoin Core 26.0.
 
 This process will become simpler once popular on-chain wallets (such as [electrum](https://electrum.org/)) add supports for output script descriptors.
```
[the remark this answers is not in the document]
sstone: done in https://github.com/ACINQ/lightning-kmp/pull/563/commits/be62fc5dfb5ab4378d949cbe7f0436b4b8762a29

RECOVERY.md
```diff
@@ -19,89 +21,99 @@ For example, when using [electrum](https://electrum.org/):
...
 ### Create recovery wallet
 
-Create a wallet to recover your funds using the following command:
+#### Compute your refund master private key
 
-```sh
-bitcoin-cli createwallet recovery
-```
+Use the [mnemonic code converter](https://iancoleman.io/bip39/) to compute your refund master key from your seed word (we
+strongly recommend that you use this tool in "offline" mode, see the "Offline usage section for more information)"
```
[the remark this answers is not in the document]
sstone: It turns out that we can use electrum (thanks @dpad85)  with a simple change from the original one, I've updated the recovery procedure.

RECOVERY.md
```diff
@@ -159,30 +171,23 @@ bitcoin-cli -rpcwallet=recovery listtransactions
 Once those funds have been recovered and the refund delay has expired (the `confirmations` field of the previous command exceeds `25920`), you can send them to your normal on-chain wallet.
 Compute the total amount received (in our example, 1.5 BTC), choose the address to send to (for example, `bcrt1q9ez7rt33wynwpah582lnqlj3u0tpzsrkj2flas`) and create a transaction using all of the received funds:
 
-```sh
-bitcoin-cli -rpcwallet=recovery walletcreatefundedpsbt '[{"txid":"d9940b7eb709ff8eaec307bdd6d20633e30a6eb1627d9ef8c8e03dfd28298c75","vout":1,"sequence":25920}]' '[{"bcrt1q9ez7rt33wynwpah582lnqlj3u0tpzsrkj2flas":1.5}]' 0 '{"subtractFeeFromOutputs":[0]}'
-
+```shell
+bitcoin-cli -rpcwallet=recovery walletcreatefundedpsbt '[{"txid":"4c3236b1fa1f3ed124ab83b1667be95f855952e68729eae54a9f511c8c8cb993", "vout":0, "sequence":144}]' '[{"bcrt1qzy4h8dux6pjl8ys979632uynqffd53vjkzffjl":0.09}]'
```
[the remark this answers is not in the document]
sstone: Yes, a blank wallet will not work (for now), it must also contain regular descriptors to generate changed addresses if needed.
Unfortunately we cannot yet use `sendtoaddress` (Bitcoin Core won't be able to sign) and have to create and sign PSBTs manually, but this will probably be simpler in future versions of Bitcoin Core.

RECOVERY.md
```diff
@@ -19,77 +21,90 @@ For example, when using [electrum](https://electrum.org/):
...
 
 ### Create recovery wallet
 
-Create a wallet to recover your funds using the following command:
+#### Compute your refund master private key
 
-```sh
-bitcoin-cli createwallet recovery
-```
+For security reasons, we don't directly the refund master private key used for swap-ins, so you will need to manually insert it in the descriptor.
```
t-bast, line 45:
```suggestion
For security reasons, we don't directly export the refund master private key used for swap-ins, so you will need to manually insert it in the descriptor.
```

RECOVERY.md
```diff
@@ -19,77 +21,90 @@ For example, when using [electrum](https://electrum.org/):
...
-
+For example:
 ```txt
-wsh(and_v(v:pk([14620948/51h/0h/0h]tpubDCvYeHUZisCMV3h1zPevPWQmNPfA3g3vnu7gDqskXVCbJB1VKk2F7LApV6TTdm1sCyGout8ma27CCHvYTuMZxpwrcHnLwL4kaXW8z2KfFcW),or_d(pk(0256e948180f33f067246710a41656084fc245b97eda081efe1e488b21577d60fd),older(25920))))
+tr(1fc559d9c96c5953895d3150e64ebf3dd696a0b08e758650b48ff6251d7e60d1,and_v(v:pk(xpub6EE2N7jrues5kfjrsyFA5f7hknixqqAEKs8vyMN4QW9vDmYnChzpeBPkBYduBobbe4miQ34xHG4Jpwuq5bHXLZY1xixoGynW31ySUqqVvcU/*),older(2590)))#sv8ug44m
 ```
 
-Replace the `extended_public_key` and the `derivation_path` with the extended private key obtained in the [first step](#extract-master-keys).
-In our example, the extended private key matching our seed is `tprv8ZgxMBicQKsPdKRFLVct6VDpfmCxk6aC7iAF8tb6roQ7hv1zFCyGwDLBUUxMVJ95dTiQS5VvCbQ6J7CcGqguw5SbnDpNjbjpfVwcMwUtmjS`, so we create the following private descriptor:
+You can check that the extended public key in this descriptors matches the extended public key of the wallet your created with Electrum to compute your refund master private key.
```
t-bast, line 67:
```suggestion
You can check that the extended public key in this descriptor matches the extended public key of the wallet your created with Electrum to compute your refund master private key.
```

RECOVERY.md
```diff
@@ -19,77 +21,90 @@ For example, when using [electrum](https://electrum.org/):
...
-  "hasprivatekeys": true
-}
+```shell
+bitcoin-cli -named createwallet wallet_name=recovery
 ```
 
-We can the append this checksum to our private descriptor and import it into our recovery wallet:
+### Import your descriptor into the recovery wallet
+
+We can import our private descriptor into our recovery wallet. Since you replaced you refund master public key with your refund master private key, the descriptor checksum is no long valid, but bitcoin core will give you the correct checksum:
```
t-bast, line 84:
```suggestion
We can import our private descriptor into our recovery wallet. Since you replaced your refund master public key with your refund master private key, the descriptor checksum is no longer valid, but bitcoin core will give you the correct checksum:
```

RECOVERY.md
```diff
@@ -19,77 +21,90 @@ For example, when using [electrum](https://electrum.org/):
...
+
+```shell
+bitcoin-cli -rpcwallet=recovery1 importdescriptors '[{ "desc": "tr(1fc559d9c96c5953895d3150e64ebf3dd696a0b08e758650b48ff6251d7e60d1,and_v(v:pk(xprvA1EfxcCy5HJnYBfPmwi9iXAyCktUSNSNxeDLAxxSrAcwLyDdfAga6P5GLHNdq7EiXe8Pzu6Py6xGwT7UTkw824FYf3v6fbRStvYsWqFTu29/*),older(2590)))#sv8ug44m", "timestamp":0}]'
```
t-bast, line 87:
```suggestion
bitcoin-cli -rpcwallet=recovery importdescriptors '[{ "desc": "tr(1fc559d9c96c5953895d3150e64ebf3dd696a0b08e758650b48ff6251d7e60d1,and_v(v:pk(xprvA1EfxcCy5HJnYBfPmwi9iXAyCktUSNSNxeDLAxxSrAcwLyDdfAga6P5GLHNdq7EiXe8Pzu6Py6xGwT7UTkw824FYf3v6fbRStvYsWqFTu29/*),older(2590)))#sv8ug44m", "timestamp":0}]'
```

RECOVERY.md
```diff
@@ -19,77 +21,90 @@ For example, when using [electrum](https://electrum.org/):
...
+    }
+  }
+]
+```
 
-```sh
-bitcoin-cli -rpcwallet=recovery importdescriptors '[{ "desc": "wsh(and_v(v:pk(tprv8ZgxMBicQKsPdKRFLVct6VDpfmCxk6aC7iAF8tb6roQ7hv1zFCyGwDLBUUxMVJ95dTiQS5VvCbQ6J7CcGqguw5SbnDpNjbjpfVwcMwUtmjS/51h/0h/0h),or_d(pk(0256e948180f33f067246710a41656084fc245b97eda081efe1e488b21577d60fd),older(25920))))#dlcgkrnc", "timestamp": 0 }]'
+Update the checksum and try again:
+```shell
+bitcoin-cli -rpcwallet=recovery1 importdescriptors '[{ "desc": "tr(1fc559d9c96c5953895d3150e64ebf3dd696a0b08e758650b48ff6251d7e60d1,and_v(v:pk(xprvA1EfxcCy5HJnYBfPmwi9iXAyCktUSNSNxeDLAxxSrAcwLyDdfAga6P5GLHNdq7EiXe8Pzu6Py6xGwT7UTkw824FYf3v6fbRStvYsWqFTu29/*),older(2590)))#ksphr9r4", "timestamp":0}]'
```
t-bast, line 101:
```suggestion
bitcoin-cli -rpcwallet=recovery importdescriptors '[{ "desc": "tr(1fc559d9c96c5953895d3150e64ebf3dd696a0b08e758650b48ff6251d7e60d1,and_v(v:pk(xprvA1EfxcCy5HJnYBfPmwi9iXAyCktUSNSNxeDLAxxSrAcwLyDdfAga6P5GLHNdq7EiXe8Pzu6Py6xGwT7UTkw824FYf3v6fbRStvYsWqFTu29/*),older(2590)))#ksphr9r4", "timestamp":0}]'
```

RECOVERY.md
```diff
@@ -119,70 +134,70 @@ bitcoin-cli -rpcwallet=recovery getwalletinfo
...
 
 ### Send funds to a different address
 
 Once those funds have been recovered and the refund delay has expired (the `confirmations` field of the previous command exceeds `25920`), you can send them to your normal on-chain wallet.
-Compute the total amount received (in our example, 1.5 BTC), choose the address to send to (for example, `bcrt1q9ez7rt33wynwpah582lnqlj3u0tpzsrkj2flas`) and create a transaction using all of the received funds:
+For now, this process involves selecting the inputs that you want to spend and creating the spending transaction manually, as documented below, but future versions of Bitcoin Core will probably make this easier.
 
-```sh
-bitcoin-cli -rpcwallet=recovery walletcreatefundedpsbt '[{"txid":"d9940b7eb709ff8eaec307bdd6d20633e30a6eb1627d9ef8c8e03dfd28298c75","vout":1,"sequence":25920}]' '[{"bcrt1q9ez7rt33wynwpah582lnqlj3u0tpzsrkj2flas":1.5}]' 0 '{"subtractFeeFromOutputs":[0]}'
+For example, if `listtransactions` lists a UTXO `5e9d2a387572fe0c8a4996c2f34373b3fbbdb19ff106b84fc91c2450eb27cbe7:0` of `0.002` Bitcoin, this is how you would send it to you another on-chain address.
```
t-bast, line 179:
```suggestion
For example, if `listtransactions` lists a UTXO `5e9d2a387572fe0c8a4996c2f34373b3fbbdb19ff106b84fc91c2450eb27cbe7:0` of `0.002` Bitcoin, this is how you would send it to another on-chain address.
```

t-bast (commented):
A lot of typos in that recovery file :sweat_smile:, but apart from that it's looking all good, I tested all those steps and everything worked fine!

RECOVERY.md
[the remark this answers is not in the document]
t-bast: I indeed tested that `sendtoaddress` doesn't work (fails at signing time), but the PSBT procedure works when not using a blank wallet :+1:

RECOVERY.md
sstone: fixed in https://github.com/ACINQ/lightning-kmp/pull/563/commits/5e18d0b06f2f6439f8ba2d3c52398f340bcff569

RECOVERY.md
sstone: fixed in https://github.com/ACINQ/lightning-kmp/pull/563/commits/5e18d0b06f2f6439f8ba2d3c52398f340bcff569

RECOVERY.md
sstone: fixed in https://github.com/ACINQ/lightning-kmp/pull/563/commits/5e18d0b06f2f6439f8ba2d3c52398f340bcff569

RECOVERY.md
sstone: fixed in https://github.com/ACINQ/lightning-kmp/pull/563/commits/5e18d0b06f2f6439f8ba2d3c52398f340bcff569

RECOVERY.md
sstone: fixed in https://github.com/ACINQ/lightning-kmp/pull/563/commits/5e18d0b06f2f6439f8ba2d3c52398f340bcff569

RECOVERY.md
sstone: fixed in https://github.com/ACINQ/lightning-kmp/pull/563/commits/5e18d0b06f2f6439f8ba2d3c52398f340bcff569

RECOVERY.md
```diff
@@ -19,77 +21,90 @@ For example, when using [electrum](https://electrum.org/):
...
-  "hasprivatekeys": true
-}
+```shell
+bitcoin-cli -named createwallet wallet_name=recovery
 ```
 
-We can the append this checksum to our private descriptor and import it into our recovery wallet:
+### Import your descriptor into the recovery wallet
+
+We can import our private descriptor into our recovery wallet. Since your replaced you refund master public key with your refund master private key, the descriptor checksum is no longer valid, but bitcoin core will give you the correct checksum:
```
t-bast, line 84:
```suggestion
We can import our private descriptor into our recovery wallet. Since you replaced your refund master public key with your refund master private key, the descriptor checksum is no longer valid, but bitcoin core will give you the correct checksum:
```

sstone: d'oh! :( fixed in https://github.com/ACINQ/lightning-kmp/pull/563/commits/7d17663e63529fb1131044717d1f63d89fb00f5e

t-bast (commented):
LGTM!

build.gradle.kts
```diff
@@ -11,7 +11,7 @@ plugins {
 
 allprojects {
     group = "fr.acinq.lightning"
-    version = "1.6-SNAPSHOT"
+    version = "1.6-SWAPIN2-SNAPSHOT"
```
t-bast, line 14:
```suggestion
    version = "1.6-SNAPSHOT"
```

t-bast (approved):

pm47 (approved):

PR #564, opened 2023-12-08
Unlock utxos when peer is unavailable for splice

pm47:
If the peer is offline when the swap-in job requests a splice-in, the request will be ignored, but we weren't unlocking UTXOs, preventing any further attempt.

Also add an optimistic synchronization in the swap-in job, which as a side effect will cause us to retry swaps everytime the peer gets reconnected (vs at every new block or liquidity policy update previously).

cc @dpad85 @robbiehanson

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -453,7 +453,6 @@ class Peer(
                 .combine(currentTipFlow.filterNotNull()) { walletState, currentTip -> Pair(walletState, currentTip.first) }
                 .combine(swapInFeeratesFlow.filterNotNull()) { (walletState, currentTip), feerate -> Triple(walletState, currentTip, feerate) }
                 .combine(nodeParams.liquidityPolicy) { (walletState, currentTip, feerate), policy -> TrySwapInFlow(currentTip, walletState, feerate, policy) }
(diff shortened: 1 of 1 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -453,6 +453,7 @@ class Peer(
                 .combine(currentTipFlow.filterNotNull()) { walletState, currentTip -> Pair(walletState, currentTip.first) }
                 .combine(swapInFeeratesFlow.filterNotNull()) { (walletState, currentTip), feerate -> Triple(walletState, currentTip, feerate) }
                 .combine(nodeParams.liquidityPolicy) { (walletState, currentTip, feerate), policy -> TrySwapInFlow(currentTip, walletState, feerate, policy) }
+                .combine(channelsFlow.filter { it.values.all { channel -> channel !is Offline && channel !is Syncing } }) { req, _ -> req } // this is only for synchronization, we discard the channels data
```
pm47, line 456: NB: this is a simplification of `waitForPeerReady()`, which doesn't handles the "local syncing" corner case.

pm47: If you look at the handler for `RequestChannelOpen`, in the "local syncing" case requests always end up being ignored. It's probably a (rare) bug, but technically not a regression 😓

PR #565, opened 2023-12-08
Upgrade to Kotlin 1.9

dpad85:
This PR upgrades the project from Kotlin 1.8.21 to 1.9.21.

Summary of 1.9 changes:
- https://kotlinlang.org/docs/whatsnew19.html
- https://kotlinlang.org/docs/whatsnew1920.html

Notes:
- Objective-C or Swift object deallocation hook on the main thread﻿
@robbiehanson This is enabled by default, but can be opted-out if you think we should.
See [this entry for details](https://kotlinlang.org/docs/whatsnew19.html#objective-c-or-swift-object-deallocation-hook-on-the-main-thread).

- New custom memory allocator that improves performances
@t-bast @sstone Introduced in 1.9, and enabled by default with 1.9.20. There's an option to disable it if the memory cost is too high or if you think we should.
See [this entry for details](https://kotlinlang.org/docs/whatsnew1920.html#custom-memory-allocator-enabled-by-default)

- Several annotations added, e.g. `ExperimentalForeignApi`
See [this entry for details](https://kotlinlang.org/docs/whatsnew1920.html#opt-in-requirement-for-all-cinterop-declarations)

- Use `data object` instead of `object` for `sealed` classes
Not doing so leaves warnings in the code, and it seems to bring some convenience (better `toString()`, same `hashCode()`)
See [this entry for details](https://kotlinlang.org/docs/whatsnew19.html#stable-data-objects-for-symmetry-with-data-classes)

- Default hierarchy template is enabled with 1.9.20
See [this entry for details](https://kotlinlang.org/docs/whatsnew1920.html#set-up-the-target-hierarchy)
@sstone I updated the `build.gradle` file to use this template, which is cleaner. However there are some issues with the `PhoenixCrypto` cinterop compilation task that is launched even on the Ubuntu CI, even though it should not since it's declared in `ios` tasks. So [I had to keep the platform checks for Mac](https://github.com/ACINQ/lightning-kmp/pull/565/commits/c8d278a9fa293a92be4b079fba3e8f2bf78d9b17).


src/nativeInterop/cinterop/PhoenixCrypto.def
```diff
@@ -9,5 +9,5 @@ libraryPaths.ios_x64 = PhoenixCrypto/build/Release-iphonesimulator
 libraryPaths.ios_arm64 = PhoenixCrypto/build/Release-iphoneos
 
 linkerOpts = -L/usr/lib/swift
-linkerOpts.ios_x64 = -ios_simulator_version_min 14.0.0 -L/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/iphonesimulator/
+linkerOpts.ios_x64 = -L/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/iphonesimulator/
```
dpad85, line 12: This option caused the build to fail (`ld: unknown options: -ios_simulator_version_min`

The core issue seems to be that support for this option has been removed with Xcode 15. So I removed it, and the build passes.

However some people [suggest using `-platform_version`](https://developer.apple.com/forums/thread/733536). @robbiehanson wdyt?

dpad85: Eventually used `platform_version` with min 15.0.0 in https://github.com/ACINQ/lightning-kmp/pull/565/commits/8f8624d7c41c78a83806fcf127eec9d2a355efbd.

robbiehanson:
> New custom memory allocator that improves performances

The previous memory allocator (mimalloc) caused problems on iOS because:
> Apple limits the amount of memory that a notification-service-extension is allowed to use. The memory is capped at 24 MB (for iOS 16).

See [PR #324](https://github.com/ACINQ/phoenix/pull/324) for details. But this means that, currently, the standard allocator is being used via `-Xallocator=std`.

dpad85:
> See [PR #324](https://github.com/ACINQ/phoenix/pull/324) for details. But this means that, currently, the standard allocator is being used via `-Xallocator=std`.

Good catch, so we should not be affected on Phoenix for now. Though we may want to test this new allocator at some point, the performance improvements may be interesting for the background service extension provided we stay below this 24MB limit.

sstone (approved):

PR #568, opened 2023-12-11
Splice with htlcs

remyers:
This PR builds on splices #430 to account for pending htlcs that may exist if the channel has first negotiated quiescence #558. This PR is modeled after [Eclair PR # 2720](https://github.com/ACINQ/eclair/pull/2720).

The helper function `makeCommitTxsWithoutHtlcs` is renamed to `makeCommitTxs` and uses the set of committed htlcs that exist when `InteractiveTxSigningSession` is instantiated for a splice.

A new attribute `htlcsAmount` is added to `InteractiveTxOutput.Shared` to account for the value committed to htlcs from the shared funding input but not assigned to `localAmount` or `remoteAmount`.

This PR assumes the `quiescence` feature is mandatory rather than handle splices without exchanging `stfu`.

Updated tests for #421 with a pending splice with 5 instead of 6 htlcs maximum on each side.


src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
```diff
@@ -254,6 +254,7 @@ object Deserialization {
             pubkeyScript = readDelimitedByteArray().toByteVector(),
             localAmount = readNumber().msat,
             remoteAmount = readNumber().msat,
+            htlcAmount = readNumber().msat
```
t-bast, line 257: That's not backwards-compatible, is it? You must introduce a new discriminator (e.g. `0x02`) for which we encode the `htlcAmount`. When you encounter the `0x01` discriminator, you should just fill this with `0.msat`.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Serialization.kt
```diff
@@ -305,6 +305,7 @@ object Serialization {
         writeDelimited(pubkeyScript.toByteArray())
         writeNumber(localAmount.toLong())
         writeNumber(remoteAmount.toLong())
+        writeNumber(htlcAmount.toLong())
```
t-bast, line 308: Sane here, you must change the discriminator line 303 to `write(0x02)` since you introduce a new field.

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -8,6 +8,7 @@ import fr.acinq.lightning.MilliSatoshi
 import fr.acinq.lightning.blockchain.fee.FeeratePerByte
 import fr.acinq.lightning.blockchain.fee.FeeratePerKw
 import fr.acinq.lightning.blockchain.fee.FeerateTolerance
+import fr.acinq.lightning.channel.Commitments.Companion.makeLocalTxs
```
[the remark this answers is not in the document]
remyers: Fixed in [9aa3be5](https://github.com/ACINQ/lightning-kmp/pull/568/commits/9aa3be57e10fadd8fa0e3db1d8e21cdb0255d8e5).

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -464,62 +520,15 @@ data class Commitment(
...
-                    // which was created with SIGHASH_SINGLE || SIGHASH_ANYONECANPAY
-                    if (!Transactions.checkSig(htlcTx, remoteSig, remoteHtlcPubkey, SigHash.SIGHASH_SINGLE or SigHash.SIGHASH_ANYONECANPAY)) {
-                        return Either.Left(InvalidHtlcSignature(params.channelId, htlcTx.tx.txid))
-                    }
-                    HtlcTxAndSigs(htlcTx, localSig, remoteSig)
-                }
+        return LocalCommit.fromCommitSig(channelKeys, params, fundingTxIndex, remoteFundingPubkey, commitInput, commit, localCommit.index + 1, spec, localPerCommitmentPoint, log).map { localCommit1 ->
+            log.info {
+                val htlcsIn = spec.htlcs.outgoings().map { it.id }.joinToString(",")
+                val htlcsOut = spec.htlcs.incomings().map { it.id }.joinToString(",")
```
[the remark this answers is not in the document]
remyers: Doh! Fixed in [9aa3be5](https://github.com/ACINQ/lightning-kmp/pull/568/commits/9aa3be57e10fadd8fa0e3db1d8e21cdb0255d8e5).

src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -543,6 +552,7 @@ data class FullCommitment(
         params.channelFeatures.hasFeature(Feature.ZeroReserveChannels) -> 0.sat
         else -> (fundingAmount / 100).max(params.localParams.dustLimit)
     }
+    val capacity = commitInput.txOut.amount
```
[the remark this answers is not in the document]
remyers: Ah good catch. Fixed in [9aa3be5](https://github.com/ACINQ/lightning-kmp/pull/568/commits/9aa3be57e10fadd8fa0e3db1d8e21cdb0255d8e5).

src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
```diff
@@ -336,9 +337,9 @@ object Helpers {
                 commitmentInput,
                 remotePerCommitmentPoint = remotePerCommitmentPoint,
                 remoteSpec
-            ).first
+            )
 
-            return Either.Right(PairOfCommitTxs(localSpec, localCommitTx, remoteSpec, remoteCommitTx))
+            return Either.Right(PairOfCommitTxs(localSpec, localCommitTx, localHtlcTxs, remoteSpec, remoteCommitTx, remoteHtlcTxs))
```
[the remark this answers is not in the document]
remyers: Good idea; fixed in [9aa3be5](https://github.com/ACINQ/lightning-kmp/pull/568/commits/9aa3be57e10fadd8fa0e3db1d8e21cdb0255d8e5).

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -46,8 +48,9 @@ sealed class SharedFundingInput {
 }
 
 /** The current balances of a [[SharedFundingInput]]. */
-data class SharedFundingInputBalances(val toLocal: MilliSatoshi, val toRemote: MilliSatoshi) {
-    val fundingAmount: Satoshi = (toLocal + toRemote).truncateToSatoshi()
+data class SharedFundingInputBalances(val toLocal: MilliSatoshi, val toRemote: MilliSatoshi, val htlcs: Set<DirectedHtlc>) {
```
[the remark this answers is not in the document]
remyers: That makes sense. Fixed in [9aa3be5](https://github.com/ACINQ/lightning-kmp/pull/568/commits/9aa3be57e10fadd8fa0e3db1d8e21cdb0255d8e5)

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -232,7 +235,8 @@ data class FundingContributions(val inputs: List<InteractiveTxInput.Outgoing>, v
                 return Either.Left(FundingContributionFailure.InvalidFundingBalances(params.fundingAmount, nextLocalBalance, nextRemoteBalance))
             }
 
-            val sharedOutput = listOf(InteractiveTxOutput.Shared(0, fundingPubkeyScript, nextLocalBalance, nextRemoteBalance))
+            val nextHtlcBalance = (sharedUtxo?.second?.toHtlc ?: 0.msat)
+            val sharedOutput = listOf(InteractiveTxOutput.Shared(0, fundingPubkeyScript, nextLocalBalance, nextRemoteBalance, nextHtlcBalance))
```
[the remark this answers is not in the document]
remyers: Ah right, that is true. Fixed [9aa3be5](https://github.com/ACINQ/lightning-kmp/pull/568/commits/9aa3be57e10fadd8fa0e3db1d8e21cdb0255d8e5).

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -829,44 +837,54 @@ data class InteractiveTxSigningSession(
             localCommitmentIndex: Long,
             remoteCommitmentIndex: Long,
             commitTxFeerate: FeeratePerKw,
-            remotePerCommitmentPoint: PublicKey
+            remotePerCommitmentPoint: PublicKey,
+            previousHtlcs: Set<DirectedHtlc>
```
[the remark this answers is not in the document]
remyers: Also renamed in [9aa3be5](https://github.com/ACINQ/lightning-kmp/pull/568/commits/9aa3be57e10fadd8fa0e3db1d8e21cdb0255d8e5).

I don't understand what you mean about just being the local ones. It should be everything in the local node spec, right? It's currently initialized with `parentCommitment.localCommit.spec.htlcs`. This includes both incoming and outgoing htlcs and will be identical to the remote node spec but, as you said, reversed.

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -829,44 +837,54 @@ data class InteractiveTxSigningSession(
...
+                        CommitSigTlv.AlternativeFeerateSig(feerate, sig)
+                    }
+                    TlvStream(CommitSigTlv.AlternativeFeerateSigs(commitSigTlvs) as CommitSigTlv)
+                } else {
+                    TlvStream.empty()
                 }
-                val commitSig = CommitSig(channelParams.channelId, localSigOfRemoteTx, listOf(), TlvStream(CommitSigTlv.AlternativeFeerateSigs(alternativeSigs)))
-                val unsignedLocalCommit = UnsignedLocalCommit(localCommitmentIndex, firstCommitTx.localSpec, firstCommitTx.localCommitTx, listOf())
+                val commitSig = CommitSig(channelParams.channelId, localSigOfRemoteCommitTx, localSigsOfRemoteHtlcTxs, alternativeSigs)
+                val unsignedLocalCommit = UnsignedLocalCommit(localCommitmentIndex, firstCommitTx.localSpec, firstCommitTx.localCommitTx, sortedHtlcTxs)
```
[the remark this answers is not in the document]
remyers: I don't understand why `firstCommitTx.localHtlcTxs` should not be used here, except to reduce serialization size. This is the list of htlcs that are waiting for the `CommitSig` from our remote peer, but we should have all the information we need to create unsigned htlc txs, Right?

That said, because we recreate those htlc txs in `fromCommitSig` when we receive their `CommitSig` there's no need to set them here.

Which makes me think...  I couldn't find anywhere that `UnsignedLocalCommit.htlcTxs` is used. 

Do we even need the `htlcTxs` parameter at all now in `UnsignedLocalCommit`? I think it can be removed. Though that would require updating serialization and might not be worth the added discriminator logic.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
remyers: Fixed in [99ff0f6](https://github.com/ACINQ/lightning-kmp/pull/568/commits/99ff0f693fc569bb89008a97791f5340d6b8478d). But it makes me nervous that we don't have any serialization tests that would have caught this.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Serialization.kt
remyers: Also fixed in [99ff0f6](https://github.com/ACINQ/lightning-kmp/pull/568/commits/99ff0f693fc569bb89008a97791f5340d6b8478d).

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
t-bast: You should definitely add one! :)

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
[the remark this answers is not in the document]
t-bast: > This includes both incoming and outgoing htlcs and will be identical to the remote node spec but, as you said, reversed.

The htlcs from `parentCommitment.localCommit.spec` and `parentCommitment.remoteCommit.spec` are the same, but reversed. Since we create HTLC transactions based on that, for our _local_ commit, it's important to use the HTLCs from the local spec, otherwise we would generate (invalid) HTLC-success transactions instead of HTLC-timeout transactions, and the opposite, because we would treat incoming HTLCs as outgoing and the opposite.

What I meant is that it's important that we consistently use what's coming from the `localCommit.spec`, not the `remoteCommit.spec`.

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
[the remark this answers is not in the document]
t-bast: > Do we even need the htlcTxs parameter at all now in UnsignedLocalCommit? I think it can be removed. Though that would require updating serialization and might not be worth the added discriminator logic.

Yes, that's exactly why: we added this field early on because we though we'd need it, and wanted to be future-proof to avoid changing the codecs when adding support for splice with htlcs. But it turns out that this is never used and we should have removed that field, but removing would require a codec change, so we instead just always keep it empty.

Filling it with the unsigned local htlc transactions would not be useful, and would only waste space in the backups, that's why we don't bother and always set this to `listOf()`.

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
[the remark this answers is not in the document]
remyers: Ok - gotcha. So that just supports the name change to `localHtlcs`. We can't prevent a caller from accidentally passing the htlcs from the `remoteCommit.spec`, but that would atleast cause the closing tests to fail.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
remyers: Added in [b72530d](https://github.com/ACINQ/lightning-kmp/pull/568/commits/b72530df3c547ac67a20058afbba959aa6b8835e).

remyers:
@t-bast sorry for the delay in getting the rebased version. I'm getting an error with the `liquidity ads lease backwards compatibility` test where it compares the current state with the checkpoints on lines [L119](https://github.com/remyers/lightning-kmp/blob/18c4a12a8f0509ce988f5c14ba22828588cca3cb/src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationTestsCommon.kt#L119) and [L138](https://github.com/remyers/lightning-kmp/blob/18c4a12a8f0509ce988f5c14ba22828588cca3cb/src/commonTest/kotlin/fr/acinq/lightning/serialization/StateSerializationTestsCommon.kt#L119). When I review the difference with the checkpoint I notice what must be the additional discriminator and zero value added in [9aa3be5](https://github.com/ACINQ/lightning-kmp/pull/568/commits/9aa3be57e10fadd8fa0e3db1d8e21cdb0255d8e5). 

I don't think it makes sense here that the checkpoint should exactly match all future states. I believe we should remove these assert lines otherwise it will always fail when we change the state data structures, even if backwards compatibility is preserved.

t-bast:
> I don't think it makes sense here that the checkpoint should exactly match all future states. I believe we should remove these assert lines otherwise it will always fail when we change the state data structures, even if backwards compatibility is preserved.

Sure, modify what you need to modify to get the rebase working, and then we'll see during the review.

src/commonTest/kotlin/fr/acinq/lightning/channel/states/SpliceTestsCommon.kt
```diff
@@ -22,20 +26,33 @@ import fr.acinq.lightning.wire.*
...
     fun `splice funds out`() {
         val (alice, bob) = reachNormal()
-        spliceOut(alice, bob, 50_000.sat)
+        val (nodes, preimage, _) = addHtlc(15_000_000.msat, alice, bob)
+        val (alice1, bob1) = crossSign(nodes.first, nodes.second)
+        val (alice2, bob2) = spliceOut(alice1, bob1, 50_000.sat)
+        val fee = spliceFee(alice2, capacity = 950_000.sat)
+        fulfillHtlc(0, preimage, alice2, bob2)
+        assertEquals(alice2.state.commitments.latest.localCommit.spec.toLocal, 750_000_000.msat - fee.toMilliSatoshi() - 15_000_000.msat)
+        assertEquals(alice2.state.commitments.latest.localCommit.spec.toRemote, 200_000_000.msat)
```
t-bast, line 43: Changes to this test and the following one are redundant with the other tests that contain HTLCs, we could keep those unchanged to have tests for the case where no HTLCs are in the commitment?

src/commonTest/kotlin/fr/acinq/lightning/channel/states/SpliceTestsCommon.kt
```diff
@@ -155,32 +176,41 @@ class SpliceTestsCommon : LightningTestSuite() {
...
-        val (_, bob) = reachNormal(aliceFundingAmount = 100_000.sat, bobFundingAmount = 10_000.sat, alicePushAmount = 0.msat, bobPushAmount = 0.msat)
+        val (alice, bob) = reachNormal(aliceFundingAmount = 100_000.sat, bobFundingAmount = 10_000.sat, alicePushAmount = 0.msat, bobPushAmount = 0.msat)
         val leaseRate = LiquidityAds.LeaseRate(0, 0, 100 /* 5% */, 1.sat, 200, 100.msat)
         run {
             val liquidityRequest = LiquidityAds.RequestRemoteFunding(1_000_000.sat, bob.currentBlockHeight, leaseRate)
             assertEquals(10_001.sat, liquidityRequest.rate.fees(FeeratePerKw(1000.sat), liquidityRequest.fundingAmount, liquidityRequest.fundingAmount).total)
             val cmd = ChannelCommand.Commitment.Splice.Request(CompletableDeferred(), null, null, liquidityRequest, FeeratePerKw(1000.sat))
-            val (_, actions1) = bob.process(cmd)
-            assertTrue(actions1.isEmpty())
+            val (bob1, actionsBob1) = bob.process(cmd)
```
t-bast, line 185: nit: we can use `reachQuiescent`

src/commonTest/kotlin/fr/acinq/lightning/channel/states/SpliceTestsCommon.kt
```diff
@@ -192,22 +222,27 @@ class SpliceTestsCommon : LightningTestSuite() {
     fun `reject splice_ack`() {
         val cmd = createSpliceOutRequest(25_000.sat)
         val (alice, bob) = reachNormal()
-        val (_, actionsAlice1) = alice.process(cmd)
-        val (bob1, actionsBob1) = bob.process(ChannelCommand.MessageReceived(actionsAlice1.hasOutgoingMessage<SpliceInit>()))
+        val (nodes, _, _) = addHtlc(15_000_000.msat, alice, bob)
+        val (alice0, bob0) = crossSign(nodes.first, nodes.second)
+        val (_, bobQuiescent, spliceInit) = reachQuiescent(cmd, alice0, bob0)
```
t-bast, line 227: Naming this state `bobQuiescent` isn't consistent, we end up mixing naming for alice and bob (you never use `aliceQuiescent`). It's tedious but more consistent to just increase the index of each state.

src/commonTest/kotlin/fr/acinq/lightning/channel/states/SpliceTestsCommon.kt
```diff
@@ -1169,11 +1380,80 @@ class SpliceTestsCommon : LightningTestSuite() {
...
+            assertIs<LNChannel<Normal>>(bob8)
+            val commitSigBob = actionsBob8.findOutgoingMessage<CommitSig>()
+            actionsBob8.has<ChannelAction.Storage.StoreState>()
+
+            checkCommandResponse(cmd.replyTo, parentCommitment, spliceInit)
+            assertIs<LNChannel<Normal>>(alice8)
+            return UnsignedSpliceFixture(alice8, commitSigAlice, bob8, commitSigBob)
+        }
+
+        private fun initiateSplice(alice: LNChannel<Normal>, bob: LNChannel<Normal>, inAmounts: List<Satoshi> = emptyList(), pushAmount: MilliSatoshi = 0.msat, outAmount: Satoshi? = null): Pair<LNChannel<Normal>, LNChannel<Normal>> {
```
t-bast, line 1444: It's more consistent with other functions to introduce one that combines splice-in and splice-out explicitly (not optionally like you're doing, which makes it hard to verify what exactly we're testing). Also, having a `pushAmount` here is overkill and makes it harder to check that the test does what it's supposed to do, we're explicitly testing the push mechanism in dedicated tests.

src/commonTest/kotlin/fr/acinq/lightning/channel/states/SpliceTestsCommon.kt
```diff
@@ -1187,9 +1467,10 @@ class SpliceTestsCommon : LightningTestSuite() {
             val (alice1, actionsAlice1) = alice.process(ChannelCommand.MessageReceived(commitSigBob))
             assertTrue(actionsAlice1.isEmpty())
             val (bob1, actionsBob1) = bob.process(ChannelCommand.MessageReceived(commitSigAlice))
+            val incomingHtlcs = bob1.commitments.latest.localCommit.spec.htlcs.incomings()
             when {
-                bob1.staticParams.useZeroConf -> assertEquals(actionsBob1.size, 4)
-                else -> assertEquals(actionsBob1.size, 3)
+                bob1.staticParams.useZeroConf -> assertEquals(actionsBob1.size, 4 + incomingHtlcs.size)
```
t-bast, line 1472: Instead of checking the number of actions, we should rather check that we have the right actions for pending HTLCs.

src/commonTest/kotlin/fr/acinq/lightning/channel/states/SpliceTestsCommon.kt
```diff
@@ -1385,7 +1672,11 @@ class SpliceTestsCommon : LightningTestSuite() {
 
             // Alice's transactions confirm.
             val (alice4, actionsAlice4) = alice3.process(ChannelCommand.WatchReceived(WatchEventConfirmed(alice3.channelId, BITCOIN_TX_CONFIRMED(bobCommitTx), alice3.currentBlockHeight, 43, bobCommitTx)))
-            assertEquals(actionsAlice4.size, 1)
+            val failedOutgoingHtlcs = when (bobCommitTx.txid == alice3.commitments.latest.remoteCommit.txid) {
```
t-bast, line 1675: Same here, it's better to check that we have the right actions for pending HTLCs. That allowed me to find a bug, which is fixed in https://github.com/ACINQ/lightning-kmp/commit/b7b436466d9b3e0d1ca962ab4794f74cc41aedeb

src/commonTest/kotlin/fr/acinq/lightning/channel/states/SpliceTestsCommon.kt
remyers: `reachQuiescent` expects and returns a `SpliceInit` message, so it can't be used for splice failures.

remyers:
> Looking good, I improved the tests and found a small bug in [b7b4364](https://github.com/ACINQ/lightning-kmp/commit/b7b436466d9b3e0d1ca962ab4794f74cc41aedeb), which you can probably apply to your branch and then squash with the existing commit?

Thanks for the review/fixes. I've squashed them into [699d52f](https://github.com/ACINQ/lightning-kmp/commit/699d52f75cf19c36dc12099305c042146086b016).

t-bast (commented):
ACK https://github.com/ACINQ/lightning-kmp/pull/568/commits/699d52f75cf19c36dc12099305c042146086b016, let's wait to make sure liquidity ads is stable before merging this to `master`, but it's looking good to me :+1:

t-bast (approved):
LGTM!

PR #569, opened 2023-12-12
Fix `InboundLiquidityOutgoingPayment.miningFees`

pm47:


src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
```diff
@@ -773,7 +773,7 @@ data class Normal(
                 )
             )
             liquidityLease?.let {
-                add(ChannelAction.Storage.StoreOutgoingPayment.ViaInboundLiquidityRequest(txId = action.fundingTx.txId, lease = it))
+                add(ChannelAction.Storage.StoreOutgoingPayment.ViaInboundLiquidityRequest(miningFees = action.fundingTx.sharedTx.tx.fees, txId = action.fundingTx.txId, lease = it))
```
t-bast, line 776: I think that would be confusing, because the `miningFees` would only include the splice shared input and output. For clarify it should also include the `miningFee` of the lease, shouldn't it? I'll update that directly on my branch and we'll see what seems best.

pm47: I agree, following up on the parent PR.

PR #575, opened 2023-12-15
Hardcode liquidity lease rate

pm47:
In the future it will be provided by the peer in an `Init` TLV.

Also, make `fundingWeight` depend on the requested amount.


pm47:
Superseded by #576.

PR #577, opened 2023-12-19
Update bitcoin-kmp and secp256k1-kmp

sstone:
Upgrading secp256k1-kmp will fix the payment request parsing test that is currently failing (see https://github.com/ACINQ/secp256k1-kmp/issues/91)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -33,7 +33,7 @@ kotlin {
 
     val commonMain by sourceSets.getting {
         dependencies {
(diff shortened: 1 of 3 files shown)


t-bast:
I think we should make `PaymentRequest.read` return a `Try`, and we should add the test vectors from #567. I updated #567 with the latest version of bitcoin-kmp, I think it should supersede this PR?

sstone:
Yes it's not needed any more.

PR #578, opened 2024-01-02
Add `require_confirmed_inputs` to RBF messages

t-bast:
This was missing from the spec, but is more flexible and clearer than inheriting values from the previous attempt.
We don't really support checking that flag yet, but we should ensure our codecs support reading it.


modules/core/src/commonMain/kotlin/fr/acinq/lightning/wire/InteractiveTxTlv.kt
```diff
@@ -228,6 +235,13 @@ sealed class TxAckRbfTlv : Tlv {
             override fun read(input: Input): SharedOutputContributionTlv = SharedOutputContributionTlv(LightningCodecs.int64(input).sat)
         }
     }
+
+    /** Require confirmed inputs for the transaction being built. */
+    object RequireConfirmedInputsTlv : TxAckRbfTlv(), TlvValueReader<RequireConfirmedInputsTlv> {
+        override val tag: Long get() = 2
+        override fun write(out: Output) = Unit
+        override fun read(input: Input): RequireConfirmedInputsTlv = this
```
sstone, line 243: nice! (though it looks a bit puzzling at first)

modules/core/src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -575,6 +575,7 @@ data class TxInitRbf(
     constructor(channelId: ByteVector32, lockTime: Long, feerate: FeeratePerKw, fundingContribution: Satoshi) : this(channelId, lockTime, feerate, TlvStream(TxInitRbfTlv.SharedOutputContributionTlv(fundingContribution)))
 
     val fundingContribution = tlvs.get<TxInitRbfTlv.SharedOutputContributionTlv>()?.amount ?: 0.sat
+    val requireConfirmedInputs: Boolean = tlvs.get<TxInitRbfTlv.RequireConfirmedInputsTlv>()?.let { true } ?: false
```
sstone, line 578: this is never used so all I can do is check that the code is correct (it is). Not sure it's worth adding a test though ?

sstone (approved):

PR #581, opened 2024-01-10
Send `batch_size` on `commit_sig` retransmit

t-bast:
If we get disconnected after sending `commit_sig`, we will retransmit that message when reconnecting if our peer has not received it.

When we have multiple commitments, we need to retransmit one message per commitment, and include the `batch_size` tlv.

We incorrectly stored `commit_sig` without the `batch_size` tlv in the next remote commit field, and thus retransmitted without that tlv.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/json/JsonSerializers.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/json/JsonSerializers.kt
@@ -88,6 +88,7 @@
     JsonSerializers.ChannelReadyTlvSerializer::class,
     JsonSerializers.CommitSigTlvAlternativeFeerateSigSerializer::class,
     JsonSerializers.CommitSigTlvAlternativeFeerateSigsSerializer::class,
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/fr/acinq/lightning/json/JsonSerializers.kt
```diff
@@ -512,6 +514,9 @@ object JsonSerializers {
     @Serializer(forClass = CommitSigTlv.AlternativeFeerateSigs::class)
     object CommitSigTlvAlternativeFeerateSigsSerializer
 
+    @Serializer(forClass = CommitSigTlv.Batch::class)
+    object CommitSigTlvBatchSerializer
```
pm47, line 518: What prompted you to add this serializer? Was something failing?

t-bast: Yes, tests are otherwise failing.

pm47 (approved):

PR #583, opened 2024-01-18
Purge pending incoming payments on disconnection

t-bast:
We were keeping pending incoming payments in the internal state of the `IncomingPaymentHandler` even when disconnected. This could lead to inconsistencies with the channel state: if an HTLC that was added to the `IncomingPaymentHandler` was failed by another instance of the app (for example a background task), that failure wouldn't propagate to the state of the `IncomingPaymentHandler`. This can lead to partially fulfilled payments, where we think we have received the complete payment, but one of the part was actually failed so we shouldn't release the preimage.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
@@ -32,12 +32,14 @@ data class HtlcPart(val htlc: UpdateAddHtlc, override val finalPayload: PaymentO
     override val amount: MilliSatoshi = htlc.amountMsat
     override val totalAmount: MilliSatoshi = finalPayload.totalAmount
     override val paymentHash: ByteVector32 = htlc.paymentHash
+    override fun toString(): String = "htlc(channelId=${htlc.channelId},id=${htlc.id})"
 }
 
 data class PayToOpenPart(val payToOpenRequest: PayToOpenRequest, override val finalPayload: PaymentOnion.FinalPayload) : PaymentPart() {
     override val amount: MilliSatoshi = payToOpenRequest.amountMsat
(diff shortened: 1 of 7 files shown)


src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -402,15 +402,11 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val db: IncomingPayment
...
+     * Pay-to-open requests will be forgotten by the LSP, so we need to do the same otherwise we will accept outdated ones.
+     * Offered HTLCs that haven't been resolved will be re-processed when we reconnect.
      */
-    fun purgePayToOpenRequests() {
-        val valuesToReplace = pending.mapValues { entry -> entry.value.copy(parts = entry.value.parts.filter { it !is PayToOpenPart }.toSet()) }
-        pending.plusAssign(valuesToReplace)
-        val keysToRemove = pending.filterValues { it.parts.isEmpty() }.keys
-        pending.minusAssign(keysToRemove)
-    }
+    fun purgePendingPayments() = pending.clear()
```
pm47, line 409: That was easy! 🧹 

Some logging would be nice, also in `purgeExpiredPayments`, to have a better idea of what is going on when debugging.

pm47 (approved):

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
t-bast: Good idea, done in https://github.com/ACINQ/lightning-kmp/pull/583/commits/f3b79f5a4f17a60bfb10a73674314bb37b7e826d

pm47 (approved):

PR #584, opened 2024-01-18
Rotate taproot swap-in addresses

pm47:
It uses groundwork made in branch [swap-in-potentiam-taproot-rotate](https://github.com/ACINQ/lightning-kmp/tree/snapshot/swap-in-potentiam-taproot-rotate), but replaces hardcoded addresses by proper unbounded rotation.

Some cleanup is done at the end, by extracting wallets outside of `Peer`.

_Best reviewed commit-by-commit._


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -189,12 +189,8 @@ class Peer(
...
     }
 
-    val finalWallet = ElectrumMiniWallet(nodeParams.chainHash, watcher.client, scope, nodeParams.loggerFactory, name = "final")
-    val finalAddress: String = nodeParams.keyManager.finalOnChainWallet.address(addressIndex = 0L).also { finalWallet.addAddress(it) }
-
-    val swapInWallet = ElectrumMiniWallet(nodeParams.chainHash, watcher.client, scope, nodeParams.loggerFactory, name = "swap-in")
-    val legacySwapInAddress: String = nodeParams.keyManager.swapInOnChainWallet.legacySwapInProtocol.address(nodeParams.chain).also { swapInWallet.addAddress(it) }
-    val swapInAddress: String = nodeParams.keyManager.swapInOnChainWallet.swapInProtocol.address(nodeParams.chain).also { swapInWallet.addAddress(it) }
+    val finalWallet = FinalWallet(nodeParams.chain, nodeParams.keyManager.finalOnChainWallet, watcher.client, scope, nodeParams.loggerFactory)
+    val swapInWallet = SwapInWallet(nodeParams.chain, nodeParams.keyManager.swapInOnChainWallet, watcher.client, addressGenerationWindow = 20, scope, nodeParams.loggerFactory)
```
pm47, line 193: Notice the default address generation window.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/SwapInWallet.kt
```diff
@@ -0,0 +1,52 @@
...
+                it.addresses
+                    .filter { it.value.unspent.isEmpty() }
+                    .mapNotNull { (key, value) -> value.meta.index?.let { key to it } }
+                    .minByOrNull { it.second }
+            }
+                .filterNotNull()
+                .distinctUntilChanged()
+                .collect { (address, index) ->
+                    logger.info { "setting current swap-in address=$address index=$index" }
+                    swapInAddressFlow.emit(address to index)
```
pm47, line 46: If the wallet has already used N addresses, we do not want to emit all those addresses sequentially at startup until we detect the first unused address.

Current implementation guarantees that we will not emit addresses that have already been used, because an address will not be added to `WalletState` before we receive the first `ScriptHashSubscriptionResponse` from Electrum.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -143,19 +159,31 @@ class ElectrumMiniWallet(
...
+                    logger.info { "${unspents.size} utxo(s) for address=$bitcoinAddress index=${addressMeta.indexOrNull ?: "n/a"} balance=${nextWalletState.totalBalance}" }
                     unspents.forEach { logger.debug { "utxo=${it.outPoint.txid}:${it.outPoint.index} amount=${it.value} sat" } }
+                    when (val generator = addressGenerator) {
+                        null -> {}
+                        else -> if (addressMeta.indexOrNull == generator.second - 1 && nextAddressState.unspent.isNotEmpty()) {
+                            logger.info { "requesting generation of next sequence of addresses" }
+                            // if the window = 10 then the first series of address is 0 to 9 and addressGenerator.second = 10
+                            // then when address 9 has utxos, we request generation up until (and excluding) index 10 + 10 = 20, this will generate addresses 10 to 19
+                            mailbox.send(WalletCommand.Companion.GenerateAddress(addressMeta.indexOrNull!! + 1 + generator.first.window))
```
sstone, line 184: Going through the mailbox means that if a user has received funds on 2 different swap-in addresses but in the same block,  there is a (small) chance that the wallet state will be updated twice and result in 2 different channel updates ? Since the generator is local to the wallet it should be possible to avoid since and only update the wallet state once scanning for the next unused addresses has completed ?

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
sstone: IIUC it works a bit differently from other wallets: we will first scan `addressGenerationWindow` addresses but then unused addresses will be scan one by one (and not in blocks of `addressGenerationWindow` addresses)

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
pm47: The `until` below guarantees that we will not call `AddAddress` twice for the same address:
https://github.com/ACINQ/lightning-kmp/blob/16af706680f5b280ae113ce8c38f636f0f690746/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt#L258-L260

So we will subscribe exactly once to each address and then the wallet state will be updated everytime we receive a `ScriptHashSubscriptionResponse`, I think that's the best we can do?

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
pm47: The last version ([5c4cbcb](https://github.com/ACINQ/lightning-kmp/pull/584/commits/5c4cbcb5973f65488a5f74854f493c32067bed3f)) removes this window altogether since we are simply iterating "synchronously" over addresses until we find the first unused one.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
pm47: Latest version implements your suggestion: [5c4cbcb](https://github.com/ACINQ/lightning-kmp/pull/584/commits/5c4cbcb5973f65488a5f74854f493c32067bed3f).

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -75,6 +76,19 @@ data class WalletState(val addresses: Map<String, List<Utxo>>) {
...
         val empty: WalletState = WalletState(emptyMap())
+
+        data class AddressState(val meta: AddressMeta, val alreadyUsed: Boolean, val utxos: List<Utxo>)
+
+        sealed class AddressMeta {
+            data object Single : AddressMeta()
+            data class Derived(val index: Int) : AddressMeta()
+        }
+
+        val AddressMeta.indexOrNull: Int?
```
sstone, line 87: nit:  `indexOrNull`  could be a property of `AddressMeta` (so its type could be `Int` and not `Int?` in `Derived`)

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWalletTest.kt
```diff
@@ -120,7 +145,85 @@ class ElectrumMiniWalletTest : LightningTestSuite() {
...
+                else -> error("should not go that far")
+            }
+        }
+
+        val walletState = wallet.walletStateFlow.drop(1).first() // first emitted wallet is empty
+
+        // this has been checked on the blockchain
+        assertEquals(4 + 6 + 1, walletState.utxos.size)
+        assertEquals(72_000_000.sat + 30_000_000.sat + 2_000_000.sat, walletState.totalBalance)
+        assertEquals(11, walletState.utxos.size)
```
sstone, line 200: nit: this is already checked above

sstone:
LGTM! There is yet no limit on the number of address that can be generated (it could be set explicitly in the wallet, or implicitly by the generator which could return `null` to signal that there are no more addresses)

src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWalletTest.kt
pm47: Done in [6eba0f8](https://github.com/ACINQ/lightning-kmp/pull/584/commits/6eba0f8c05bb90c3a74e6b2daf29cd6770bee56d).

pm47:
Rebased on master.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
pm47: It would still have to be nullable though, because `Single` doesn't have one.

sstone (commented):
LGTM, just a nit.
There is no cap on the number of addresses that can be generated. I don't think it's a problem: eletrum servers will add a 30 seconds delay when the "cost" of a connection becomes too expensive (only if a user has 100s of addresses).

pm47:
> There is no cap on the number of addresses that can be generated. I don't think it's a problem: eletrum servers will add a 30 seconds delay when the "cost" of a connection becomes too expensive (only if a user has 100s of addresses).

I think that it's better than worrying about unintentional address reuse, and it will be regulated by increasingly slow sync times anyway.

src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -172,6 +201,30 @@ class ElectrumMiniWallet(
...
+                } else this
+            } ?: this
+        }
+
+        suspend fun WalletState.addAddress(generator: WalletCommand.Companion.AddressGenerator, addressIndex: Int): WalletState {
+            return this.addAddress(generator.generateAddress(addressIndex), WalletState.Companion.AddressMeta.Derived(addressIndex))
+        }
+
+        suspend fun WalletState.maybeGenerateNext(generator: WalletCommand.Companion.AddressGenerator): WalletState {
+            val lastDerivedAddressState = this.addresses[this.lastDerivedAddress?.first]
```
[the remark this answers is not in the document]
pm47: Done in 82c0d227.

sstone (approved):

PR #585, opened 2024-01-19
Emit sensitive task event for mpp

pm47:
When the first part of a multi-part payment is received, we emit a `SensitiveTaskEvents.TaskStarted` hoping that iOS will let the app running long enough for subsequent parts to arrive.

@robbiehanson we discussed providing some metadata on the payment within this mechanism (amount/total received, etc.), it doesn't seem to work as only the first part triggers a `TaskStarted`. This implementation doesn't have any metadata aside from the `paymentHash`.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -795,7 +795,17 @@ class Peer(
             is Either.Left -> incomingPaymentHandler.process(item.value, currentBlockHeight)
         }
         when (result) {
-            is IncomingPaymentHandler.ProcessAddResult.Accepted -> _eventsFlow.emit(PaymentReceived(result.incomingPayment, result.received))
+            is IncomingPaymentHandler.ProcessAddResult.Accepted -> {
+                if ((result.incomingPayment.received?.receivedWith?.size ?: 0) > 1) {
+                    // this was a multi-part payment, we signal that the task is finished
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -795,7 +795,17 @@ class Peer(
...
+                    // this was a multi-part payment, we signal that the task is finished
+                    nodeParams._nodeEvents.tryEmit(SensitiveTaskEvents.TaskEnded(SensitiveTaskEvents.TaskIdentifier.IncomingMultiPartPayment(result.incomingPayment.paymentHash)))
+                }
+                _eventsFlow.emit(PaymentReceived(result.incomingPayment, result.received))
+            }
+            is IncomingPaymentHandler.ProcessAddResult.Pending -> if (result.pendingPayment.parts.size == 1) {
+                // this is the first part of a multi-part payment, we request to keep the app alive to receive subsequent parts
+                nodeParams._nodeEvents.tryEmit(SensitiveTaskEvents.TaskStarted(SensitiveTaskEvents.TaskIdentifier.IncomingMultiPartPayment(result.incomingPayment.paymentHash)))
+            }
             else -> Unit
```
t-bast, line 809: Shouldn't we also stop the task when failing an HTLC set? Or are you afraid that it will create a mess if we fail a first HTLC set and receive new parts in parallel (we'd be both trying to start and end the task)?

t-bast: Oh right, I missed that detail, we only start the task when we're fulfilling the payment, that makes sense.

t-bast (approved):

PR #586, opened 2024-01-25
Add postman to send and receive onion messages

thomash-acinq:
Adds basic plumbing for onion messages:
`OnionMessages` to build blinded routes and onion messages
`Postman` to send and receive messages

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/message/Postman.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/message/Postman.kt
@@ -1,0 +1,153 @@
+package fr.acinq.lightning.message
+
+import fr.acinq.bitcoin.ByteVector
+import fr.acinq.bitcoin.ByteVector32
+import fr.acinq.bitcoin.PrivateKey
+import fr.acinq.bitcoin.PublicKey
+import fr.acinq.lightning.Lightning.randomBytes32
+import fr.acinq.lightning.Lightning.randomKey
+import fr.acinq.lightning.crypto.RouteBlinding
+import fr.acinq.lightning.crypto.sphinx.Sphinx
+import fr.acinq.lightning.utils.Either
+import fr.acinq.lightning.utils.Either.Left
+import fr.acinq.lightning.utils.Either.Right
+import fr.acinq.lightning.wire.*
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.channels.SendChannel
+import kotlinx.coroutines.withTimeoutOrNull
+import kotlin.math.max
+import kotlin.time.Duration
+
+class Postman(
+    val privateKey: PrivateKey,
+    val remoteNodeId: PublicKey,
+    val sendOnionMessage: (onionMessage: OnionMessage) -> SendMessageError?
+) {
+    private val nodeId = privateKey.publicKey()
+
+    data class MessageWithRecipientData(val messageOnion: MessageOnion, val recipientData: RouteBlindingEncryptedData)
+
+    data class SendMessageError(val error: String)
+
+    private val subscribed: HashMap<ByteVector32, SendChannel<MessageWithRecipientData>> = HashMap()
+
+    suspend fun processOnionMessage(msg: OnionMessage) {
+        val blindedPrivateKey = RouteBlinding.derivePrivateKey(privateKey, msg.blindingKey)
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/fr/acinq/lightning/message/Postman.kt
```diff
@@ -0,0 +1,153 @@
...
+import fr.acinq.lightning.utils.Either.Left
+import fr.acinq.lightning.utils.Either.Right
+import fr.acinq.lightning.wire.*
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.channels.SendChannel
+import kotlinx.coroutines.withTimeoutOrNull
+import kotlin.math.max
+import kotlin.time.Duration
+
+class Postman(
```
t-bast, line 21: Since we're only supporting quite a small set of actions compared to eclair (sending and receiving onion messages in the context of bolt12 payments), and only plan on supporting very specific onion messages flows in the future on a case-by-case basis, I think we should optimize for readability and be much more explicit about what we are doing. I believe we only support the following scenarios:

- receiving an `onion_message` containing an `invoice_request`, to which we respond with an invoice if it matches one of our offers
- sending an `invoice_request`
- receiving an `onion_message` in response to one of our `invoice_request`, which can either:
  - contain a bolt 12 invoice that we'll want to pay
  - contain an `invoice_error`
- everything else should just be silently ignored

Is that correct? If that's the case, I think a clear API exposed by the `Postman` should look like this:

```kotlin
sealed class OnionMessageAction {
    data class SendMessage(val message: OnionMessage): OnionMessageAction()
    data class PayInvoice(val payment: SendPayment, val invoice: Bolt12Invoice): OnionMessageAction() 
    data class ReportPaymentFailure(val payment: SendPayment, val failure: PaymentFailure): OnionMessageAction()
}

sealed class PaymentFailure {
    data object InvoiceNotReceived: PaymentFailure()
    data class InvoiceError(val message: String): PaymentFailure()
}

class Postman(val privateKey: PrivateKey, val remoteNodeId: PublicKey) {
    // I believe that this function doesn't need to suspend, as we don't want to block waiting for the invoice.
    // We'll need to slightly modify the existing SendPayment class (from Peer.kt) to handle bolt 12 payments (maybe make it a sealed class).  
    fun requestInvoice(payment: SendPayment): List<OnionMessageAction> {
        // Create the onion message that sends the invoice_request.
        // Record this outgoing message in our subscribed map (which can probably be renamed pending_invoice_requests for clarity).
        // Return OnionMessageAction.SendMessage.
        TODO()
    }

    suspend fun receiveMessage(msg: OnionMessage): List<OnionMessageAction> {
        // Decrypt onion message.
        // If this is an invoice_request:
        //  - check that the offer matches an offer in our DB
        //  - if it does, generate an invoice, create an outgoing onion message using the reply path, and return OnionMessageAction.SendMessage
        //  - otherwise, just ignore and return listOf()
        // If this is a valid Bolt 12 invoice that can be found in our pending_invoice_requests map:
        //  - return OnionMessageAction.PayInvoice and remove from our map
        // If this is an invoice_error that can be found in our pending_invoice_requests map:
        //  - return OnionMessageAction.ReportPaymentFailure and remove from our map
        // Otherwise:
        //  - just ignore and return listOf()
        TODO()
    }

    fun checkInvoiceRequestTimeout(timeout: Duration): List<OnionMessageAction> {
        // For each request in pending_invoice_requests that was created more than <timeout> ago:
        //  - remove it from the map
        //  - return OnionMessageAction.ReportPaymentFailure with PaymentFailure.InvoiceNotReceived
        TODO()
    }
}
```

t-bast (commented):
I don't think this PR should be the first PR that other PRs depend on, as it contains a lot of important architectural choices (in `Postman.kt`) that only need to be implemented once we have all the helper functions around onion messages and bolt 12.

I think the first set of PRs that we can very easily integrate are:

- adding all the types related to onion messages and bolt 12 with encoding/decoding tests that match eclair
- adding all the helpers function to encrypt/decrypt onion messages

And after that we can assemble those into a flow for paying bolt 12 offers and receiving bolt 12 payments. What do you think? Am I missing something important here?

thomash-acinq:
I've removed the dependence on Postman for the other PRs: #601, #602

t-bast:
Replaced by #624

PR #587, opened 2024-01-26
Remove `WalletState.consistent` boolean

pm47:
In the initial implementation of the wallet, we were asynchronously retrieving the parent transaction of utxos. When a new utxo was detected, there would be a period of time where `WalletState` would contain the utxo, but not the parent tx, hence the need for the `consistent` boolean.

But since #411 the Electrum interface has become "synchronous" with the use of `suspend` functions.

Note that if for some reason we cannot retrieve the transaction from Electrum, then we ignore the utxo, which is what we were doing previously with a `walletStateFlow.filter { it.consistent }` in Peer.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
@@ -15,13 +15,8 @@ import org.kodein.log.Logger
 import org.kodein.log.LoggerFactory
 import org.kodein.log.newLogger
 
-data class WalletState(val addresses: Map<String, List<UnspentItem>>, val parentTxs: Map<TxId, Transaction>) {
-    /** Electrum sends parent txs separately from utxo outpoints, this boolean indicates when the wallet is consistent */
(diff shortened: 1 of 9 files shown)


src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWallet.kt
```diff
@@ -34,7 +29,7 @@ data class WalletState(val addresses: Map<String, List<UnspentItem>>, val parent
         swapInParams = swapInParams, currentBlockHeight = currentBlockHeight, all = utxos,
     )
 
-    data class Utxo(val previousTx: Transaction, val outputIndex: Int, val blockHeight: Long) {
+    data class Utxo(val txId: TxId, val outputIndex: Int, val blockHeight: Long, val previousTx: Transaction) {
```
sstone, line 32: why not use `previousTx.txId` here ?

pm47: It's not strictly necessary, but it's a first step before making `previousTx` optional in a future PR.

PR #593, opened 2024-02-05
Switch to Kermit logger dependency

dpad85:
This PR switches the logging utility from kodein-log to [kermit](https://github.com/touchlab/Kermit). With kotlin 1.9, we are reworking the way logging is written to disk on Phoenix on iOS.

### No API changes to create loggers

A `LoggerFactory` class has been created to avoid API changes. Creating a logger is done like before : use the `loggerFactory` in `NodeParams`, and call `newLogger(this::class)`.

Projects (like Phoenix) using lightning-kmp will have to create a logger factory with their `LoggerConfig` (a simple data class containing the minimum log severity filter for the project, and the actual log writers).

### Small API change for MDC

Context is provided as a `Map<String, Any>` to the logger:

```kotlin
// before
logger.mdcinfo { "hello" }

// now
logger.info(mdc()) { "hello" }

// where mdc() would be:
private fun mdc(): Map<String, Any> {
    return mapOf(
        "utxos" to walletStateFlow.value.utxos.size,
        "balance" to walletStateFlow.value.totalBalance
    )
}
````

### Added a test logger

A logger factory has been added to the tests (see `testLoggerFactory`). It is available in `LightningTestSuite`. Logs are written as basic print outputs.

Note that Kermit provides [a specific testing lib](https://kermit.touchlab.co/docs/TESTING) to assert conditions on logs during tests. We don't use it for now, but it might be useful later on.


pm47 (approved):
LGTM.

Some bikeshedding may be needed around MDC, but it's better to get the library change out of the way asap.

PR #594, opened 2024-02-07
Fix CI issues

sstone:
We upgrade gradle to 8.4, which is the version of gradle used in KMP templates, and it may fix a problem we have on some Github Actions builds.
We also modify our test logger to write to stderr and not stdout, which seemed to cause problems with the processing of XML test results by our CI.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
-distributionSha256Sum=f6b8596b10cce501591e92f229816aa4046424f3b24d771751b06779d58c8ec4
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-bin.zip
```
dpad85, line 3: Why remove the shasum ?

pm47, line 3:
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-bin.zip
distributionSha256Sum=3e1af3ae886920c3ac87f7a91f816c0c7c436f276a6eefdb3da152100fef72ae
```

Source: https://gradle.org/release-checksums/

sstone: done  in https://github.com/ACINQ/lightning-kmp/pull/594/commits/87d472a87e6ee6b491aca0a8899038ecac1df4c7

pm47 (approved):

pm47:
Wait, the last build is failing so the issue remains.

sstone:
Our build error is related to the generation of XML  result files used to report test status to CI systems, but I cannot reproduce it locally. Most builds fail but not all of them.

sstone:
https://github.com/ACINQ/lightning-kmp/pull/594/commits/fb32db0ec82329913b9c8bb34a25376ac1787c52 seems to fix our build issue

pm47:
> https://github.com/ACINQ/lightning-kmp/commit/fb32db0ec82329913b9c8bb34a25376ac1787c52 seems to fix our build issue

Why would it help that logs are written to stderr instead of stdout? NB: My PR #596 was forked _before_ the kermit commit and build still fails.

sstone:
Because I think that the XML result files are piped from `stdout` and writing to it (stack traces for example) may produce invalid XML files.

sstone:
> Why would it help that logs are written to stderr instead of stdout? NB: My PR #596 was forked _before_ the kermit commit and build still fails.

The build error in this case was different (`java.lang.ArrayIndexOutOfBoundsException`) and may be linked to a pb with kryo that gets fixed by upgrading gradle. It's hard to be 100% since I cannot reproduce the pb locally but in both cases (`java.lang.ArrayIndexOutOfBoundsException` and `Cannot write XML results`) the problem is not with the tests (they all pass) but with how the results are reported to the CI environment.

pm47 (approved):
🕯️

PR #595, opened 2024-02-07
Add `Stfu` type to decode logic

pm47:
Oops! 😄


t-bast (approved):
Damn, we should have added a decoding test to catch that earlier!

t-bast (approved):

remyers (approved):

PR #597, opened 2024-02-09
Implement `option_simple_close`

t-bast:
Implement `option_simple_close` as defined in https://github.com/lightning/bolts/pull/1096
This is a pre-requisite for taproot channels.


modules/core/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -628,6 +628,27 @@ class Peer(
             }
     }
 
+    /**
+     * Estimate the actual fee that will be paid when closing the given channel at the target feerate.
+     */
+    fun estimateFeeForMutualClose(channelId: ByteVector32, targetFeerate: FeeratePerKw): ChannelManagementFees? {
```
pm47, line 634: There should be a symmetrical function for the actual close, for consistency, it should be between `requestInboundLiquidity()` and `payInvoice()`. Here is the one I used during testing:

```kotlin
    suspend fun mutualClose(channelId: ByteVector32, scriptPubKey: ByteVector, feerate: FeeratePerKw): ChannelCloseOutgoingPayment? {
        return channels.values
            .filterIsInstance<ChannelStateWithCommitments>()
            .filter { it is Normal || it is ShuttingDown || it is Negotiating }
            .find { it.channelId == channelId }
            ?.let {
                val res = this.async {
                    nodeParams.nodeEvents
                        .filterIsInstance<PaymentEvents.PaymentSent>()
                        .map { it.payment }
                        .filterIsInstance<ChannelCloseOutgoingPayment>()
                        .first()
                }
                send(WrappedChannelCommand(channelId, ChannelCommand.Close.MutualClose(scriptPubKey, feerate)))
                res.await()
            }
    }
```

modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
```diff
@@ -304,88 +297,127 @@ object Helpers {
...
-            remoteScriptPubkey: ByteArray,
-            requestedFeerate: ClosingFeerates
-        ): Pair<ClosingTx, ClosingSigned> {
-            val closingFees = firstClosingFee(commitment, localScriptPubkey, remoteScriptPubkey, requestedFeerate)
-            return makeClosingTx(channelKeys, commitment, localScriptPubkey, remoteScriptPubkey, closingFees)
+            localScriptPubkey: ByteVector,
+            remoteScriptPubkey: ByteVector,
+            feerate: FeeratePerKw,
+            lockTime: Long,
+        ): Either<ChannelException, Pair<Transactions.ClosingTxs, ClosingComplete>> {
```
pm47, line 316: Having a `logger` here to log our feerate would have been extra nice for testing, but prob not worth the hassle.

pm47 (commented):
LGTM.

Tested against eclair in nominal and RBF cases.

modules/core/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
pm47: Totally, this should ideally return a `TxId` (that's what the phoenixd api returns currently).

pm47: In my basic eclair-phoenix integration test with a close initiated by phoenix, I reach:
https://github.com/ACINQ/lightning-kmp/blob/92d22d6338afc1eed12d02f1112099ead2463022/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Negotiating.kt#L64

Before reaching:
https://github.com/ACINQ/lightning-kmp/blob/92d22d6338afc1eed12d02f1112099ead2463022/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Negotiating.kt#L87

So the `txId` returned in the `ChannelCloseResponse` doesn't match the first published closing tx. I don't think it makes sense?

pm47: > On top of that, I think that we should update the ACINQ node: when a phoenix wallet initiates a mutual close, we should not generate a mutual close transaction where the fees are paid by our node, we should let the user pay the closing fees instead? Except in the case where the user doesn't have any funds left in the channel, because in that case we still benefit from doing a mutual close. What do you think?

Agreed!

pm47 (approved):

sstone (approved):

PR #598, opened 2024-02-12
Compact node id

thomash-acinq:
Adds the `NodeId` interface that can be either a public key or a pair channel id, direction that's more compact.
This is used as introduction node for blinded routes.
We also use `NodeId` for the next node to relay a message to. Which is not in the spec but is necessary for us since we have no way to resolve the compact node id to a public key ourselves.  It requires a compatible peer that accepts this format.

Depends on #586


thomash-acinq:
Replace by #601 that removes the dependence on Postman

PR #599, opened 2024-02-12
Offer types

thomash-acinq:
Adds types for Bolt 12 offers and invoice requests (but not invoice yet).

Depends on #598


thomash-acinq:
Replace by #602 that removes the dependence on Postman

PR #601, opened 2024-02-13
NodeId

thomash-acinq:
Adds the NodeId interface that can be either a public key or a pair channel id, direction that's more compact. This is used as introduction node for blinded routes. We also use NodeId for the next node to relay a message to. Which is not in the spec but is necessary for us since we have no way to resolve the compact node id to a public key ourselves. It requires a compatible peer that accepts this format.


src/commonMain/kotlin/fr/acinq/lightning/wire/MessageOnion.kt
```diff
@@ -6,19 +6,20 @@ import fr.acinq.bitcoin.io.ByteArrayInput
 import fr.acinq.bitcoin.io.ByteArrayOutput
 import fr.acinq.bitcoin.io.Input
 import fr.acinq.bitcoin.io.Output
+import fr.acinq.lightning.NodeId
 import fr.acinq.lightning.crypto.RouteBlinding
 
-
-sealed class OnionMessagePayloadTlv : Tlv {
+sealed interface OnionMessagePayloadTlv : Tlv {
```
t-bast, line 12: Why did you change that to a `sealed interface`? For that type of cases (functionally, an `enum`), Kotlin usually recommends using a `sealed class`.

src/commonMain/kotlin/fr/acinq/lightning/wire/RouteBlinding.kt
```diff
@@ -22,14 +23,14 @@ sealed class RouteBlindingEncryptedDataTlv : Tlv {
     }
 
     /** Id of the next node. */
-    data class OutgoingNodeId(val nodeId: PublicKey) : RouteBlindingEncryptedDataTlv() {
+    data class OutgoingNodeId(val nodeId: NodeId) : RouteBlindingEncryptedDataTlv() {
```
t-bast, line 26: You're modifying the official spec TLV here (tag = 4). We need to get agreement on this change at the spec level for that, otherwise we may have painful compatibility issues to handle in the future...I believe that changing the `outgoing_node_id` TLV to use this new `sciddir_or_pubkey` makes a lot of sense and should be in the BOLTs, don't you? This is an opportunity to create a small spec PR based on `master` that:

- extracts the `sciddir_or_pubkey` from the offers spec PR
- uses it in the `outgoing_node_id` TLV field of onion messages / blinded paths
- adds or modifies a test vector to use this

I expect that this spec PR would be accepted somewhat quickly, and it would make the offers spec PR slightly smaller in scope. It's also important to create that spec PR before people actually start using offers, this way we can just change the TLV without caring about backwards-compat.

src/commonMain/kotlin/fr/acinq/lightning/wire/MessageOnion.kt
thomash-acinq: I actually thought that sealed interface was the way to go. I'll use sealed class from now on.

src/commonMain/kotlin/fr/acinq/lightning/NodeId.kt
```diff
@@ -0,0 +1,41 @@
+package fr.acinq.lightning
+
+import fr.acinq.bitcoin.PublicKey
+import fr.acinq.bitcoin.io.Input
+import fr.acinq.bitcoin.io.Output
+import fr.acinq.lightning.wire.LightningCodecs
+
+sealed interface NodeId {
```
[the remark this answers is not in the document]
thomash-acinq: Done

src/commonMain/kotlin/fr/acinq/lightning/wire/RouteBlinding.kt
t-bast: We can merge that code to both `lightning-kmp` and `eclair` though, even before the spec PR gets accepted, because:

- until we ship bolt 12 support in phoenix, we actually will never create such tlv fields
- on the eclair side, we're just adding support for _reading_ this new case, not writing it, so it won't impact other nodes and we can update the code without caring about backwards-compat

But we should still create a spec PR to get wide support for this feature!

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningCodecs.kt
```diff
@@ -225,4 +224,26 @@ object LightningCodecs {
...
+        val firstByte = byte(input)
+        if (firstByte == 0 || firstByte == 1) {
+            val isNode1 = firstByte == 0
+            val scid = ShortChannelId(int64(input))
+            return EncodedNodeId.ShortChannelIdDir(isNode1, scid)
+        } else if (firstByte == 2 || firstByte == 3) {
+            val publicKey = PublicKey(ByteArray(1) { firstByte.toByte() } + bytes(input, 32))
+            return EncodedNodeId.Plain(publicKey)
+        } else {
+            throw IllegalArgumentException("unexpected first byte: $firstByte")
```
t-bast, line 237: Note that you'll need to be careful when decrypting an onion message and decoding its content to catch that exception.

t-bast (approved):

src/commonTest/kotlin/fr/acinq/lightning/wire/RouteBlindingTestsCommon.kt
```diff
@@ -14,7 +15,7 @@ class RouteBlindingTestsCommon : LightningTestSuite() {
             ByteVector("01080000000000000000 042102edabbd16b41c8371b92ef2f04c1185b4f03b6dcd52ba9b78d9d7c89c8f221145") to RouteBlindingEncryptedData(
                 TlvStream(
                     Padding(ByteVector("0000000000000000")),
-                    OutgoingNodeId(PublicKey(ByteVector("02edabbd16b41c8371b92ef2f04c1185b4f03b6dcd52ba9b78d9d7c89c8f221145")))
+                    OutgoingNodeId(EncodedNodeId(PublicKey(ByteVector("02edabbd16b41c8371b92ef2f04c1185b4f03b6dcd52ba9b78d9d7c89c8f221145"))))
```
t-bast, line 18: Actually it would be useful to have an encoding test that uses the scid_dir format, that we would share with eclair (and maybe even the spec).

t-bast (changes requested):

src/commonTest/kotlin/fr/acinq/lightning/wire/RouteBlindingTestsCommon.kt
thomash-acinq: I've added a test, here and in eclair too (https://github.com/ACINQ/eclair/pull/2821)

t-bast: It looks like you didn't push that commit though, I don't see any test...

thomash-acinq: Indeed, pushed just now

t-bast (approved):

PR #602, opened 2024-02-13
Offer types

thomash-acinq:
Adds types for Bolt 12 offers and invoice requests (but not invoice yet).

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/wire/MessageOnion.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/wire/MessageOnion.kt
@@ -6,7 +6,6 @@ import fr.acinq.bitcoin.io.ByteArrayInput
 import fr.acinq.bitcoin.io.ByteArrayOutput
 import fr.acinq.bitcoin.io.Input
 import fr.acinq.bitcoin.io.Output
-import fr.acinq.lightning.EncodedNodeId
 import fr.acinq.lightning.crypto.RouteBlinding
 
 sealed class OnionMessagePayloadTlv : Tlv {
--- a/src/commonMain/kotlin/fr/acinq/lightning/wire/MessageOnion.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/wire/MessageOnion.kt
@@ -81,48 +81,13 @@ sealed class OnionMessagePayloadTlv : Tlv {
      */
     data class Invoice(val tlvs: TlvStream<OfferTypes.InvoiceTlv>) : OnionMessagePayloadTlv() {
         override val tag: Long get() = Invoice.tag
-        override fun write(out: Output) = tlvSerializer.write(tlvs, out)
+        override fun write(out: Output) = OfferTypes.Invoice.tlvSerializer.write(tlvs, out)
 
         companion object : TlvValueReader<Invoice> {
             const val tag: Long = 66
 
-            val tlvSerializer = TlvStreamSerializer(
-                true, @Suppress("UNCHECKED_CAST") mapOf(
-                    // Invoice request part that must be copy-pasted from above
-                    OfferTypes.InvoiceRequestMetadata.tag to OfferTypes.InvoiceRequestMetadata.Companion as TlvValueReader<OfferTypes.InvoiceTlv>,
-                    OfferTypes.OfferChains.tag to OfferTypes.OfferChains.Companion as TlvValueReader<OfferTypes.InvoiceTlv>,
-                    OfferTypes.OfferMetadata.tag to OfferTypes.OfferMetadata.Companion as TlvValueReader<OfferTypes.InvoiceTlv>,
-                    OfferTypes.OfferCurrency.tag to OfferTypes.OfferCurrency.Companion as TlvValueReader<OfferTypes.InvoiceTlv>,
-                    OfferTypes.OfferAmount.tag to OfferTypes.OfferAmount.Companion as TlvValueReader<OfferTypes.InvoiceTlv>,
-                    OfferTypes.OfferDescription.tag to OfferTypes.OfferDescription.Companion as TlvValueReader<OfferTypes.InvoiceTlv>,
-                    OfferTypes.OfferFeatures.tag to OfferTypes.OfferFeatures.Companion as TlvValueReader<OfferTypes.InvoiceTlv>,
-                    OfferTypes.OfferAbsoluteExpiry.tag to OfferTypes.OfferAbsoluteExpiry.Companion as TlvValueReader<OfferTypes.InvoiceTlv>,
-                    OfferTypes.OfferPaths.tag to OfferTypes.OfferPaths.Companion as TlvValueReader<OfferTypes.InvoiceTlv>,
-                    OfferTypes.OfferIssuer.tag to OfferTypes.OfferIssuer.Companion as TlvValueReader<OfferTypes.InvoiceTlv>,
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
```diff
@@ -0,0 +1,1003 @@
...
+
+    sealed interface Bolt12Tlv : Tlv
+
+    sealed interface InvoiceTlv : Bolt12Tlv
+
+    sealed interface InvoiceRequestTlv : InvoiceTlv
+
+    sealed interface OfferTlv : InvoiceRequestTlv
+
+    sealed interface InvoiceErrorTlv : Bolt12Tlv
```
t-bast, line 60: Same here, those should be `sealed class` (I believe this should work the same, you won't need to nest the types inside the `sealed class`).

src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
```diff
@@ -0,0 +1,1003 @@
...
+        companion object : TlvValueReader<InvoiceRequestQuantity> {
+            const val tag: Long = 86
+            override fun read(input: Input): InvoiceRequestQuantity {
+                return InvoiceRequestQuantity(LightningCodecs.tu64(input))
+            }
+        }
+    }
+
+    /**
+     * A public key for which the sender know the corresponding private key.
```
t-bast, line 368: nit:
```suggestion
     * A public key for which the sender knows the corresponding private key.
```

src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
```diff
@@ -0,0 +1,1003 @@
...
+    fun filterOfferFields(tlvs: TlvStream<InvoiceRequestTlv>): TlvStream<OfferTlv> {
+        // Offer TLVs are in the range (0, 80).
+        return TlvStream<OfferTlv>(
+            tlvs.records.filterIsInstance<OfferTlv>().toSet(),
+            tlvs.unknown.filter{it.tag < 80}.toSet()
+        )
+    }
+
+    fun filterInvoiceRequestFields(tlvs: TlvStream<InvoiceTlv>): TlvStream<InvoiceRequestTlv>
+    {
```
t-bast, line 643: nit: formatting

```suggestion
    fun filterInvoiceRequestFields(tlvs: TlvStream<InvoiceTlv>): TlvStream<InvoiceRequestTlv> {
```

src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
```diff
@@ -0,0 +1,1003 @@
...
+                return Signature(ByteVector64(LightningCodecs.bytes(input, input.availableBytes)))
+            }
+        }
+    }
+
+    fun filterOfferFields(tlvs: TlvStream<InvoiceRequestTlv>): TlvStream<OfferTlv> {
+        // Offer TLVs are in the range (0, 80).
+        return TlvStream<OfferTlv>(
+            tlvs.records.filterIsInstance<OfferTlv>().toSet(),
+            tlvs.unknown.filter{it.tag < 80}.toSet()
```
t-bast, line 638: This matches what we do in eclair, but I don't really remember the spec, can you refresh my memory? That means we don't use the "standard" mechanisms for unknown TLVs here (a TLV with tag 65501 would be dropped), is that expected?

src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
```diff
@@ -0,0 +1,1003 @@
...
+
+        companion object : TlvValueReader<Error> {
+            const val tag: Long = 5
+            override fun read(input: Input): Error {
+                return Error(LightningCodecs.bytes(input, input.availableBytes).decodeToString())
+            }
+        }
+    }
+
+    sealed interface InvalidTlvPayload {
```
t-bast, line 696:
```suggestion
    sealed class InvalidTlvPayload {
```

src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
```diff
@@ -0,0 +1,1003 @@
...
+                    OfferIssuer.tag to OfferIssuer as TlvValueReader<OfferTlv>,
+                    OfferQuantityMax.tag to OfferQuantityMax as TlvValueReader<OfferTlv>,
+                    OfferNodeId.tag to OfferNodeId as TlvValueReader<OfferTlv>,
+                )
+            )
+
+            fun decode(s: String): Offer {
+                val (prefix, encoded, encoding) = Bech32.decodeBytes(s.lowercase(), true)
+                require(prefix == hrp)
+                require(encoding == Bech32.Encoding.Beck32WithoutChecksum)
```
t-bast, line 786: We shouldn't throw exceptions here, it will crash the app when scanning something that isn't an offer (we've seen issues like this when parsing invalid Bolt 11 invoices that threw exceptions and fixed it since). The `decode` functions (all of them) should return a `Try`.

src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
```diff
@@ -0,0 +1,1003 @@
...
+            fun validate(records: TlvStream<InvoiceErrorTlv>): Either<InvalidTlvPayload, InvoiceError> {
+                if (records.get<Error>() == null) return Left(MissingRequiredTlv(5))
+                return Right(InvoiceError(records))
+            }
+        }
+    }
+
+    fun <T : Tlv> rootHash(tlvStream: TlvStream<T>, serializer: TlvStreamSerializer<T>): ByteVector32 {
+        // This encoding/decoding step ensures that the resulting tlvs are ordered.
+        val encoded = serializer.write(tlvStream)
```
t-bast, line 942: I think that this can be simplified. In eclair we needed to do this because the `Tlv` trait didn't expose the `tag` as it is entirely handled by `scodec`. But here in `lightning-kmp`, the `Tlv` sealed class exposes the `tag` field, so we can rely on this instead of unnecessarily encoding/decoding.

src/commonMain/kotlin/fr/acinq/lightning/wire/MessageOnion.kt
```diff
@@ -58,6 +58,97 @@ sealed class OnionMessagePayloadTlv : Tlv {
...
+     * requested offer (if it was an offer we published).
+     */
+    data class Invoice(val tlvs: TlvStream<OfferTypes.InvoiceTlv>) : OnionMessagePayloadTlv() {
+        override val tag: Long get() = Invoice.tag
+        override fun write(out: Output) = tlvSerializer.write(tlvs, out)
+
+        companion object : TlvValueReader<Invoice> {
+            const val tag: Long = 66
+
+            val tlvSerializer = TlvStreamSerializer(
```
t-bast, line 89: I think it would be more consistent to have this in `OfferTypes`, like we do for `Offer` and `InvoiceRequest`.

src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
thomash-acinq: Done

t-bast (approved):

PR #607, opened 2024-02-20
Read and write PaymentRequest

thomash-acinq:


src/commonMain/kotlin/fr/acinq/lightning/payment/PaymentRequest.kt
```diff
@@ -1,14 +1,28 @@
...
+    abstract val paymentHash: ByteVector32
+    abstract val features: Features
 
-    fun isExpired(currentTimestampSeconds: Long = currentTimestampSeconds()): Boolean
+    abstract fun isExpired(currentTimestampSeconds: Long = currentTimestampSeconds()): Boolean
+
+    abstract fun write(): String
+
+    companion object {
+        fun read(input: String): Try<PaymentRequest> = runTrying {
```
t-bast, line 20: Why the `runTrying` since `Bolt11Invoice.read` and `Bolt12Invoice.fromString` already return a `Try`? This is unnecessary, in the exception case you are now explicitly throwing then re-catching the exception, which is extremely inefficient?

thomash-acinq: Actually, I've found a way to make it both readable and efficient with `.map { it }`.

PR #610, opened 2024-02-21
Revert "Enable paying Bolt12Invoice (#606)"

pm47:
PR #606 seems to cause a regression when testing against eclair (`InvalidOnionPayload`).

I believe the issue is due to the change related to `TrampolineOnion`/`OnionRoutingPacket.TrampolinePacketLength`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 
 allprojects {
     group = "fr.acinq.lightning"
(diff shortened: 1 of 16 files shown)


t-bast:
I believe #611 is sufficient, and will be easier to revert once https://github.com/ACINQ/eclair/pull/2810 is deployed on our node.

pm47:
Closed in favor of #611.

PR #613, opened 2024-02-29
Refactor key management

pgrange:
⚠️ Disclaimer:
* Being my first time with Kotlin, code might not be very idiomatic;
* It might be the case that this topic is already somewhere in your roadmap and you might have given it some though about some target architecture that could be very different from what I propose here;
* I've done my best to carefully craft each commit but, please, feel free to share any suggestion that could make this PR easier to review;
* any other feedback you can think of are welcome.

# Why

I'm planning to explore alternative implementations of a `KeyManager` and especially what can be done with [hardware](https://delvingbitcoin.org/t/lightning-hardware-wallet/555). As a pre-requisite, I believe some refactoring is needed to ensure that:
1. the `KeyManager` interface is fully agnostic to the key material implementation (currently there are some dependencies to `DeterministicWallet`);
2. private key material does not _leak_ outside `LocalKeyManager` (currently `DeterministicWallet.PrivateKey` appears in `Transactions`, `KeyManager`, `Channel`, `Bolt3Derivation`, `RouteBlinding` for instance.

I've noticed key material is way more concealed in [eclair-core](https://github.com/ACINQ/eclair/blob/android-phoenix/eclair-core/src/main/scala/fr/acinq/eclair/crypto/LocalKeyManager.scala#L54) (the `android-phoenix` branch).

# What

In the target architecture, the user application would be in charge of instantiating the `LocalKeyManager` as it's the application (or any lightning-kmp client for that matter) that's in charge of initiating, backing up or restoring the key material (in the form of a seed for instance). Then the application trusts lightning-kmp for the lightning logic, lightning-kmp using some abstract `KeyManager` implemented by `LocalKeyManager`:

```mermaid
block-beta
columns 3
Start(("User")) space:2
u2a<[" "]>(down) space:2

Application                                            a2l<["uses"]>(right)       LNKMP["Lighning Logic"]
downAgain<["Instantiates"]>(down) space                                l2k<["uses"]>(down)
LocalKeyManager                                 k2l<["delegates"]>(left) KeyManager 

style Start fill:#969;
```

That way, we may replace `LocalKeyManager` by alternative implementations like, for instance, a `HardwareKeyManager` or a `WhateverKeyManager`:

```mermaid
block-beta
columns 3
LNKMP["Lighning Logic"]:3
KeyManager:3
LocalKeyManager HardwareKeyManager WhateverKeyManager
```

# How

To limit the exposure of the key material, I've decided to create `PrivateKeyDescriptor` and `ExtendedPrivateKeyDescriptor` which are two interfaces, describing a private key that can be implemented to suit different crypto engines.

I've only implemented them interfaces for the `LocalKeyManager`.  `LocalKeyManager` and the local package is the only places where such keys can be instantiated and private key material  used for cryptography.

There are some cases where exporting private key material seems ok:
* NodeKeys, which are used for network authentication and communication;
* shaSeed, used for commitment points and eventually shared with channel peers;

It's also the case that Phoenix exports private key material to cypher cloud stored data for instance. In which case it seems very handy to rely on the `KeyManager` of lightning-kmp so that there is no need for another seed, or secret to backup/restore. But to open the way to alternative implementation, I've introduced the `deterministicKeyMaterial()` function.

Also note that, regarding the way Phoenix uses lightning-kmp, this PR introduces a breaking change. See commit `Refactor key management - BREAKING API`.

After this PR, to implement a `KeyManager`, one will have to also implement both interfaces as needed:
* PrivateKeyDescriptor
   *  publicKey
   * deriveForRevocation
   * deriveForCommitment
   * sign
   * signInputTaprootScriptPath
   * signMusig2TaprootInput
   * generateMusig2Nonce
* ExtendedPrivateKeyDescriptor
   *  privateKey
   * publicKey
   * derive
   * derivePrivateKey

# Notes

I've tried to carefully craft each commit to ease the review process. I suggest to review each commit separately instead of just the whole PR outcome as they come with detailed explanation of what's done and why it's done that way:
* _Refactor key management - fundingKey/paymentKey_ is the first commit, it introduces `PrivateKeyDescriptor` and handles the easiest part with the _funding_ and _payment_ keys which are used without basic derivation strategies;
* _Refactor key management - Bolt3_ focuses on everything that's Bolt3 related and, in particular it introduces two new implementations of `PrivateKeyDescriptor` to handle revocation and commitment derivations;
* _Refactor key management - swap in protocol_ focuses on the swap in protocol and introduces `ExtendedPrivateKeyDescriptor`;
* _Refactor key management - Bip84_ focuses on Bip84
* _Refactor key management - encapsulate Transactions.sign_ moves `Transactions.sign()` into the `PrivateKeyDescriptor` interface and is the first commit to migrate key operation into this interface
* _Refactor key management - segregate instantiate() to LocalKeyManager_ make it so that only the `LocalKeyManager` or code in the `local` package can instantiate actual private key material
* _Refactor key management - segregate instantiate() again_ does the same thing but for extended private keys
* _Refactor key management - BREAKING API_ introduces a breaking change. The commit message describes what kind of changes (which look minimal) are to be done to adapt the code of Phoenix to this breaking change.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
@@ -126,7 +126,7 @@ data class LocalCommit(val index: Long, val spec: CommitmentSpec, val publishabl
             if (commit.htlcSignatures.size != sortedHtlcTxs.size) {
                 return Either.Left(HtlcSigCountMismatch(params.channelId, sortedHtlcTxs.size, commit.htlcSignatures.size))
             }
-            val htlcSigs = sortedHtlcTxs.map { Transactions.sign(it, keyManager.htlcKey.deriveForCommitment(localPerCommitmentPoint), SigHash.SIGHASH_ALL) }
+            val htlcSigs = sortedHtlcTxs.map { Transactions.sign2(it, keyManager.htlcKey.deriveForCommitment(localPerCommitmentPoint), SigHash.SIGHASH_ALL) }
             val remoteHtlcPubkey = params.remoteParams.htlcBasepoint.deriveForCommitment(localPerCommitmentPoint)
             // combine the sigs to make signed txs
             val htlcTxsAndSigs = Triple(sortedHtlcTxs, htlcSigs, commit.htlcSignatures).zipped().map { (htlcTx, localSig, remoteSig) ->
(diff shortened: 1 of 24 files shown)


pgrange:
Hi @t-bast ,

Let's close it and I'll re-open if it makes sense after taproot channels.

PR #619, opened 2024-03-11
Minor API improvements for sending/receiving payments

pm47:
Commits are independent.

Define a `NodeEvent.PaymentReceived` that aims to replace the eponym class in the `Peer` (which is marked as deprecated, but otherwise left untouched). The previous impl wasn't emitted for pay-to-open parts.

Also define a "synchronous' `sendLightning` method similar to other peer methods.


dpad85 (approved):
Looks good to me, won't cause integartion issues with Phoenix.

t-bast (approved):
LGTM, conflicts with #618 should be easy to handle.

PR #622, opened 2024-03-26
Fix mempool.space ios tests

sstone:
These tests include https calls which won't work if the ios simulator is started in standalone mode, which is the default mode.


sstone:
See https://youtrack.jetbrains.com/issue/KT-38317 for a relevant issue, and the workaround I'm using here.

pm47:
Wow, the fix is pretty involved. For the CI it's probably fine, but I'm a bit worried about the implication on a normal dev machine, would such global calls not have side effects and break the environment? Let's wait for @robbiehanson's opinion

sstone:
> Wow, the fix is pretty involved. For the CI it's probably fine, but I'm a bit worried about the implication on a normal dev machine, would such global calls not have side effects and break the environment? Let's wait for @robbiehanson's opinion

There's another workaround in https://youtrack.jetbrains.com/issue/KT-61470/Kotlin-Native-Keychain-operations-fail-in-iOS-tests-due-to-missing-entitlements but it's even more complex... it also mentions that the underlying issue will be fixed in Kotlin 2.1.0.

I agree that this fix may break things on dev machine, so now it will only be used on github's CI,

sstone:
> So the tests will keep failing on dev machine, that's right? There is no filtering to only run test on CI. I think it's good that way, just making sure.

Yes they will run and fail on dev machines.

robbiehanson:
I can confirm that the various MempoolSpace tests fail on my machine (macOS). But all other tests succeed. And a `publishToMavenLocal` also works.

pm47 (approved):

PR #623, opened 2024-03-29
Build onion messages

thomash-acinq:
Build onion messages and routes for onion messages. Also make blinded paths mandatory when building offers as we are a wallet hidden behind a trampoline node.


t-bast (commented):
I've done some refactoring and documentation in #625, but otherwise LGTM

thomash-acinq:
I've merged your refactor, thanks.

t-bast (approved):
LGTM! It would be nice to get rid of the `offer_node_id`, let's keep track of this potential change in #626

PR #626, opened 2024-04-11
Generate default deterministic offer

t-bast:
Add the option to generate a default, deterministic offer that can be restored based on the node's seed and the trampoline node it uses. This offer has the minimal amount of information necessary. It doesn't expire, which makes it suitable to save for long-term, recurring payments.

We also include the latest changes to the specification that allow us to create this minimal offer:

- the `offer_node_id` can be omitted if a blinded path is provided
- the `offer_description` can be omitted if the offer doesn't have an amount

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
@@ -714,7 +714,7 @@ object OfferTypes {
         } else {
             null // TODO: add exchange rates
         }
-        val description: String = records.get<OfferDescription>()!!.description
+        val description: String? = records.get<OfferDescription>()?.description
         val features: Features = records.get<OfferFeatures>()?.features ?: Features.empty
         val expirySeconds: Long? = records.get<OfferAbsoluteExpiry>()?.absoluteExpirySeconds
         private val paths: List<ContactInfo.BlindedPath>? = records.get<OfferPaths>()?.paths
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
```diff
@@ -714,15 +714,14 @@ object OfferTypes {
...
         val features: Features = records.get<OfferFeatures>()?.features ?: Features.empty
         val expirySeconds: Long? = records.get<OfferAbsoluteExpiry>()?.absoluteExpirySeconds
         private val paths: List<ContactInfo.BlindedPath>? = records.get<OfferPaths>()?.paths
         val issuer: String? = records.get<OfferIssuer>()?.issuer
         val quantityMax: Long? = records.get<OfferQuantityMax>()?.max?.let { if (it == 0L) Long.MAX_VALUE else it }
-        val nodeId: PublicKey = records.get<OfferNodeId>()!!.publicKey
-
-        val contactInfos: List<ContactInfo> = paths ?: listOf(ContactInfo.RecipientNodeId(nodeId))
+        val nodeId: PublicKey? = records.get<OfferNodeId>()?.publicKey
+        val contactInfos: List<ContactInfo> = paths ?: listOfNotNull(nodeId?.let { ContactInfo.RecipientNodeId(it) })
```
thomash-acinq, line 724:
```suggestion
        val contactInfos: List<ContactInfo> = paths ?: listOf(ContactInfo.RecipientNodeId(nodeId!!))
```

thomash-acinq (approved):

src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
t-bast: Good idea, that makes it clear that `contactInfos` cannot be empty. Done in https://github.com/ACINQ/lightning-kmp/pull/626/commits/43d40efc3c0f89a51645c5f970554072b1547347

thomash-acinq (approved):

Issue #616, opened 2024-03-06
More useful errors when sending fails


pm47:
Our sending code results in generic `NoAvailableChannels` errors when a failure is returned by the channel (`Commitment.sendAdd()`). This logic seems to be inherited from eclair and makes sense in a multi-channels context, but not in lightning-kmp where we basically have 0 or 1 channel in `Normal` state. It would be much more helpful to return the error from the last channel tried.

PR #628, opened 2024-04-13
NoAvailableChannels error to include specific details

ndungudedan:
This PR addresses issue [616](https://github.com/ACINQ/lightning-kmp/issues/616).

Updated the `NoAvailableChannels` object to accommodate an error description.


t-bast:
Thanks for your attempt, this is actually more complex than this and requires knowledge of the channel state machine. I've fixed this issue in #634.

PR #634, opened 2024-05-02
More detailed error when HTLC could not be sent

### LINKED ISSUE #616: More useful errors when sending fails

t-bast:
We use more detailed errors when we fail to send an HTLC, since this may happen because the channel is offline, syncing, opening or closing.

It may also happen that we have a channel in the `Normal` state that has enough funds, but we cannot send the HTLC because it hits a channel limit (too many HTLCs pending, a splice is ongoing, etc). When that happens we will retry ignoring the channel, which leads to the default `NoAvailableChannels` error: in that case, the "real" error can be found in the `failures` list of the `OutgoingPaymentFailure` instance.

We now explicitly map every failure to a potentially user-friendly error (`PartFailure`), whenever possible. When the error cannot be easily interpreted, we use a `PartFailure.Uninterpretable` error to wrap the actual error. Applications like Phoenix should define localized error messages for each of these errors and print them on failures.

:warning: We change the format of `LightningOutgoingPayment.Part.Status.Failed`, which is stored in the DB. The migration should be handled by:

- adding a new column `error_code` and a mapping from `PartFailure` to an error code
- for `PartFailure.Uninterpretable`, the existing `details` column should be used for the error message
- old rows should use the `error_code` of `PartFailure.Uninterpretable` and just display the stored `details` message

What do you think @dpad85, is that an acceptable way of handling backwards-compat for previously failed payments?

Fixes #616

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentFailure.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentFailure.kt
@@ -56,12 +56,13 @@ sealed class PartFailure {
     data object ChannelIsSplicing : PartFailure()
     /** The channel is closing: another channel should be created to send the payment. */
     data object ChannelIsClosing : PartFailure()
+    sealed class RouteFailure : PartFailure()
     /** A remote node had a temporary failure: the payment may succeed if retried. */
-    data object TemporaryRemoteFailure : PartFailure()
+    data object TemporaryRemoteFailure : RouteFailure()
     /** The payment amount could not be relayed to the recipient, most likely because they don't have enough inbound liquidity. */
-    data object RecipientLiquidityIssue : PartFailure()
+    data object RecipientLiquidityIssue : RouteFailure()
     /** The payment recipient is offline and could not accept the payment. */
-    data object RecipientIsOffline : PartFailure()
+    data object RecipientIsOffline : RouteFailure()
     /** The payment recipient received the payment but rejected it. */
     data object RecipientRejectedPayment : PartFailure()
     /** This is an error that cannot be easily interpreted: we don't know what exactly went wrong and cannot correctly inform the user. */
@@ -130,20 +131,6 @@ data class OutgoingPaymentFailure(val reason: FinalFailure, val failures: List<L
                 }
             }
             return LightningOutgoingPayment.Part.Status.Failed(converted, completedAt)
-        }
-
-        /** Return true if this is a remote routing error. */
-        fun isRouteError(failure: LightningOutgoingPayment.Part.Status.Failed) = when (failure.failure) {
-            is PartFailure.TemporaryRemoteFailure -> true
-            is PartFailure.RecipientLiquidityIssue -> true
-            is PartFailure.RecipientIsOffline -> true
-            else -> false
-        }
-
-        /** Return true if the recipient received the payment and rejected it, indicating that we shouldn't retry. */
-        fun isRejectedByRecipient(failure: LightningOutgoingPayment.Part.Status.Failed) = when (failure.failure) {
-            is PartFailure.RecipientRejectedPayment -> true
-            else -> false
         }
     }
 }
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentFailure.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentFailure.kt
@@ -1,7 +1,6 @@
 package fr.acinq.lightning.payment
 
 import fr.acinq.bitcoin.utils.Either
-import fr.acinq.lightning.MilliSatoshi
 import fr.acinq.lightning.channel.*
 import fr.acinq.lightning.db.LightningOutgoingPayment
 import fr.acinq.lightning.utils.currentTimestampMillis
@@ -25,7 +24,7 @@ sealed class FinalFailure {
     data object ChannelOpening : FinalFailure() { override fun toString(): String = "channel creation is in progress, please retry when ready" }
     data object ChannelClosing : FinalFailure() { override fun toString(): String = "channel closing is in progress, please retry when a new channel has been created" }
     data object NoAvailableChannels : FinalFailure() { override fun toString(): String = "payment could not be sent through existing channels, check individual failures for more details" }
-    data class InsufficientBalance(val available: MilliSatoshi) : FinalFailure() { override fun toString(): String = "not enough funds in wallet to afford payment (available = $available)" }
+    data object InsufficientBalance : FinalFailure() { override fun toString(): String = "not enough funds in wallet to afford payment" }
     data object NoRouteToRecipient : FinalFailure() { override fun toString(): String = "unable to route payment to recipient" }
     data object RecipientUnreachable : FinalFailure() { override fun toString(): String = "the recipient was offline or did not have enough liquidity to receive the payment" }
     data object RetryExhausted: FinalFailure() { override fun toString(): String = "payment attempts exhausted without success" }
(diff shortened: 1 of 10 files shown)


src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentFailure.kt
```diff
@@ -1,29 +1,71 @@
...
+    /** Payments are temporarily paused while a channel is splicing: the payment can be retried after the splice. */
+    data object ChannelIsSplicing : PartFailure()
+    /** The channel is closing: another channel should be created to send the payment. */
+    data object ChannelIsClosing : PartFailure()
+    /** A remote node had a temporary failure: the payment may succeed if retried. */
+    data object TemporaryRemoteFailure : PartFailure()
+    /** The payment amount could not be relayed to the recipient, most likely because they don't have enough inbound liquidity. */
+    data object RecipientLiquidityIssue : PartFailure()
+    /** The payment recipient is offline and could not accept the payment. */
+    data object RecipientIsOffline : PartFailure()
```
pm47, line 64: This would remove `isRouteError()`

```suggestion
    sealed class RouteError : FinalFailure()
    /** A remote node had a temporary failure: the payment may succeed if retried. */
    data object TemporaryRemoteFailure : RouteError ()
    /** The payment amount could not be relayed to the recipient, most likely because they don't have enough inbound liquidity. */
    data object RecipientLiquidityIssue : RouteError ()
    /** The payment recipient is offline and could not accept the payment. */
    data object RecipientIsOffline : RouteError ()
```

src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentFailure.kt
```diff
@@ -34,26 +76,73 @@ data class OutgoingPaymentFailure(val reason: FinalFailure, val failures: List<L
...
             else -> false
         }
 
-        fun isRejectedByRecipient(failure: LightningOutgoingPayment.Part.Status.Failed) = when (failure.remoteFailureCode) {
-            IncorrectOrUnknownPaymentDetails.code -> true
+        /** Return true if the recipient received the payment and rejected it, indicating that we shouldn't retry. */
+        fun isRejectedByRecipient(failure: LightningOutgoingPayment.Part.Status.Failed) = when (failure.failure) {
+            is PartFailure.RecipientRejectedPayment -> true
             else -> false
         }
```
pm47, line 147: This is so straightforward we can probably remove this function entirely?

```suggestion
        /** Return true if the recipient received the payment and rejected it, indicating that we shouldn't retry. */
        fun isRejectedByRecipient(failure: LightningOutgoingPayment.Part.Status.Failed) = failure.failure is PartFailure.RecipientRejectedPayment
```

pm47 (commented):
LGTM, just nits

src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentFailure.kt
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/634/commits/ea4eb74b9fe11277ccf93a28ebb249820b322374

src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentFailure.kt
t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/634/commits/ea4eb74b9fe11277ccf93a28ebb249820b322374

src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentFailure.kt
```diff
@@ -34,27 +77,60 @@ data class OutgoingPaymentFailure(val reason: FinalFailure, val failures: List<L
...
-            PermanentChannelFailure.code -> true
-            TemporaryNodeFailure.code -> true
-            PermanentNodeFailure.code -> true
-            else -> false
-        }
-
-        fun isRejectedByRecipient(failure: LightningOutgoingPayment.Part.Status.Failed) = when (failure.remoteFailureCode) {
-            IncorrectOrUnknownPaymentDetails.code -> true
-            else -> false
+        fun convertFailure(failure: Either<ChannelException, FailureMessage>, completedAt: Long = currentTimestampMillis()): LightningOutgoingPayment.Part.Status.Failed {
```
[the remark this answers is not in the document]
t-bast: Good catch, this also means we should move this new error sealed class to the `PaymentsDb` file. Done in https://github.com/ACINQ/lightning-kmp/pull/634/commits/bec5f00ebac1c4c05278e0956aeb7a07d32c66db

t-bast:
> What would be great is a method that takes the FinalFailure + the list of PartFailure and returns a single, most pertinent type.

I did something in https://github.com/ACINQ/lightning-kmp/pull/634/commits/293234718daa98e7c7cd0bcf0debd8a9cbde602f, I'm not sure we can do better. There are always cases that won't be easy to programmatically analyze and will require the exact details of each payment attempt to figure out what went wrong, but it should work in most cases.

src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentFailure.kt
```diff
@@ -1,60 +1,111 @@
...
+    data object FeaturesNotSupported : FinalFailure() { override fun toString(): String = "payment request features not supported" }
+    data object InvalidPaymentId : FinalFailure() { override fun toString(): String = "payment ID must be unique" }
+    data object ChannelNotConnected : FinalFailure() { override fun toString(): String = "channel is not connected yet, please retry when connected" }
+    data object ChannelOpening : FinalFailure() { override fun toString(): String = "channel creation is in progress, please retry when ready" }
+    data object ChannelClosing : FinalFailure() { override fun toString(): String = "channel closing is in progress, please retry when a new channel has been created" }
+    data object NoAvailableChannels : FinalFailure() { override fun toString(): String = "payment could not be sent through existing channels, check individual failures for more details" }
+    data class InsufficientBalance(val available: MilliSatoshi) : FinalFailure() { override fun toString(): String = "not enough funds in wallet to afford payment (available = $available)" }
```
dpad85, line 28: Making `InsufficientBalance` a data class is not consistent with the other failures, and brings some complexity down the line for the application. Having the available amount at the time of the payment is not very useful (at least on Phoenix, even if we had that data, we would not display it)

t-bast: Yeah, I wasn't sure whether it was a good idea or not. I removed it in https://github.com/ACINQ/lightning-kmp/pull/634/commits/8bd90a371fde70736b8128eedc6742c8ebe65627

dpad85:
Testing this PR on Phoenix, and trying to pay an already paid invoice a second time, I hoped to get a `Part.Status...RecipientRejectedPayment` failure.

Instead, the over payment status is `FinalFailure.RetryExhausted`. The part child status is `Part.Status...RecipientIsOffline` (seems to have failed with a `UnknownNextPeer` error which maps to that failure).

FYI, the `explain` method in that case returns `FinalFailure.RetryExhausted`.

So we are still not able to provide a meaningful error message to the user, even with the improvements added in the PR.

pm47 (approved):

dpad85 (approved):
LGTM, it's a nice improvement over what we have already.

PR #637, opened 2024-05-13
Improve bolt11 serialization performance

sstone:
It was extremely slow on native targets (linux, ios...), reworked method is about a 100 times faster.
We also remove our custom `BitStream` class (replaced by low-level bech32 methods provided by `bitcoin-kmp`).


src/commonTest/kotlin/fr/acinq/lightning/payment/Bolt11InvoiceTestsCommon.kt
```diff
@@ -530,6 +532,20 @@ class Bolt11InvoiceTestsCommon : LightningTestSuite() {
...
+        val int5s = (0..1000).map { random5() }
+        var foo = 0
+        val start = currentTimestampMillis()
+        (0..1000).forEach {
+            foo += Bolt11Invoice.toByteArray(int5s).size
+        }
+        val end = currentTimestampMillis()
+        println("${end  - start} $foo")
+    }
+
```
pm47, line 548: This test should be removed now that the issue is solved.

I used a variant in an intermediate commit to test non-regression:
```kotlin
    @Test
    fun `non-reg test`() {
        val random = Random.Default
        fun random5(): Byte = (random.nextInt() and 0x1f).toByte()
        for (i in 0..10000) {
            val data = buildList {
                for (j in 0..random.nextInt(1000)) {
                    add(random5())
                }
            }
            println("i=$i j=${data.size}")
            val b1 = Bolt11Invoice.toByteArray(data)
            val stream = BitStream()
            data.forEach {
                val bits = Bolt11Invoice.toBits(it)
                stream.writeBits(bits)
            }
            val b2 = stream.getBytes()
            assertArrayEquals(b1, b2)
        }
    }
```

pm47 (commented):
Less code, better perfs 🍾

src/commonTest/kotlin/fr/acinq/lightning/payment/Bolt11InvoiceTestsCommon.kt
sstone: Removed in https://github.com/ACINQ/lightning-kmp/pull/637/commits/2bc522a43df3ae308ff16d082d5c8d0eae029651

pm47 (approved):

PR #638, opened 2024-05-15
Add `toString` to payment part failures

pm47:
Follow-up to #634.

This makes `FinalFailure.explain()` return consistent results.


t-bast:
It was on purpose that this doesn't have a `toString()`, since wallet applications must provide their own localized string for each failure (this is in the comment on that `trait`). Why do you need it?

pm47:
`PaymentFailure` also must be localized, but we still provide human-readable `toString`s.

I'm trying to avoid doing this on phoenixd: https://github.com/ACINQ/phoenixd/pull/51/commits/27799dbd588c30e7893611705477f36aff403000

t-bast (dismissed):

t-bast (approved):

PR #642, opened 2024-05-20
Fix timeout for offer invoice requests

thomash-acinq:
Previously we triggered a timeout check with
```kotlin
coroutineScope {
    launch {
	delay(cmd.fetchInvoiceTimeout)
	input.send(CheckInvoiceRequestTimeout(pathId, cmd))
    }
}
```
However this blocks the current function which means the message is sent only after it has timed out. Instead we introduce delayed commands which are only processed when their time comes without blocking other peer commands. In addition to `CheckInvoiceRequestTimeout`, we also make `CheckPaymentsTimeout` a delayed command.


t-bast:
This is a dangerous change that has several potential drawbacks and subtle bugs (mostly around disconnections and management of the `input` channel and flushing of messages). I'm very wary of the way you're trying to achieve that. If we go that route, we'll need a thorough review from @dpad85 and @pm47 as well, and e2e tests in a version of Phoenix that uses this code.

> However this blocks the current function which means the message is sent only after it has timed out.

Can you detail that more? When a "normal" dispatcher is used, I don't see how this would end up blocking anything. Are you hitting this issue in tests, or in a real app?

If you're hitting this issue in unit tests, it's potentially a problem with the dispatcher (tests use a single-threaded dispatcher that doesn't play well with launching coroutines).

I think we should test this with integration in a real app to see if the timeout works or not (which will need @dpad85 to create a phoenix build using the current `master` branch of `lightning-kmp`), and a "hacky" payment recipient that ignores the invoice request (probably a tweaked eclair testnet node).

thomash-acinq:
Replaced by https://github.com/ACINQ/lightning-kmp/pull/645

PR #643, opened 2024-05-20
Fix serialization of onion messages

thomash-acinq:
The onion routing packet's size is 66 bytes more than its payload's size. We were taking this into account in the `read` function but not in the `write` one.


vincenzopalazzo (approved):
ACK https://github.com/ACINQ/lightning-kmp/pull/643/commits/4bbfdea6247019e96bdf6d63a2f0c43a245c3f21

t-bast (approved):

PR #644, opened 2024-05-21
Use `Peer` scope for invoice request timeout coroutines

t-bast:
We don't want to create a new coroutine scope, otherwise we block until its child tasks are completed. We instead want to launch a new task in the `Peer`'s main scope and don't want to wait for its completion.


t-bast:
Done in #645

PR #646, opened 2024-05-21
Enable route blinding

thomash-acinq:
Cleans the mess around `RouteBlinding` being an implicit feature in Bolt12 invoices.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
@@ -168,6 +168,7 @@ data class NodeParams(
         require(features.hasFeature(Feature.PaymentSecret, FeatureSupport.Mandatory)) { "${Feature.PaymentSecret.rfcName} should be mandatory" }
         require(features.hasFeature(Feature.ChannelType, FeatureSupport.Mandatory)) { "${Feature.ChannelType.rfcName} should be mandatory" }
         require(features.hasFeature(Feature.DualFunding, FeatureSupport.Mandatory)) { "${Feature.DualFunding.rfcName} should be mandatory" }
+        require(features.hasFeature(Feature.RouteBlinding)) { "${Feature.RouteBlinding.rfcName} should be supported" }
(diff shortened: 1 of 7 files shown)


src/commonTest/kotlin/fr/acinq/lightning/FeaturesTestsCommon.kt
```diff
@@ -5,6 +5,7 @@ import fr.acinq.lightning.Features.Companion.validateFeatureGraph
 import fr.acinq.lightning.tests.utils.LightningTestSuite
 import fr.acinq.lightning.utils.BitField
 import fr.acinq.secp256k1.Hex
+import org.kodein.memory.text.toHex
```
t-bast, line 8: nit: remove

src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
```diff
@@ -189,6 +189,7 @@ data class NodeParams(
             Feature.Wumbo to FeatureSupport.Optional,
             Feature.StaticRemoteKey to FeatureSupport.Mandatory,
             Feature.AnchorOutputs to FeatureSupport.Optional, // can't set Mandatory because peers prefers AnchorOutputsZeroFeeHtlcTx
+            Feature.RouteBlinding to FeatureSupport.Optional,
```
t-bast, line 192: Should we force this to be activated in `init` with a `require`, like other features that must be activated? This would be more consistent since we implicitly use route blinding without checking the feature bit.

thomash-acinq: Done

src/commonTest/kotlin/fr/acinq/lightning/FeaturesTestsCommon.kt
thomash-acinq: Done

t-bast (approved):

t-bast (approved):

PR #648, opened 2024-05-24
Smaller default offer

thomash-acinq:
Shaved 32 bytes from the default offer by removing the pathId.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
@@ -240,15 +240,14 @@ data class NodeParams(
      * This offer will stay valid after restoring the seed on a different device.
      * We also return the path_id included in this offer, which should be used to route onion messages.
      */
-    fun defaultOffer(trampolineNode: NodeUri): Pair<ByteVector32, OfferTypes.Offer> {
-        // We generate a deterministic path_id based on:
+    fun defaultOffer(trampolineNode: NodeUri): OfferTypes.Offer {
+        // We generate a deterministic blindingSecret based on:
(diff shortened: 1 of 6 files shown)


thomash-acinq:
We could save even more by replacing the trampoline node id with a `ShortChannelIdDir`.

src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
```diff
@@ -781,12 +781,12 @@ object OfferTypes {
                 nodeParams: NodeParams,
                 trampolineNode: NodeUri,
                 features: Features,
-                pathId: ByteVector32,
```
t-bast, line 784: nit: can you update the documentation line 776 accordingly? It should probably be:

```
@param blindingSecret session key used for the blinded path included in the offer.
```

t-bast (commented):
Nice catch, this is a good quick win :+1:

src/commonMain/kotlin/fr/acinq/lightning/wire/OfferTypes.kt
thomash-acinq: Done

src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
```diff
@@ -240,15 +240,14 @@ data class NodeParams(
      * This offer will stay valid after restoring the seed on a different device.
      * We also return the path_id included in this offer, which should be used to route onion messages.
```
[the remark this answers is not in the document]
thomash-acinq: Done

PR #649, opened 2024-05-24
On-the-fly channel funding based on splicing and liquidity ads

t-bast:
This is a work-in-progress, will be best reviewed commit-by-commit.
Supersedes #618.


src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -243,82 +219,192 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val db: IncomingPayment
...
+                                    is Either.Right -> {
+                                        val (requestedAmount, fundingRate) = result.value
+                                        val addToFeeCredit = run {
+                                            val featureOk = nodeParams.features.hasFeature(Feature.FundingFeeCredit)
+                                            val remoteNodeOk = remoteFundingRates.paymentTypes.contains(LiquidityAds.PaymentType.FromFeeCredit)
+                                            // We use an arbitrary threshold that is higher than just the current liquidity fees.
+                                            // This reduces the frequency of on-chain operations for payments that are about the size of the fees.
+                                            // It also ensures that if we end up splicing with a higher feerate because we have unconfirmed parent
+                                            // transactions, we will have enough fee credit to cover this higher feerate.
+                                            val feeCreditThreshold = fundingRate.fees(currentFeerate, requestedAmount, requestedAmount).total * 5
```
t-bast, line 248: The 5x multiplier is maybe too high and can end up with a non-negligible fee credit that stays forever on the LSP side. Should we reduce this to a simple 2x, with the risk that it creates more frequent on-chain transactions?

pm47 (commented):
Review for commit [e307660](https://github.com/ACINQ/lightning-kmp/pull/649/commits/e3076602d30733df11dd3a4061c34e288ba2bd35).

src/commonMain/kotlin/fr/acinq/lightning/channel/ChannelData.kt
```diff
@@ -350,23 +350,29 @@ data class LocalParams(
     val htlcMinimum: MilliSatoshi,
     val toSelfDelay: CltvExpiryDelta,
     val maxAcceptedHtlcs: Int,
-    val isInitiator: Boolean,
+    val isChannelOpener: Boolean,
```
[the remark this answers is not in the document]
t-bast: This ensures that every usage of that field is correctly updated (so that the compiler helps find them all). This is particularly useful when rebasing to correctly catch potential changes!

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -1241,18 +1171,238 @@ class Peer(
...
                 )
                 val msg = actions1.filterIsInstance<ChannelAction.Message.Send>().map { it.message }.filterIsInstance<OpenDualFundedChannel>().first()
                 _channels = _channels + (msg.temporaryChannelId to state1)
                 processActions(msg.temporaryChannelId, peerConnection, actions1)
             }
-            is PayToOpenResponseCommand -> {
-                logger.info { "sending ${cmd.payToOpenResponse::class.simpleName}" }
-                peerConnection?.send(cmd.payToOpenResponse)
+            is OpenOrSpliceChannel -> {
+                when (val channel = channels.values.firstOrNull { it is Normal }) {
```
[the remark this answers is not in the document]
t-bast: This increases the risk number of changes we need to review...that's probably worth doing in a preparatory PR that we can merge quickly, not in this one. I'll do that.

src/commonMain/kotlin/fr/acinq/lightning/payment/LiquidityPolicy.kt
```diff
@@ -3,38 +3,47 @@ package fr.acinq.lightning.payment
...
 sealed class LiquidityPolicy {
-    /** Never initiates swap-ins, never accept pay-to-open */
+    /** Never initiates swap-ins, never accept on-the-fly funding requests. */
     data object Disable : LiquidityPolicy()
 
     /**
-     * Allow automated liquidity managements, within relative and absolute fee limits. Both conditions must be met.
+     * Allow automated liquidity management, within relative and absolute fee limits. Both conditions must be met.
+     *
+     * @param inboundLiquidityTarget amount of inbound liquidity the buyer would like to maintain (can be set to null to disable)
```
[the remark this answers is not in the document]
t-bast: That's a good point, initially I wanted to only request additional inbound liquidity if our current inbound was below our target. But for simplicity I currently implemented the following behavior:

- on new channel creation, we request `inboundLiquidityTarget`
- on splice-in, we don't request anything
- on on-the-fly funding, we request `inboundLiquidityTarget` (even if we already had that amount available)

Which of those do you think I should change?

Note that your proposed comment wouldn't be correct either, since right now we don't re-request inbound liquidity when doing a splice-in.

src/commonMain/kotlin/fr/acinq/lightning/wire/ChannelTlv.kt
```diff
@@ -67,125 +66,26 @@ sealed class ChannelTlv : Tlv {
...
-                maxRelayFeeProportional = LightningCodecs.u16(input),
-                maxRelayFeeBase = LightningCodecs.u32(input).msat,
-            )
+        companion object : TlvValueReader<RequestFundingTlv> {
+            const val tag: Long = 1339
+            override fun read(input: Input): RequestFundingTlv = RequestFundingTlv(LiquidityAds.RequestFunds.read(input))
         }
     }
 
-    data class OriginTlv(val origin: Origin) : ChannelTlv() {
```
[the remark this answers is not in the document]
t-bast: Exactly, and I think it's a nice win!

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -1241,18 +1171,238 @@ class Peer(
...
+                            val (localFundingAmount, fees) = run {
+                                val dummyFundingScript = Script.write(Scripts.multiSig2of2(Transactions.PlaceHolderPubKey, Transactions.PlaceHolderPubKey)).byteVector()
+                                val localMiningFee = Transactions.weight2fee(currentFeerates.fundingFeerate, FundingContributions.computeWeightPaid(isInitiator = true, null, dummyFundingScript, cmd.walletInputs, emptyList()))
+                                // We directly pay the on-chain fees for our inputs/outputs of the transaction.
+                                val localFundingAmount = cmd.totalAmount - localMiningFee
+                                val leaseFees = requestRemoteFunding.fees(currentFeerates.fundingFeerate)
+                                // We also refund the liquidity provider for some of the on-chain fees they will pay for their inputs/outputs of the transaction.
+                                val totalFees = TransactionFees(miningFee = localMiningFee + leaseFees.miningFee, serviceFee = leaseFees.serviceFee)
+                                Pair(localFundingAmount, totalFees)
+                            }
```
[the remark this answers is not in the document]
t-bast: Yes, I think there is no way around that, we start with the amount we have available in our wallet, and we need to pre-compute all fees to tell our peer how much funds we want to have **in the channel** once funding is complete.

src/commonMain/kotlin/fr/acinq/lightning/payment/LiquidityPolicy.kt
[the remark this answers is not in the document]
pm47: > Which of those do you think I should change?

I think the current behavior is fine for now.

> Note that your proposed comment wouldn't be correct either, since right now we don't re-request inbound liquidity when doing a splice-in.

Right. The `LiquidityPolicy` probably needs to handle separately the on-chain and off-chain case, but this has high-level UX ramifications and can also be dealt with later.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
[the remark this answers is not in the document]
pm47: Here is my best try:

```suggestion
                            val (localFundingAmount, fees) = run {
                                // We need the local funding amount to be able use standard channel opening messages. Since fees
                                // are fully deterministic based on liquidity ads and feerates, we compute them first and proceed
                                // backwards to retrieve the funding amount.
                                val dummyFundingScript = Script.write(Scripts.multiSig2of2(Transactions.PlaceHolderPubKey, Transactions.PlaceHolderPubKey)).byteVector()
                                val localMiningFee = Transactions.weight2fee(currentFeerates.fundingFeerate, FundingContributions.computeWeightPaid(isInitiator = true, null, dummyFundingScript, cmd.walletInputs, emptyList()))
                                // We directly pay the on-chain fees for our inputs/outputs of the transaction.
                                val localFundingAmount = cmd.totalAmount - localMiningFee
                                val leaseFees = requestRemoteFunding.fees(currentFeerates.fundingFeerate)
                                // We also refund the liquidity provider for some of the on-chain fees they will pay for their inputs/outputs of the transaction.
                                val totalFees = TransactionFees(miningFee = localMiningFee + leaseFees.miningFee, serviceFee = leaseFees.serviceFee)
                                Pair(localFundingAmount, totalFees)
                            }
```

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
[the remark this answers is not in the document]
t-bast: Done in #659

src/commonMain/kotlin/fr/acinq/lightning/channel/ChannelData.kt
```diff
@@ -384,20 +390,33 @@ data class RemoteParams(
...
+data class ChannelFlags(val announceChannel: Boolean, val nonInitiatorPaysCommitFees: Boolean)
 
 data class ClosingTxProposed(val unsignedTx: ClosingTx, val localClosingSigned: ClosingSigned)
 
-/** Reason for creating a new channel or a splice. */
+/**
+ * @param miningFee fee paid to miners for the underlying on-chain transaction.
+ * @param serviceFee fee paid to our peer for any service provided with the on-chain transaction.
+ */
+data class TransactionFees(val miningFee: Satoshi, val serviceFee: Satoshi) {
```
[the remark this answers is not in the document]
t-bast: Done in #659

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
```diff
@@ -20,140 +24,352 @@ import fr.acinq.lightning.utils.sat
...
+                signature = LightningCodecs.bytes(input, 64).byteVector64(),
             )
         }
     }
 
     /** Request inbound liquidity from a remote peer that supports liquidity ads. */
-    data class RequestRemoteFunding(val fundingAmount: Satoshi, val leaseStart: Int, val rate: LeaseRate) {
-        private val leaseExpiry: Int = leaseStart + rate.leaseDuration
-        val requestFunds: ChannelTlv.RequestFunds = ChannelTlv.RequestFunds(fundingAmount, rate.leaseDuration, leaseExpiry)
+    data class RequestFunds(val requestedAmount: Satoshi, val fundingLease: FundingLease, val paymentDetails: PaymentDetails) {
```
pm47, line 307: Otherwise it sounds like we are creating an invoice:

```suggestion
    data class RequestFunding(val requestedAmount: Satoshi, val fundingLease: FundingLease, val paymentDetails: PaymentDetails) {
```

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
```diff
@@ -20,140 +24,352 @@ import fr.acinq.lightning.utils.sat
...
+    }
+
+    /** The fees associated with a given [LeaseRate] can be paid using various options. */
+    sealed class PaymentType {
+        /** Fees are transferred from the buyer's channel balance to the seller's during the interactive-tx construction. */
+        data object FromChannelBalance : PaymentType()
+        /** Fees will be deducted from future HTLCs that will be relayed to the buyer. */
+        data object FromFutureHtlc : PaymentType()
+        /** Fees will be deducted from future HTLCs that will be relayed to the buyer, but the preimage is revealed immediately. */
+        data object FromFutureHtlcWithPreimage : PaymentType()
```
pm47, line 152: I would completely drop that case.

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
```diff
@@ -20,140 +24,352 @@ import fr.acinq.lightning.utils.sat
...
+        /** Fees will be taken from the buyer's current fee credit. */
+        data object FromFeeCredit : PaymentType()
```
pm47, line 154: It should not be an either/or between `FromFeeCredit` on one hand, and `FromChannelBalance `/`FromFutureHtlc` on the other hand. Quite often the fee credit will only cover part of the channel management fee.

src/commonMain/kotlin/fr/acinq/lightning/channel/states/WaitForAcceptChannel.kt
```diff
@@ -52,40 +51,64 @@ data class WaitForAcceptChannel(
...
-                            when (val fundingContributions = FundingContributions.create(channelKeys, keyManager.swapInOnChainWallet, fundingParams, init.walletInputs)) {
+                            when (val lease = LiquidityAds.validateLease(
+                                lastSent.requestFunds,
+                                staticParams.remoteNodeId,
+                                channelId,
+                                fundingParams.fundingPubkeyScript(channelKeys),
+                                accept.fundingAmount,
+                                lastSent.fundingFeerate,
+                                accept.willFund
+                            )) {
```
pm47, line 62: Just for my understanding: adding this code here (and commit b5c187db6fc4079984838fc4b1eba9a2ef421a87 in general) does not change anything in terms of trust for the fee computation and overall funding amount from the point of view of the buyer right? Meaning that those were already computed in `fundingParams` and enforced by `InteractiveTx`?

pm47 (commented):
Review for commit b5c187db6fc4079984838fc4b1eba9a2ef421a87.

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -243,82 +219,192 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val db: IncomingPayment
...
+                                                )) {
+                                                    is LiquidityEvents.Rejected -> rejectOnTheFlyFunding(payment, incomingPayment, rejected)
+                                                    else -> {
+                                                        val actions = listOf(OpenOrSplicePayment(payment.amountReceived, requestedAmount, fundingRate, incomingPayment.preimage))
+                                                        val paymentOnlyHtlcs = payment.copy(
+                                                            // We need to splice before receiving the remaining HTLC parts.
+                                                            // We extend the duration of the MPP timeout to give more time for funding to complete.
+                                                            startedAtSeconds = payment.startedAtSeconds + 30,
+                                                            // We keep the currently added HTLCs, and should receive the remaining HTLCs after the open/splice.
+                                                            parts = htlcParts.toSet()
```
pm47, line 283: 👓  This is where we clean up the `WillAddHtlcPart`s and wait for the actual htlc parts to arrive.

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -243,82 +219,192 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val db: IncomingPayment
...
-                                    logger.info { "rejecting pay-to-open: reason=${rejected.reason}" }
-                                    nodeParams._nodeEvents.emit(rejected)
-                                    val actions = payment.parts.map { part ->
-                                        val failureMsg = TemporaryNodeFailure
-                                        when (part) {
-                                            is HtlcPart -> actionForFailureMessage(failureMsg, part.htlc)
-                                            is PayToOpenPart -> actionForPayToOpenFailure(privateKey, failureMsg, part.payToOpenRequest) // NB: this will fail all parts, we could only return one
+                            val htlcParts = payment.parts.filterIsInstance<HtlcPart>()
+                            val willAddHtlcParts = payment.parts.filterIsInstance<WillAddHtlcPart>()
+                            when {
```
pm47, line 236: I did not initially understand that, in the case where we are lacking inbound liquidity, we will enter this `when` twice:
- a first time in the `willAddHtlcParts.isNotEmpty()` branch, to *compute* the fee and ask for liquidity
- a second time in the `else` branch, to *verify* the fee and release the preimage

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -1241,18 +1171,238 @@ class Peer(
...
+                            else -> {
+                                val targetFeerate = when {
+                                    localBalance >= localMiningFee * 0.75 -> fundingFeerate
+                                    // Our current balance is too low to pay the on-chain fees.
+                                    // We will consume all of it in on-chain fees, and target a higher feerate.
+                                    // This ensures that the resulting feerate won't be too low compared to our target.
+                                    // We must cover the shared input and the shared output, which is a lot of weight, so we add 50%.
+                                    else -> fundingFeerate * 1.5
+                                }
```
pm47, line 1318: I don't get this

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -1075,7 +1068,17 @@ data class InteractiveTxSigningSession(
...
+            val liquidityFees = liquidityLease?.let { l ->
+                val fees = l.fees.total.toMilliSatoshi()
+                when (l.paymentDetails) {
+                    is LiquidityAds.PaymentDetails.FromChannelBalance -> if (fundingParams.isInitiator) fees else -fees
+                    // Fees will be paid later, from relayed HTLCs.
+                    is LiquidityAds.PaymentDetails.FromFutureHtlc -> 0.msat
+                    is LiquidityAds.PaymentDetails.FromFutureHtlcWithPreimage -> 0.msat
+                    // Fees are taken from the current fee credit.
+                    is LiquidityAds.PaymentDetails.FromFeeCredit -> 0.msat
+                }
```
pm47, line 1080: This is a criticial part for the `FromChannelBalance` case, which counter-intuitively happens after the interactive-tx is built (but before it is signed). The reason being that fees are taken from the commitment tx, not from the funding tx.

pm47 (commented):
Review for commit 22316a52f69663fd9578edf98e672399bdc2196b.

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -1178,10 +1181,10 @@ sealed class SpliceStatus {
         val localPushAmount: MilliSatoshi,
         val remotePushAmount: MilliSatoshi,
         val liquidityLease: LiquidityAds.Lease?,
-        val origins: List<Origin.PayToOpenOrigin>
+        val origins: List<Origin>
     ) : QuiescentSpliceStatus()
     /** The splice transaction has been negotiated, we're exchanging signatures. */
-    data class WaitingForSigs(val session: InteractiveTxSigningSession, val origins: List<Origin.PayToOpenOrigin>) : QuiescentSpliceStatus()
+    data class WaitingForSigs(val session: InteractiveTxSigningSession, val liquidityLease: LiquidityAds.Lease?, val origins: List<Origin>) : QuiescentSpliceStatus()
```
t-bast, line 1187: Note to self: see if we can get rid of the liquidity lease here and in `InProgress` to simplify serialization.

src/commonMain/kotlin/fr/acinq/lightning/channel/states/WaitForAcceptChannel.kt
t-bast: Exactly, this is used only when the wallet initiates funding because they have swap-in-potentiam inputs, so the fees are paid atomically during `interactive-tx` and don't require trust on any side.

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -1595,112 +1638,191 @@ data class OnionMessage(
...
- * @param paymentHash payment hash.
- * @param expireAt after the proposal expires, our peer will fail the payment and won't open a channel to us.
- * @param finalPacket onion packet that we would have received if there had been a channel to forward the payment to.
+ * This message is sent when an HTLC couldn't be relayed to our node because we don't have enough inbound liquidity.
+ * This allows us to treat it as an incoming payment, and request on-the-fly liquidity accordingly if we wish to receive that payment.
+ * If we accept the payment, we will send an [OpenDualFundedChannel] or [SpliceInit] message containing [ChannelTlv.RequestFundingTlv].
+ * Our peer will then provide the requested funding liquidity and will relay the corresponding HTLC(s) afterwards.
  */
-data class PayToOpenRequest(
+data class WillAddHtlc(
```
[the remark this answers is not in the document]
t-bast: From a spec point of view, it would be hacky to have a dummy `id` field. Also, what would you do with the tlv stream of the `update_add_htlc`? Just use an empty one/ignore it?

I'm also afraid it would add more confusion and we'd risk incorrectly confusing an `UpdateAddHtlc` actually coming from `WillAddHtlc` and a normal one, and creating bugs because of it...

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -1119,24 +1101,28 @@ class Peer(
...
+                                nodeParams._nodeEvents.emit(LiquidityEvents.Rejected(msg.amount, 0.msat, LiquidityEvents.Source.OffChainPayment, LiquidityEvents.Rejected.Reason.PolicySetToDisabled))
+                            }
+                            channelFundingIsInProgress() -> {
+                                // Once the channel funding is complete, we may have enough inbound liquidity to receive the payment without
+                                // an on-chain operation, which is more efficient. We thus reject that payment and wait for the sender to retry.
+                                logger.warning { "cannot accept on-the-fly funding: another funding attempt is already in-progress" }
+                                val failure = OutgoingPaymentPacket.buildWillAddHtlcFailure(nodeParams.nodePrivateKey, msg, TemporaryNodeFailure)
+                                input.send(SendOnTheFlyFundingMessage(failure))
+                                nodeParams._nodeEvents.emit(LiquidityEvents.Rejected(msg.amount, 0.msat, LiquidityEvents.Source.OffChainPayment, LiquidityEvents.Rejected.Reason.ChannelFundingInProgress))
+                            }
```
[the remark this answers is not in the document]
t-bast: I think in that case we could keep the `will_add_htlc` messages around, and replay them once the current splice is finished. This would create more risks around timeouts though, so we'll need to carefully think about this.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -38,22 +38,35 @@ import kotlin.time.Duration.Companion.seconds
...
+
+/**
+ * Initiate a channel open or a splice to allow receiving an off-chain payment.
+ *
+ * @param paymentAmount total amount of the off-chain payment (before fees are paid).
+ * @param requestedAmount requested inbound liquidity, which will allow receiving the off-chain payment.
+ * @param fundingLease funding rates applied by our peer for this amount.
+ * @param preimage preimage of the off-chain payment.
+ */
+data class OpenOrSplicePayment(val paymentAmount: MilliSatoshi, val requestedAmount: Satoshi, val fundingLease: LiquidityAds.FundingLease, val preimage: ByteVector32) : PeerCommand() {
```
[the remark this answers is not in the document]
t-bast: Exactly :)

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -1241,18 +1171,238 @@ class Peer(
...
+            }
+            is OpenOrSplicePayment -> {
+                val channel = channels.values.firstOrNull { it is Normal }
+                val currentFeerates = peerFeeratesFlow.filterNotNull().first()
+                val currentFeeCredit = feeCreditFlow.first().truncateToSatoshi()
+                when {
+                    channelFundingIsInProgress() -> {
+                        // Once the channel funding is complete, we may have enough inbound liquidity to receive the payment without an on-chain operation
+                        // This would be more economical, so we simply ignore the funding request.
+                        logger.warning { "ignoring on-the-fly funding, funding is already in progress" }
```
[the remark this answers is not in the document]
t-bast: We won't, in that case it will just timeout on the LSP side after 90 seconds.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -1241,18 +1171,238 @@ class Peer(
...
+                                val leaseFees = cmd.leaseFees(targetFeerate)
+                                val totalFees = TransactionFees(miningFee = localMiningFee.min(localBalance.truncateToSatoshi()) + leaseFees.miningFee, serviceFee = leaseFees.serviceFee)
+                                logger.info { "requesting on-the-fly splice for paymentHash=${cmd.paymentHash} feerate=$targetFeerate fee=${totalFees.total} paymentType=${paymentDetails.paymentType}" }
+                                val spliceCommand = ChannelCommand.Commitment.Splice.Request(
+                                    replyTo = CompletableDeferred(),
+                                    spliceIn = null,
+                                    spliceOut = null,
+                                    requestRemoteFunding = LiquidityAds.RequestFunds(cmd.requestedAmount, cmd.fundingLease, paymentDetails),
+                                    feerate = targetFeerate,
+                                    origins = listOf(Origin.OffChainPayment(cmd.preimage, cmd.paymentAmount, totalFees))
```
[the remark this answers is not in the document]
t-bast: > Isn't the channel origin kind of redundant?

I think the `paymentAmount` provided in `Origin.OffChainPayment` is useful, and cannot be included in the `payment_details`.

> I'm still not clear on how exactlt we deal with multiple parallel payments on the sender's side.

Can you detail what you mean by "multiple parallel payments on the sender's side"? Who is the sender here?

> I also assume there will be a timeout for replies to WillAddHtlc (even in the single-payment case).

Yes, that is specified in the gist, the LSP times out `will_add_htlc` if it doesn't receive an answer or a matching open/splice.

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
t-bast: Yes exactly, the first time triggers the open/splice, and then we will receive the `update_add_htlc` messages that couldn't initially be relayed (and were replaced by `will_add_htlc`) and go through the second step.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
t-bast: The wallet must pays on-chain fees during the `interactive-tx` session for its weight of the transaction (it must pay `localMiningFee`). Since we're not adding any additional on-chain inputs, the only way to pay those fees is from our channel balance. If our balance is too low, we will pay less fees than expected, so the resulting feerate of the transaction will be lower than our target. Since we cannot do anything about it (we don't have funds to pay for it), we instead tell the LSP to target a **higher** feerate:

- we won't be able to contribute enough, but we'll contribute everything we have
- the LSP will contribute more fees (because we asked for a higher feerate)
- so the resulting feerate shouldn't be too low

Does that make more sense?

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
pm47: > We would *not* create a `from_fee_credit` payment type and would instead implicitly use it

That is much closer to the current implementation

> unless the LSP always consumes the whole channel balance and then takes the rest from fee credit?

We need to spend first from the fee credit, so it always converges to zero.

> or the LSP sends some data in splice_ack to let the wallet know how much will be taken from fee credit and how much from channel balance?

The LSP could provide the current fee credit as a tlv in `splice_ack`. Is there a scenario where the fee credit would be updated while a splice is in progress?

> the LSP uses the fee credit in priority, and deduces the remaining amount from the future HTLCs

That is safer for the LSP than the non-fee-credit variant because the buyer has something at stake.

t-bast: SGTM, feel free to juste ignore the last commit for now, I'll extract it to a separate PR during my next rebase and will rework it to match that. But I'll focus on finalizing the eclair implementation of the on-the-fly funding part first, and will come back to fee credit afterwards.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -38,22 +38,35 @@ import kotlin.time.Duration.Companion.seconds
...
     val pushAmount: MilliSatoshi,
     val walletInputs: List<WalletState.Utxo>,
     val commitTxFeerate: FeeratePerKw,
     val fundingTxFeerate: FeeratePerKw,
-    val channelFlags: Byte,
     val channelType: ChannelType.SupportedChannelType
 ) : PeerCommand()
 
+/** Consume all the spendable utxos in the wallet state provided to open a channel or splice into an existing channel. */
+data class OpenOrSpliceChannel(val walletInputs: List<WalletState.Utxo>) : PeerCommand() {
```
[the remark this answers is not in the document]
t-bast: Renamed in the latest rebase.

src/commonMain/kotlin/fr/acinq/lightning/payment/LiquidityPolicy.kt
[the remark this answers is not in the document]
t-bast: I slightly changed the wording in the latest rebase to be more vague :)

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -137,9 +150,10 @@ class Peer(
     val client: IClient,
     val watcher: IWatcher,
     val db: Databases,
+    // TODO: once standardized, we should get this data from our peer's init message.
+    val remoteFundingRates: LiquidityAds.WillFundRates,
```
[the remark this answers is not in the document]
t-bast: Moved it to `WalletParams`, we can rename this `LSPParams` in a follow-up PR.

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
t-bast: Good point, done.

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
t-bast: I kept it for now since it's very few additional code, and people wanted to have both options in the spec. We can disable it by default though.

src/commonMain/kotlin/fr/acinq/lightning/wire/LiquidityAds.kt
t-bast: I moved fee credit to #660 and will re-work it once I'm done with the `eclair` side.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
t-bast: I've reworked the comment in the latest rebase, I hope it's clearer.

src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
t-bast: We can't easily get rid of this, otherwise we won't be able to store the lease in the DB, which is useful, even though it means we'll need to handle backwards-compatibility if the format changes...

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -1294,18 +1217,239 @@ class Peer(
...
+                                swapInCommands.trySend(SwapInCommand.UnlockWalletInputs(cmd.walletInputs.map { it.outPoint }.toSet()))
+                            }
+                            else -> {
+                                val spliceCommand = ChannelCommand.Commitment.Splice.Request(
+                                    replyTo = CompletableDeferred(),
+                                    spliceIn = ChannelCommand.Commitment.Splice.Request.SpliceIn(cmd.walletInputs),
+                                    spliceOut = null,
+                                    requestRemoteFunding = null,
+                                    feerate = feerate,
+                                    origins = listOf(Origin.OnChainWallet(cmd.walletInputs.map { it.outPoint }.toSet(), cmd.totalAmount.toMilliSatoshi(), ChannelManagementFees(fee, 0.sat)))
```
pm47, line 1252: The fee should be deducted from the amount, see: 2f8544c43ec975d52c43b36659aca6572313eb7e

src/commonMain/kotlin/fr/acinq/lightning/payment/LiquidityPolicy.kt
```diff
@@ -3,38 +3,47 @@ package fr.acinq.lightning.payment
...
             }
         }?.let { reason -> LiquidityEvents.Rejected(amount, fee, source, reason) }
     }
 
+    companion object {
+        /**
+         * We usually need our peer to contribute to channel funding, because they must have enough funds to pay the commitment fees.
+         * When we don't have an inbound liquidity target set, we use the following default amount.
+         */
+        val minInboundLiquidityTarget: Satoshi = 100_000.sat
```
pm47, line 47: Following yesterday's offline discussion, I think we can remove this?

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -1294,18 +1217,241 @@ class Peer(
...
+                            // This will be taken from our channel balance during the interactive-tx construction, they shouldn't be deducted from our funding amount.
+                            val totalFees = ChannelManagementFees(miningFee = localMiningFee + fundingFees.miningFee, serviceFee = fundingFees.serviceFee)
+                            Pair(localFundingAmount, totalFees)
+                        }
+                        if (cmd.totalAmount - fees.total < nodeParams.dustLimit) {
+                            logger.warning { "cannot create channel, not enough funds to pay fees (fees=${fees.total}, available=${cmd.totalAmount})" }
+                            swapInCommands.trySend(SwapInCommand.UnlockWalletInputs(cmd.walletInputs.map { it.outPoint }.toSet()))
+                        } else {
+                            when (val rejected = nodeParams.liquidityPolicy.first().maybeReject(requestRemoteFunding.requestedAmount.toMilliSatoshi(), fees.total.toMilliSatoshi(), LiquidityEvents.Source.OnChainWallet, logger)) {
```
pm47, line 1302: We must include the swapped amount in the liquidity policy check, like we do in the splice case.

```suggestion
                            when (val rejected = nodeParams.liquidityPolicy.first().maybeReject(cmd.walletInputs.balance.toMilliSatoshi() + requestRemoteFunding.requestedAmount.toMilliSatoshi(), fees.total.toMilliSatoshi(), LiquidityEvents.Source.OnChainWallet, logger)) {
```

pm47 (approved):

PR #650, opened 2024-05-24
Make `ChannelContext.process` a suspend function

pm47:
The first commit touches many files but there are no functional changes. Making `process()` a `suspend` function opens the way for the channel FSM to itself call `suspend` functions, but for now we don't call any.

In the next two commits we leverage the `suspend` function by adding a touch of asynchronicity for blockchain-related data: feerates and block height. The goal is to make the Peer much less dependent on blockchain data:

- Getting feerates asynchronously is straightforward and doesn't seem to have many side effects. Feerates are used at closing (mutual or force close). Instead of waiting for feerates to be available before processing any channel command, we just pass the `StateFlow` to the `ChannelContext` and will attempt to get feerates only when needed. If feerates are not available, the `ChannelState.process` function will hang. Previously, the hanging would have happened in the `Peer`.

- We then do the same for the blockheight (and also simplify by removing the unused `BlockHeader`). There are more side effects because we call `currentTipFlow.filterNotNull().first()` from several places in `Peer`. Some are addressed as PR comments  below.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -191,7 +191,7 @@ class Peer(
     private val ourInit = Init(features.initFeatures(), initTlvStream)
     private var theirInit: Init? = null
 
-    val currentTipFlow = MutableStateFlow<Pair<Int, BlockHeader>?>(null)
+    val currentTipFlow = MutableStateFlow<Int?>(null)
     val onChainFeeratesFlow = MutableStateFlow<OnChainFeerates?>(null)
     val swapInFeeratesFlow = MutableStateFlow<FeeratePerKw?>(null)
 
@@ -200,7 +200,7 @@ class Peer(
         val state = this
         val ctx = ChannelContext(
             StaticParams(nodeParams, remoteNodeId),
-            currentTipFlow.filterNotNull().first().first,
+            currentTipFlow,
             onChainFeeratesFlow,
             logger = MDCLogger(
                 logger = _channelLogger,
@@ -227,7 +227,7 @@ class Peer(
         launch {
             watcher.client.notifications.filterIsInstance<HeaderSubscriptionResponse>()
                 .collect { msg ->
-                    currentTipFlow.value = msg.blockHeight to msg.header
+                    currentTipFlow.value = msg.blockHeight
                 }
         }
         launch {
@@ -472,7 +472,7 @@ class Peer(
         waitForPeerReady()
         swapInJob = launch {
             swapInWallet.wallet.walletStateFlow
-                .combine(currentTipFlow.filterNotNull()) { walletState, currentTip -> Pair(walletState, currentTip.first) }
+                .combine(currentTipFlow.filterNotNull()) { walletState, currentTip -> Pair(walletState, currentTip) }
                 .combine(swapInFeeratesFlow.filterNotNull()) { (walletState, currentTip), feerate -> Triple(walletState, currentTip, feerate) }
                 .combine(nodeParams.liquidityPolicy) { (walletState, currentTip, feerate), policy -> TrySwapInFlow(currentTip, walletState, feerate, policy) }
                 .collect { w ->
@@ -617,7 +617,7 @@ class Peer(
             .filterIsInstance<Normal>()
             .firstOrNull()
             ?.let { channel ->
-                val leaseStart = currentTipFlow.filterNotNull().first().first
+                val leaseStart = currentTipFlow.filterNotNull().first()
                 val spliceCommand = ChannelCommand.Commitment.Splice.Request(
                     replyTo = CompletableDeferred(),
                     spliceIn = null,
@@ -717,7 +717,7 @@ class Peer(
 
                     is ChannelAction.ProcessCmdRes.AddSettledFail -> {
                         val currentTip = currentTipFlow.filterNotNull().first()
-                        when (val result = outgoingPaymentHandler.processAddSettled(actualChannelId, action, _channels, currentTip.first)) {
+                        when (val result = outgoingPaymentHandler.processAddSettled(actualChannelId, action, _channels, currentTip)) {
                             is OutgoingPaymentHandler.Progress -> {
                                 _eventsFlow.emit(PaymentProgress(result.request, result.fees))
                                 result.actions.forEach { input.send(it) }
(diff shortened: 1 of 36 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -717,7 +717,7 @@ class Peer(
 
                     is ChannelAction.ProcessCmdRes.AddSettledFail -> {
                         val currentTip = currentTipFlow.filterNotNull().first()
-                        when (val result = outgoingPaymentHandler.processAddSettled(actualChannelId, action, _channels, currentTip.first)) {
+                        when (val result = outgoingPaymentHandler.processAddSettled(actualChannelId, action, _channels, currentTip)) {
```
pm47, line 720: The blockheight is needed for auto-retrying payments when we have multiple `WalletParams.trampolineFees`, but we use a single flat fee for trampoline now. We should probably remove the support for multiple fees.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -839,7 +839,7 @@ class Peer(
     }
 
     private suspend fun processIncomingPayment(item: Either<PayToOpenRequest, UpdateAddHtlc>) {
-        val currentBlockHeight = currentTipFlow.filterNotNull().first().first
+        val currentBlockHeight = currentTipFlow.filterNotNull().first()
```
pm47, line 842: This one is risky because it is needed *after* having accepted and cross-signed an htlc. In a scenario where we cannot access the blockchain this could lead to a payment timeout. We should probably reject immediately if we don't have a blockheight, instead of hanging.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -1135,7 +1135,7 @@ class Peer(
                                 else -> null
                             }
                         }
-                        offerManager.receiveMessage(msg, remoteChannelUpdates, currentTipFlow.filterNotNull().first().first)?.let {
+                        offerManager.receiveMessage(msg, remoteChannelUpdates, currentTipFlow.filterNotNull().first())?.let {
```
pm47, line 1138: This one is a bit annoying, because one of the motivations for this PR is to make phoenix wake-up lighter in the context of responding to a BOLT12 invoice request by removing the dependency on the blockchain. Unfortunately we need  the current blockheight here:

https://github.com/ACINQ/lightning-kmp/blob/8b6ac2e906e11835a1bc0fa7978463b372e6af64/src/commonMain/kotlin/fr/acinq/lightning/payment/OfferManager.kt#L159

Is there any workaround here @thomash-acinq?

thomash-acinq: Yes, we could remove the whole `RouteBlindingEncryptedDataTlv.PaymentConstraints` TLV. It would deviate from the spec but it's fine because only the trampoline node would see it (we would need to modify the trampoline node to accept a missing `PaymentConstraints` in this case).

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
t-bast: Good point, and this should be quite easy to do, with something like:

```kotlin
when (val currentBlockHeight = currentTipFlow.value) {
    null -> // fail HTLC immediately
    else -> {
        // current processing
    }
}
```

But this could also negatively impact payment success: background payments may fail because we're simply slow to receive the block height, whereas if we waited a bit more on the `currentTipFlow` the background payment could have succeeded. And that's true for non-background payments as well when getting the block height is slow but eventually works.

If we decide to do this, it should probably be a separate follow-up PR with a dedicated unit test.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
t-bast: I don't think we would gain much by removing the dependency on the blockchain here, because we will need it anyway when receiving the HTLC as highlighted by [your other comment](https://github.com/ACINQ/lightning-kmp/pull/650/commits/8b6ac2e906e11835a1bc0fa7978463b372e6af64#r1620712981).

Since we need a block height when receiving the HTLC, which should come quickly after the `invoice_request`, the only thing we would save by not needing a block height here would potentially be a small latency boost is we're slower to get the block height than we are to connect to ACINQ and receive the onion message, but that's the only thing we'd gain. And it isn't latency that is visible to the user, since they're receiving a payment (potentially in the background), not sending one. So do we really care?

t-bast (commented):
It feels a bit dangerous to allow `ChannelState.process()` to suspend, because it could be abused to do a bunch of slow suspending operations which is something we wanted to avoid, but as long as we don't abuse it it should be ok.

Since `processEvent()` was already a `suspend` function that is only called in the `run()` loop, I don't think this would introduce weird concurrency issues, so concept ACK.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
pm47: We could use a 30s timeout here to handle the "laggy" scenario. But it's starting to look a bit ugly.

By the way, simply failing the htlc is already kind of ugly due to the htlc/pay-to-open cases. It's feasible but not as straightforward as it seems.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
pm47: Maybe I should just drop the blockheight commit altogether?

t-bast: A good first step seems indeed to be to only take the first two commits for now, while we work on a different solution to obtain the block height.

pm47:
> It feels a bit dangerous to allow ChannelState.process() to suspend, because it could be abused to do a bunch of slow suspending operations which is something we wanted to avoid, but as long as we don't abuse it it should be ok.

> Since processEvent() was already a suspend function that is only called in the run() loop, I don't think this would introduce weird concurrency issues, so concept ACK.

Agree, in fact it's such a significant change that it deserves its own PR (#653). Closing this one and splitting it in two.

PR #652, opened 2024-05-30
Identify as wallet node in blinded routes

thomash-acinq:
Adds a `OutgoingPhoenixId` in `RouteBlindingEncryptedData` to signal to the trampoline node that we are a phoenix node. This is not part of the spec as only our trampoline node will see it. It removes the need for `ShortChannelId.peerId`.


src/commonMain/kotlin/fr/acinq/lightning/payment/OfferManager.kt
```diff
@@ -187,10 +187,10 @@ class OfferManager(val nodeParams: NodeParams, val walletParams: WalletParams, v
         val needIntermediateHop = when (destination) {
             is Destination.BlindedPath -> when (val introduction = destination.route.introductionNodeId) {
                 is EncodedNodeId.Plain -> introduction.publicKey != remoteNodeId
-                is EncodedNodeId.ShortChannelIdDir -> true // we don't have access to the graph data and rely on our peer to resolve the scid
+                else -> true // we don't have access to the graph data and rely on our peer to resolve the scid
```
t-bast, line 190: Can you explicitly handle every case here instead of using `else`? This ensures that if we add another `EncodedNodeId` type in the future, the compiler will force us to think about what value to use here.

src/commonMain/kotlin/fr/acinq/lightning/wire/RouteBlinding.kt
```diff
@@ -10,6 +10,17 @@ import fr.acinq.bitcoin.utils.Either
...
+    /** Id of the next node. */
+    data class OutgoingPhoenixId(val nodeId: PublicKey) : RouteBlindingEncryptedDataTlv() {
+        override val tag: Long get() = OutgoingPhoenixId.tag
+        override fun write(out: Output) = LightningCodecs.writeBytes(nodeId.value, out)
+
+        companion object : TlvValueReader<OutgoingPhoenixId> {
+            const val tag: Long = 0
+            override fun read(input: Input): OutgoingPhoenixId = OutgoingPhoenixId(PublicKey(LightningCodecs.bytes(input, 33)))
+        }
+    }
```
t-bast, line 22: We can't use TLV type `0`, this is reserved for official TLVs from the BOLTs. If we introduce a separate TLV, we need to use a value greater than 65536, as detailed in [blip-0002](https://github.com/lightning/blips/blob/master/blip-0002.md#tlv-fields-in-bolt-messages).

But I think we can actually avoid introducing a new TLV here. Since you introduce this case as a new `EncodedNodeId`, why not do this consistently and just use the `OutgoingNodeId` TLV directly?

Right now there is a slightly weird asymmetry in `LightningCodecs.kt` where you handle the new `EncodedNodeId` when writing, but not when reading (and when you write it you lose the information that it's a phoenix `node_id`). We could instead keep this information when encoding:

- if the public key starts by `0x02`, replace that first byte by `0x04`
- if the public key starts by `0x03`, replace that first byte by `0x05` 
- when reading, it lets us detect that this is a phoenix `node_id` and transform it in the other direction (`0x04` -> `0x02` and `0x05` -> `0x03`)

src/commonMain/kotlin/fr/acinq/lightning/EncodedNodeId.kt
```diff
@@ -16,6 +16,10 @@ sealed class EncodedNodeId {
         override fun toString(): String = if (isNode1) "<-$scid" else "$scid->"
     }
 
+    data class PhoenixId(val publicKey: PublicKey) : EncodedNodeId() {
+        override fun toString(): String = publicKey.toString()
+    }
```
t-bast, line 21: We should avoid explicitly naming `Phoenix` and use a more generic concept that can later be specified in a bLIP and implemented by other wallets (at some point we'll introduce a bLIP feature for mobile wake-up, that implies being able to read this kind of `encoded_node_id`). What about simply `WalletNodeId`? And can you similarly rename the `isPhoenix` argument to `isMobileWallet` everywhere?

```suggestion
    /**
     * Wallet nodes are not part of the public graph, and may not have channels yet. 
     * Wallet providers are usually able to contact such nodes using push notifications or similar mechanisms.
     */
    data class WalletNodeId(val publicKey: PublicKey) : EncodedNodeId() {
        override fun toString(): String = publicKey.toString()
    }
```

src/commonMain/kotlin/fr/acinq/lightning/message/OnionMessages.kt
```diff
@@ -11,10 +11,13 @@ import fr.acinq.lightning.utils.toByteVector
 import fr.acinq.lightning.wire.*
 
 object OnionMessages {
-    data class IntermediateNode(val nodeId: PublicKey, val outgoingChannelId: ShortChannelId? = null, val padding: ByteVector? = null, val customTlvs: Set<GenericTlv> = setOf()) {
+    data class IntermediateNode(val nodeId: PublicKey, val isPhoenix: Boolean = false, val outgoingChannelId: ShortChannelId? = null, val padding: ByteVector? = null, val customTlvs: Set<GenericTlv> = setOf()) {
```
t-bast, line 14: It isn't this `IntermediateNode` that is phoenix, it's rather the next node? Since `toTlvStream` already takes an `EncodedNodeId` that contains this information, do we really need a new field here? Would it help if we changed `val nodeId: PublicKey` to `val nodeId: EncodedNodeId`? We would probably need to slightly change the `EncodedNodeId` class to be something like:

```
sealed class EncodedNodeId {
    sealed class WithPublicKey : EncodedNodeId() {
        abstract val publicKey: PublicKey

        data class Plain(override val publicKey: PublicKey) : WithPublicKey() {
            override fun toString(): String = publicKey.toString()
        }

        data class Wallet(override val publicKey: PublicKey) : WithPublicKey() {
            override fun toString(): String = publicKey.toString()
        }
    }

    /** For compactness, nodes may be identified by the shortChannelId of one of their public channels. */
    data class ShortChannelIdDir(val isNode1: Boolean, val scid: ShortChannelId) : EncodedNodeId() {
        override fun toString(): String = if (isNode1) "<-$scid" else "$scid->"
    }
}
```

And use `val nodeId: EncodedNodeId.WithPublicKey`. It could be cleaner than boolean flags that we need to propagate everywhere?

src/commonMain/kotlin/fr/acinq/lightning/message/OnionMessages.kt
```diff
@@ -24,13 +27,15 @@ object OnionMessages {
 
     sealed class Destination {
         data class BlindedPath(val route: RouteBlinding.BlindedRoute) : Destination()
-        data class Recipient(val nodeId: PublicKey, val pathId: ByteVector?, val padding: ByteVector? = null, val customTlvs: Set<GenericTlv> = setOf()) : Destination()
+        data class Recipient(val nodeId: PublicKey, val pathId: ByteVector?, val isPhoenix: Boolean, val padding: ByteVector? = null, val customTlvs: Set<GenericTlv> = setOf()) : Destination() {
```
t-bast, line 30: Same as the previous one, we could change `val nodeId: PublicKey` to directly take an `EncodedNodeId.WithPublicKey` instead of introducing a boolean flag?

t-bast (commented):
I think it's a good idea to explicitly provide this information in the encrypted data for the LSP, but I think we can do it with a better abstraction. I haven't tried implementing it all the way though, so I may be missing something, let me know.

src/commonMain/kotlin/fr/acinq/lightning/wire/RouteBlinding.kt
thomash-acinq: Very good idea.

src/commonMain/kotlin/fr/acinq/lightning/EncodedNodeId.kt
thomash-acinq: Done

src/commonMain/kotlin/fr/acinq/lightning/message/OnionMessages.kt
thomash-acinq: > It isn't this `IntermediateNode` that is phoenix, it's rather the next node?

No, it is about this node, not the next one. As you've noticed `toTlvStream` already takes this information for the next node.

> And use `val nodeId: EncodedNodeId.WithPublicKey`. It could be cleaner than boolean flags that we need to propagate everywhere?

Done, it's much nicer indeed.

src/commonMain/kotlin/fr/acinq/lightning/message/OnionMessages.kt
thomash-acinq: Done

src/commonMain/kotlin/fr/acinq/lightning/payment/OfferManager.kt
thomash-acinq: Done

t-bast (commented):
LGTM, this is quite clean! I've added a bit of documentation and formatting nits in #655 then it should be good to go :+1:

t-bast (approved):

PR #659, opened 2024-06-05
Refactor available channel selection

t-bast:
This PR contains several small preparatory commits for on-the-fly funding, and is best reviewed commit-by-commit.
The main change is the refactoring of the available channel selection process, in the last commit.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -541,14 +541,14 @@ class Peer(
      * Estimate the actual feerate to use (and corresponding fee to pay) in order to reach the target feerate
      * for a splice out, taking into account potential unconfirmed parent splices.
      */
-    suspend fun estimateFeeForSpliceOut(amount: Satoshi, scriptPubKey: ByteVector, targetFeerate: FeeratePerKw): Pair<FeeratePerKw, ChannelCommand.Commitment.Splice.Fees>? {
+    suspend fun estimateFeeForSpliceOut(amount: Satoshi, scriptPubKey: ByteVector, targetFeerate: FeeratePerKw): Pair<FeeratePerKw, ChannelManagementFees>? {
         return channels.values
             .filterIsInstance<Normal>()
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -708,6 +707,32 @@ class Peer(
...
+            else -> SelectChannelResult.NotReady
+        }
+        else -> when {
+            // We are currently funding a new channel, which should be available soon.
+            channels.values.any { it is WaitForOpenChannel || it is WaitForAcceptChannel || it is WaitForFundingCreated || it is WaitForFundingSigned } -> SelectChannelResult.NotReady
+            // We have funded a channel and are waiting for it to be ready.
+            channels.values.any { it is WaitForFundingConfirmed || it is WaitForChannelReady } -> SelectChannelResult.NotReady
+            // All channels are closing, we need a new one.
+            channels.values.all { it is ShuttingDown || it is Negotiating || it is Closing || it is Closed || it is Aborted } -> SelectChannelResult.None
+            // Note that we don't check for Offline or Syncing, because we wait for channel reestablishment already in [waitForPeerReady].
```
t-bast, line 731: I don't know if we should still check for `Offline` and `Syncing`, I'm afraid we'd risk hitting the bug mentioned in `waitForPeerReady` and would thus incorrectly return `SelectChannelResult.NotReady` and wait forever...or maybe we should use the same logic that `waitForPeerReady` contains, which would make `selectChannelForSplicing` more exhaustive?

t-bast: Done in https://github.com/ACINQ/lightning-kmp/pull/659/commits/d39bc01a8ce020799dd685e58c0d3bfe3228d8d1.

I'll try an alternative way of fixing this issue, by having `lightning-kmp` send its own `channel_reestablish` if it doesn't receive one after 10 seconds, this way the remote node can answer with an `error` and we will clean-up the channel.

t-bast: The syncing channels issue is fixed in #664

pm47 (approved):

PR #660, opened 2024-06-13
Add fee credit feature for on-the-fly funding

t-bast:
We add an optional feature that lets on-the-fly funding clients accept payments that are too small to pay the fees for an on-the-fly funding. When that happens, the payment amount is added as "fee credit" without performing an on-chain operation. Once enough fee credit has been obtained, we can initiate an on-chain operation to create a channel or a splice by paying the fees from our fee credit.

This feature makes more efficient use of on-chain transactions by trusting that our peer will honor our fee credit in the future.

:warning: Note that this PR needs to be reworked after discussions that took place in #649:

- we shouldn't introduce a new `payment_type`
- we should instead use `from_channel_balance` or `from_future_htlc(_with_preimage)`, but take funds from the current fee credit first
- the seller should tell the buyer how much fee credit will be used by using a custom TLV in `accept_channel2` / `splice_ack`

This PR builds on top of #649


src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -226,101 +202,220 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val db: IncomingPayment
...
+                                            // in terms of fees we need to pay, otherwise we may not have enough to actually pay the liquidity fees.
+                                            val maxFeerate = currentFeerate * AddLiquidityForIncomingPayment.SpliceWithNoBalanceFeerateRatio
+                                            val maxLiquidityFees = fundingRate.fees(maxFeerate, requestedAmount, requestedAmount, isChannelCreation = true).total.toMilliSatoshi()
+                                            val feeCreditThreshold = when (val policy = nodeParams.liquidityPolicy.value) {
+                                                LiquidityPolicy.Disable -> maxLiquidityFees
+                                                is LiquidityPolicy.Auto -> maxLiquidityFees.min(policy.maxAllowedFeeCredit)
+                                            }
+                                            val amountBelowThreshold = (willAddHtlcParts.map { it.amount }.sum() + currentFeeCredit) <= feeCreditThreshold
+                                            featureOk && amountBelowThreshold
```
pm47, line 255: I had a hard time understanding this, due to mixing the `maxLiquidityFees` and `maxAllowedFeeCredit` which are separate concerns. How about:

```suggestion
                                            val maxLiquidityFees = fundingRate.fees(maxFeerate, requestedAmount, requestedAmount, isChannelCreation = true).total.toMilliSatoshi()
                                            val maxFeeCredit = when (val policy = nodeParams.liquidityPolicy.value) {
                                                LiquidityPolicy.Disable -> 0.msat
                                                is LiquidityPolicy.Auto -> policy.maxAllowedFeeCredit
                                            }
                                            val nextFeeCredit = willAddHtlcParts.map { it.amount }.sum() + currentFeeCredit
                                            val cannotCoverLiquidityFees = nextFeeCredit <= maxLiquidityFees
                                            val isBelowMaxFeeCredit = nextFeeCredit <= maxFeeCredit
                                            val assessment = featureOk && cannotCoverLiquidityFees && isBelowMaxFeeCredit
                                            logger.info { "fee credit assessment: result=$assessment featureOk=$featureOk cannotCoverLiquidityFees=$cannotCoverLiquidityFees isBelowMaxFeeCredit=$isBelowMaxFeeCredit (nextFeeCredit=$nextFeeCredit,maxLiquidityFees=$maxLiquidityFees,maxFeeCredit=$maxFeeCredit)" }
                                            assessment
```

src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt
```diff
@@ -226,101 +202,220 @@ class IncomingPaymentHandler(val nodeParams: NodeParams, val db: IncomingPayment
...
+        val actions = payment.parts.map { part ->
+            when (part) {
+                is HtlcPart -> actionForFailureMessage(failure, part.htlc)
+                is WillAddHtlcPart -> actionForWillAddHtlcFailure(nodeParams.nodePrivateKey, failure, part.htlc)
+            }
+        }
+        return ProcessAddResult.Rejected(actions, incomingPayment)
+    }
+
+    private fun validateOnTheFlyFundingRate(
```
pm47, line 347: Can you explain again why we need to call `maybeReject` two times? 🙏

In the legacy fee-credit branch, [the semantics of `maybeReject` has been changed](https://github.com/ACINQ/lightning-kmp/blob/fee-credit-11/src/commonMain/kotlin/fr/acinq/lightning/payment/LiquidityPolicy.kt), making it return a `Decision` instead of a `Rejection`. Would it help the factorization?

pm47: Fair enough, we can always review later.

pm47 (approved):

PR #664, opened 2024-06-14
Unblock channels stuck in `Syncing`

t-bast:
When using channel backups, we wait for our peer to send their `channel_reestablish` first. This can be an issue for channels that are only present locally, and have been forgotten by our peer (for example because they force-close while we were offline).

In that case we would stay in `Syncing` indefinitely: to avoid that, we send our `channel_reestablish` if we haven't received theirs, which gives them an opportunity to send back an error and forget the channel.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -69,6 +69,7 @@ data class PeerConnection(val id: Long, val output: Channel<LightningMessage>, v
 }
 
 data class Connected(val peerConnection: PeerConnection) : PeerCommand()
+data class ReestablishPendingChannels(val connectionId: Long) : PeerCommand()
 data class MessageReceived(val connectionId: Long, val msg: LightningMessage) : PeerCommand()
 data class WatchReceived(val watch: WatchEvent) : PeerCommand()
 data class WrappedChannelCommand(val channelId: ByteVector32, val channelCommand: ChannelCommand) : PeerCommand()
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -963,6 +967,16 @@ class Peer(
...
+                                if (nodeParams.features.hasFeature(Feature.ChannelBackupClient)) {
+                                    // When using channel backups, we wait for our peer to send their channel_reestablish first.
+                                    // This can be an issue for channels that are only present locally, and have been forgotten by our peer.
+                                    // In that case we would stay in Syncing indefinitely: to avoid that, we send our channel_reestablish
+                                    // if we haven't received theirs, which gives them an opportunity to send back an error and forget the channel.
+                                    this.launch {
+                                        delay(10.seconds)
+                                        input.send(ReestablishPendingChannels(cmd.connectionId))
+                                    }
+                                }
```
pm47, line 979: This is a security measure though, so that the LSP doesn't know when we are behind. We may as well never wait?

t-bast: True, but what else can we do? We're missing information here, so our only chance is to send `channel_reestablish`. If we want to protect against malicious LSPs, we can force-close if they respond with `channel_reestablish` instead of sending an `error`: this way they can only try that trick once, so it's very unlikely to succeed?

pm47: I think the only proper fix is what I suggested in https://github.com/ACINQ/lightning-kmp/pull/659#discussion_r1636080389.

The alternative is drop this security measure entirely and always proactively send `channel_reestablish`. There is a case to be made that the client could be bluffing.

t-bast: Relying on a list of channels sent by our peer in `init` may be worse: they would have the incentive to omit channels from this list, knowing that we would forget them, then they can spend using a revoked close later. Unless you planned on force-closing all channels that are not in the list sent by our peer in `init`?

On the `eclair` side, we don't have information about the channels list when sending our `init`, that would need a refactor. It's conceptually a bit weird, because it's another mechanism to achieve the same thing that a standard `channel_reestablish` -> `error` exchange would do.

> The alternative is drop this security measure entirely and always proactively send channel_reestablish. There is a case to be made that the client could be bluffing.

In the majority of cases, it works to wait for the LSP to send `channel_reestablish` first and is nicer, why do you want to remove that just to accommodate very rare cases?

pm47: > Unless you planned on force-closing all channels that are not in the list sent by our peer in init?

Yes, this would indeed be required.

> In the majority of cases, it works to wait for the LSP to send channel_reestablish first and is nicer, why do you want to remove that just to accommodate very rare cases?

Because the only reason for not proactively sending the `channel_reestablish` is not letting know the LSP that we are potentially behind. If after 10s we send the `channel_reestablish` anyway, then the whole thing falls apart: a malicious LSP would just wait for 10s to have the peer show their cards. We may as well drop this feature and return to the default spec behavior.

t-bast:
We probably don't need this at all, since in the `WaitForFundingSigned` case we always send `channel_reestablish` first. In that case it's not a security issue since the commitment must be at index `0`.

The only other case where a channel may be stuck in the `Syncing` state is if the channel was force-closed while the wallet was offline, and the electrum node doesn't detect that force-close. But it should eventually unblock itself since the commit tx will confirm at some point.

PR #670, opened 2024-06-17
Add payer note

thomash-acinq:
Allows adding an optional payer note when paying an offer and save the payer note on the recipient's side.
Payer notes are truncated to 300 bytes so that they don't take too much space in the blinded route.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
@@ -155,7 +155,6 @@ data class NodeParams(
     val minFinalCltvExpiryDelta: CltvExpiryDelta,
     val maxFinalCltvExpiryDelta: CltvExpiryDelta,
     val bolt12invoiceExpiry: Duration,
-    val maxPayerNoteLength: Int,
 ) {
     val nodePrivateKey get() = keyManager.nodeKeys.nodeKey.privateKey
     val nodeId get() = keyManager.nodeKeys.nodeKey.publicKey
@@ -234,7 +233,6 @@ data class NodeParams(
         minFinalCltvExpiryDelta = Bolt11Invoice.DEFAULT_MIN_FINAL_EXPIRY_DELTA,
         maxFinalCltvExpiryDelta = CltvExpiryDelta(360),
         bolt12invoiceExpiry = 60.seconds,
-        maxPayerNoteLength = 300,
     )
 
     /**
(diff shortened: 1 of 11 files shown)


src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
```diff
@@ -154,7 +154,8 @@ data class NodeParams(
     val liquidityPolicy: MutableStateFlow<LiquidityPolicy>,
     val minFinalCltvExpiryDelta: CltvExpiryDelta,
     val maxFinalCltvExpiryDelta: CltvExpiryDelta,
-    val bolt12invoiceExpiry: Duration
+    val bolt12invoiceExpiry: Duration,
+    val maxPayerNoteLength: Int,
```
t-bast, line 158: I'm not sure it makes sense to let wallets pick a value here. Shouldn't we define a maximum value, expose it to the wallet and document it, and reject `PayOffer` that set a large payer note? Then wallets know that they should limit the textbox they display to users.

src/commonMain/kotlin/fr/acinq/lightning/payment/OfferManager.kt
```diff
@@ -150,7 +150,16 @@ class OfferManager(val nodeParams: NodeParams, val walletParams: WalletParams, v
...
-                val pathId = OfferPaymentMetadata.V1(ByteVector32(decrypted.pathId), amount, preimage, request.payerId, request.quantity, currentTimestampMillis()).toPathId(nodeParams.nodePrivateKey)
+                val truncatedPayerNote = request.payerNote?.let {
+                    val encoded = it.encodeToByteArray()
+                    if (encoded.size <= nodeParams.maxPayerNoteLength) {
+                        it
+                    } else {
+                        val charactersToKeep = encoded.take(nodeParams.maxPayerNoteLength - 3).toByteArray().decodeToString().length - 1
+                        it.take(charactersToKeep) + "…"
+                    }
+                }
```
t-bast, line 161: Wouldn't it be simpler to truncate the `String` directly, assuming the worst-case scenario where each character uses 4 bytes?

thomash-acinq: I now truncate to 64 characters

src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
thomash-acinq: I've hardcoded 64 characters as the limit. Wallets should probably limit the textbox, however I think that's purely a UI consideration and we shouldn't reject valid payment attempts.

t-bast (approved):
LGTM :+1:

src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
t-bast: Good point, it's the responsibility of the wallet to decide the number of characters to allow :+1:

PR #671, opened 2024-06-17
Expose private key for default offer

thomash-acinq:
This key can be used as a payer key when paying offers to serve as our new identifier (the old node id is not exposed at all with BOLT 12).

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/message/OnionMessages.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/message/OnionMessages.kt
@@ -52,6 +52,24 @@ object OnionMessages {
         }
     }
 
+    fun buildRouteToRecipient(
+        blindingSecret: PrivateKey,
+        intermediateNodes: List<IntermediateNode>,
+        recipient: Destination.Recipient
+    ): RouteBlinding. BlindedRouteDetails {
+        val intermediatePayloads = buildIntermediatePayloads(intermediateNodes, recipient.nodeId)
+        val tlvs = setOfNotNull(
+            recipient.padding?.let { RouteBlindingEncryptedDataTlv.Padding(it) },
+            recipient.pathId?.let { RouteBlindingEncryptedDataTlv.PathId(it) }
+        )
+        val lastPayload = RouteBlindingEncryptedData(TlvStream(tlvs, recipient.customTlvs)).write().toByteVector()
+        return RouteBlinding.create(
+            blindingSecret,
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/fr/acinq/lightning/message/OnionMessages.kt
```diff
@@ -52,24 +52,32 @@ object OnionMessages {
         }
     }
 
+    fun buildRouteToRecipient(
```
t-bast, line 55: Why don't you simply change `buildRoute` below to return a `RouteBlinding.BlindedRouteDetails`? It's a bit confusing to introduce yet another function that does just a subset of it?

thomash-acinq: That's because in case destination is another blinded route we don't have the information we need, we can only have it when we build the last payloads ourselves.

t-bast: But that's not visible in our codepaths though? On the caller side, we indeed only derive the private key when we create an offer with a `Recipient`, so it doesn't matter whether we call `buildRoute` or `buildRouteToRecipient`. The `BlindedRouteDetails` from `buildRoute` are still valid blinded route details, it's just that you shouldn't derive a key from it for the same use-cases if it's a blinded path?

thomash-acinq: I don't understand what's the `BlindedRouteDetails` from `buildRoute`, if `destination: Destination.BlindedPath`, we can't reconstruct the `BlindedRouteDetails`.

t-bast: I don't understand what you mean, `BlindedRouteDetails` are simply a blinded route with the blinding key also exposed. But it's not important, let's go with this and we can refactor later.

t-bast (approved):

src/commonMain/kotlin/fr/acinq/lightning/message/OnionMessages.kt
thomash-acinq: What I'm saying is that if we didn't build the full route ourselves, we will know the blinded route without knowing the blinding key for the last node. And `buildRoute` handles such cases so it can't return the blinding key for the last node in the general case. However in the subcase where we know the node id of the recipient and build the full route ourselves, then we know the blinding key of the last node and can return it, that's what I do in `buildRouteToRecipient`.

PR #672, opened 2024-06-17
Add `liquidity` field to `PayToOpenRequest`

pm47:
It is currently unused on `master` but we need it for compatibility with the fee-credit feature branch.

Also remove legacy unused fields.

cc @dpad85


t-bast (approved):
LGTM, I checked against eclair's feature branch :+1:

PR #673, opened 2024-06-17
Use standard `assertContentEquals` in tests

pm47:
Instead of custom `assertArrayEquals` which was not doing anything.

### DIFF
--- a/src/jvmTest/kotlin/fr/acinq/lightning/crypto/noise/NoiseTestsJvm.kt
+++ b/src/jvmTest/kotlin/fr/acinq/lightning/crypto/noise/NoiseTestsJvm.kt
@@ -1,10 +1,10 @@
 package fr.acinq.lightning.crypto.noise
 
-import fr.acinq.lightning.crypto.assertArrayEquals
 import fr.acinq.lightning.tests.utils.LightningTestSuite
 import fr.acinq.secp256k1.Hex
 import org.bouncycastle.crypto.ec.CustomNamedCurves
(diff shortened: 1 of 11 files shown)


src/jvmTest/kotlin/fr/acinq/lightning/crypto/noise/NoiseTestsJvm.kt
```diff
@@ -1,10 +1,10 @@
...
 
-import fr.acinq.lightning.crypto.assertArrayEquals
 import fr.acinq.lightning.tests.utils.LightningTestSuite
 import fr.acinq.secp256k1.Hex
 import org.bouncycastle.crypto.ec.CustomNamedCurves
 import org.junit.Test
+import kotlin.test.assertContentEquals
 
 @ExperimentalStdlibApi
 class NoiseTestsJvm : LightningTestSuite() {
```
pm47, line 10: @sstone those tests are failing

sstone: Ouch... these tests were never fully implemented, but passed  because we used a wrong assertion!! 
I've added what is missing in  https://github.com/ACINQ/lightning-kmp/pull/674.
There are no tests vectors for the curve that we're using (secp256k1) so we imported tests for a different curve (ed25519) for which we provide a simplistic implementation (used for testing only).

sstone (approved):

PR #676, opened 2024-06-20
Expose contact node ids for offers

thomash-acinq:
Expose a list of node ids for offers. Any node id in this list is the offer recipient.


t-bast (approved):
Nice!

PR #677, opened 2024-07-04
(Minor) Add an optional description to the default offer

pm47:
Some services (e.g. https://ocean.xyz/docs/lightning) require a specific description for the offer.

Default offers all use the same blinding secret, are are really the same offer, just with a different description.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/NodeParams.kt
@@ -240,7 +240,7 @@ data class NodeParams(
      * This offer will stay valid after restoring the seed on a different device.
      * @return the default offer and the private key that will sign invoices for this offer.
      */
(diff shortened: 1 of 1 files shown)


thomash-acinq:
This is not going to work as we check that the description matches later. I'll add a way to generate new offers.

pm47:
> This is not going to work as we check that the description matches later.

I don't see where the spec says that the invoice requestor must check that the description matches? https://github.com/lightning/bolts/blob/f150782106ac604b8bb056dcd0c0cbe757274e1d/12-offer-encoding.md#requirements-1

Or is it somewhere else in the payment process?

thomash-acinq:
In "Requirements for Invoice Requests":

> The reader:
>  - if `offer_node_id` or `offer_paths` are present (response to an offer):
>     - MUST fail the request if the offer fields do not exactly match a valid, unexpired offer.

pm47:
Right, so it's really a different offer after all.

PR #681, opened 2024-07-08
Ignore expired utxos migrated from the legacy app

t-bast:
When migrating from the legacy android app, we allowed unconfirmed utxos to be able to do an instantaneous migration. But we allowed *all* utxos coming from the legacy app instead of only allowing unconfirmed and weakly confirmed ones. That also included swap-in-potentiam utxos that expired (for example because they were too small to be economical), which are automatically rejected by the LSP because they can be double spent.

The result is that we are unable to complete swaps if we have an expired utxo coming from a migration from the legacy application. We fix this by only whitelisting unconfirmed and weakly confirmed utxos.

Note that we will remove the migration code entirely when we merge #649.

Fixes https://github.com/ACINQ/phoenix/issues/582


pm47 (approved):
Nice catch!

PR #683, opened 2024-07-09
Requesting BIP353 address from peer

pm47:
The peer will only answer if there is a channel.

Optionally pass an IETF BCP 47 language tag (en, fr, de, es, ...) to indicate preference for the words that make up the address.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -715,6 +715,11 @@ class Peer(
         peerConnection?.send(message)
     }
 
+    /**
+     * Request a BIP-353's compliant DNS address from our peer.
+     *
+     * @param languageSubtag IETF BCP 47 language tag (en, fr, de, es, ...) to indicate preference for the words that make up the address
+     */
     suspend fun requestAddress(languageSubtag: String): String {
         val replyTo = CompletableDeferred<String>()
         this.launch {
@@ -723,7 +728,7 @@ class Peer(
                 .first()
                 .let { event -> replyTo.complete(event.address) }
         }
-        peerConnection?.send(DNSAddressRequest(nodeParams.defaultOffer(walletParams.trampolineNode.id).first, languageSubtag))
+        peerConnection?.send(DNSAddressRequest(nodeParams.chainHash, nodeParams.defaultOffer(walletParams.trampolineNode.id).first, languageSubtag))
         return replyTo.await()
     }
 
@@ -1204,7 +1209,7 @@ class Peer(
                         }
                     }
                     is DNSAddressResponse -> {
-                        logger.info { "dns address assigned: ${msg}" }
+                        logger.info { "bip353 dns address assigned: ${msg.address}" }
                         _eventsFlow.emit(AddressAssigned(msg.address))
                     }
                 }
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -718,6 +718,8 @@ class Peer(
     /**
      * Request a BIP-353's compliant DNS address from our peer.
      *
+     * This will only return if there are existing channels with the peer, otherwise it will hang. This should be handled by the caller.
+     *
      * @param languageSubtag IETF BCP 47 language tag (en, fr, de, es, ...) to indicate preference for the words that make up the address
      */
     suspend fun requestAddress(languageSubtag: String): String {
--- a/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -12,6 +12,7 @@ import fr.acinq.lightning.blockchain.fee.FeeratePerKw
 import fr.acinq.lightning.blockchain.fee.OnChainFeerates
 import fr.acinq.lightning.blockchain.mempool.MempoolSpaceClient
 import fr.acinq.lightning.channel.*
+import fr.acinq.lightning.channel.ChannelCommand.Commitment.Splice.Response
 import fr.acinq.lightning.channel.states.*
 import fr.acinq.lightning.crypto.noise.*
 import fr.acinq.lightning.db.*
@@ -116,6 +117,8 @@ data class ChannelClosing(val channelId: ByteVector32) : PeerEvent()
  * Useful to handle transparent migration on Phoenix Android between eclair-core and lightning-kmp.
  */
 data class PhoenixAndroidLegacyInfoEvent(val info: PhoenixAndroidLegacyInfo) : PeerEvent()
+
+data class AddressAssigned(val address: String) : PeerEvent()
 
 /**
  * The peer we establish a connection to. This object contains the TCP socket, a flow of the channels with that peer, and watches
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -1747,6 +1751,50 @@ data class PhoenixAndroidLegacyInfo(
...
+    override fun write(out: Output) {
+        val serializedOffer = offer.encode()
+        LightningCodecs.writeU16(serializedOffer.length, out)
+        LightningCodecs.writeBytes(serializedOffer.toByteArray(charset = Charsets.UTF_8), out)
+        LightningCodecs.writeU16(languageSubtype.length, out)
+        LightningCodecs.writeBytes(languageSubtype.toByteArray(charset = Charsets.UTF_8), out)
+    }
+
+    companion object : LightningMessageReader<DNSAddressRequest> {
+        const val type: Long = 35025
```
t-bast, line 1767: This value is also used by on-the-fly funding for the `recommended_feerates` message (in https://github.com/ACINQ/lightning-kmp/pull/660). I should change that, and use a value that's further from the current extension messages we used that aren't specified.

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -1747,6 +1751,50 @@ data class PhoenixAndroidLegacyInfo(
     }
 }
 
+data class DNSAddressRequest(val offer: OfferTypes.Offer, val languageSubtype: String) : LightningMessage {
```
t-bast, line 1754: The name can be confusing, so it's worth some documentation on that message to explain what it does? Also, I believe this should have a `chainHash` and extend `HasChainHash`.

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -1747,6 +1751,50 @@ data class PhoenixAndroidLegacyInfo(
...
+
+    override val type: Long get() = DNSAddressRequest.type
+
+    override fun write(out: Output) {
+        val serializedOffer = offer.encode()
+        LightningCodecs.writeU16(serializedOffer.length, out)
+        LightningCodecs.writeBytes(serializedOffer.toByteArray(charset = Charsets.UTF_8), out)
```
t-bast, line 1761: It's a bit weird to use a string encoding and then convert to bytes when we actually have a byte encoding for offers (and `encode` internally first encodes to bytes before going to a string representation using bech32m)...an offer is just a TLV stream, and we have the `Offer.tlvSerializer` available to directly get bytes?

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -1188,6 +1203,10 @@ class Peer(
                             }
                         }
                     }
+                    is DNSAddressResponse -> {
+                        logger.info { "dns address assigned: ${msg}" }
```
t-bast, line 1207: nit:

```suggestion
                        logger.info { "bip353 dns address assigned: ${msg.address}" }
```

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -712,6 +715,18 @@ class Peer(
         peerConnection?.send(message)
     }
 
+    suspend fun requestAddress(languageSubtag: String): String {
```
t-bast, line 718: This will suspend forever if there is no channel and the peer doesn't answer, we should at least document it to make sure the app correctly handles this? I think it's important to make sure we seamlessly handle the case where the peer doesn't answer, because that will allow us to update the protocol in the future and simply stop responding to those requests once we want to make replace them with a new protocol?

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
pm47: I can totally rename to `Bip353DNSAddressRequest` if you prefer?

t-bast: That's not really necessary, but I think it's important to document what this does, and that the peer may simply not respond.

pm47: Done in ad3b2efa.

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
pm47: Great point, done in ad3b2efa.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
pm47: Done in ad3b2efa.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
dpad85: Current implementation on Phoenix uses a 5s timeout, we can work with that or move the timeout into lightning-kmp.

pm47: Documented the behavior in 146ffd4b.

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -1747,6 +1751,59 @@ data class PhoenixAndroidLegacyInfo(
...
+        LightningCodecs.writeBytes(languageSubtag.toByteArray(charset = Charsets.UTF_8), out)
+    }
+
+    companion object : LightningMessageReader<DNSAddressRequest> {
+        const val type: Long = 35025
+
+        override fun read(input: Input): DNSAddressRequest {
+            return DNSAddressRequest(
+                chainHash = BlockHash(LightningCodecs.bytes(input, 32)),
+                offer = OfferTypes.Offer.decode(LightningCodecs.bytes(input, LightningCodecs.u16(input)).decodeToString()).get(),
```
t-bast, line 1778: This won't work, it's still using the string encoding! I find that it's *always* useful to add a basic encode/decode unit test in `LightningCodecTestsCommon`, that's a good safety net and makes it easier to implement the same message in eclair.

src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
t-bast: > Current implementation on Phoenix uses a 5s timeout, we can work with that or move the timeout into lightning-kmp.

I think it makes sense to have a timeout on the Phoenix side here. This way, if we make this message obsolete at some point, old versions of the app will just go through the timeout failure path.

src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
pm47: Damn! I already have an integration test with eclair, but this isn't in the code path :-/

pm47: Fixed in d1daf8de.

PR #685, opened 2024-07-12
Remove deprecated serialization format and fields

t-bast:
We do some clean-up of deprecated fields, classes and serialization formats. This is a breaking change for users who have never launched the app since the [v1.5.0 release](https://github.com/ACINQ/lightning-kmp/releases/tag/v1.5.0) (from more than 1 year ago). It is not possible to upgrade directly from a version < 1.5.0, an intermediate update will be required.

NB: builds on top of #660


modules/core/src/commonMain/kotlin/fr/acinq/lightning/crypto/KeyManager.kt
```diff
@@ -41,11 +41,7 @@ interface KeyManager {
      * Keys used for the node. They are used to generate the node id, to secure communication with other peers, and
      * to sign network-wide public announcements.
      */
-    data class NodeKeys(
-        /** The node key that the same seed would have produced on the legacy eclair-based Phoenix implementation on Android. Useful to automate the migration. */
-        val legacyNodeKey: DeterministicWallet.ExtendedPrivateKey,
-        val nodeKey: DeterministicWallet.ExtendedPrivateKey,
-    )
+    data class NodeKeys(val nodeKey: DeterministicWallet.ExtendedPrivateKey)
```
pm47: `legacyNodeKey` needs to be kept to facilitate user support

modules/core/src/commonMain/kotlin/fr/acinq/lightning/Features.kt
```diff
@@ -154,17 +154,6 @@ sealed class Feature {
         override val scopes: Set<FeatureScope> get() = setOf(FeatureScope.Invoice)
     }
 
-    // We historically used the following feature bit in our invoices.
-    // However, the spec assigned the same feature bit to `option_scid_alias` (https://github.com/lightning/bolts/pull/910).
-    // We're moving this feature bit to 148, but we have to keep supporting it until enough wallet users have migrated, then we can remove it.
-    // We cannot rename that object otherwise we will not be able to read old serialized data.
-    @Serializable
-    object TrampolinePayment : Feature() {
```
pm47, line 162: I verified that:
- those features are not permanent and thus haven't been written in `channelFeatures`
- the init features written to `localParams.features` were always defined as `optional`

modules/core/src/commonMain/kotlin/fr/acinq/lightning/serialization/channel/Serialization.kt
```diff
@@ -17,9 +16,6 @@ object Serialization {
         return when {
             // v4 uses a 1-byte version discriminator
             bin[0].toInt() == 4 -> DeserializationResult.Success(fr.acinq.lightning.serialization.channel.v4.Deserialization.deserialize(bin))
-            // v2/v3 use a 4-bytes version discriminator
-            Pack.int32BE(bin) == 3 -> DeserializationResult.Success(fr.acinq.lightning.serialization.v3.Serialization.deserialize(bin))
-            Pack.int32BE(bin) == 2 -> DeserializationResult.Success(fr.acinq.lightning.serialization.v2.Serialization.deserialize(bin))
```
pm47, line 22: This may save future headaches (otherwise it will return version 0):

```kotlin
            // v2/v3 used a 4-bytes version discriminator and are now unsupported
            Pack.int32BE(bin) == 3 -> DeserializationResult.UnknownVersion(3)
            Pack.int32BE(bin) == 2 -> DeserializationResult.UnknownVersion(2)
```

pm47 (approved):
Nice cleanup!

PR #686, opened 2024-07-15
Prune route hints and blinded paths in trampoline onion

thomash-acinq:
When using trampoline, the route hints or blinded paths may not all fit inside the onion. If that's the case, we prune some of them until it fits.


src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentPacket.kt
```diff
@@ -81,8 +82,20 @@ object OutgoingPaymentPacket {
             }
             Triple(amount + hop.fee(amount), expiry + hop.cltvExpiryDelta, listOf(payload) + payloads)
         }
+        var payloads = initialPayloads
         val nodes = hops.map { it.nextNodeId }
-        val onion = buildOnion(nodes, payloads, invoice.paymentHash, payloadLength = null)
+        var onion = buildOnion(nodes, payloads, invoice.paymentHash, payloadLength = null)
+        // Ensure that this onion can fit inside the outer 1300 bytes onion.
+        while (onion.packet.payload.size() > 1142) {
```
t-bast, line 89: Where does 1142 come from? We probably don't want to explicitly detail each byte (especially since there is a risk that it becomes outdated if something changes in the outer onion), so we should probably:

- take some safety margin to keep some buffer for future additional data and restrict to 1000 bytes
- add a comment explaining that the outer onion fields needs ~150 bytes and we're leaving some buffer space available

src/commonTest/kotlin/fr/acinq/lightning/payment/PaymentPacketTestsCommon.kt
```diff
@@ -657,5 +657,12 @@ class PaymentPacketTestsCommon : LightningTestSuite() {
...
+        val (_, _, trampolineOnion) = OutgoingPaymentPacket.buildTrampolineToNonTrampolinePacket(invoice, NodeHop(randomKey().publicKey(), randomKey().publicKey(), CltvExpiryDelta(444), 0.msat), finalAmount, finalExpiry)
+        assertTrue(trampolineOnion.packet.payload.size() < 1142)
```
t-bast, line 665: We should rather build the final payment onion that contains the trampoline onion to prove that it has the correct size?

t-bast (commented):
Thanks!

src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentPacket.kt
thomash-acinq: Done

src/commonTest/kotlin/fr/acinq/lightning/payment/PaymentPacketTestsCommon.kt
thomash-acinq: Done

t-bast (approved):

PR #687, opened 2024-07-15
Handle feature-less BOLT 11 invoices

pm47:
We were not handling correctly feature-less invoices, which are spec-compliant. We do reject invoices that don't have `var_onion_optin`, but the check happens later. Instead, we currently throw a `NullPointerException`.

The way we are checking requirements in the `init` block requires lazy-init properties, otherwise the object initialization will fail before any logic can happen.

Also, as per the spec, the feature tag must be skipped at serialization if there are no features enabled.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/Bolt11Invoice.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/Bolt11Invoice.kt
@@ -21,25 +21,25 @@ data class Bolt11Invoice(
     val tags: List<TaggedField>,
     val signature: ByteVector
 ) : PaymentRequest() {
-    val chain: Chain? = prefixes.entries.firstOrNull { it.value == prefix }?.key
+    val chain: Chain? get() = prefixes.entries.firstOrNull { it.value == prefix }?.key
 
-    override val paymentHash: ByteVector32 = tags.find { it is TaggedField.PaymentHash }!!.run { (this as TaggedField.PaymentHash).hash }
+    override val paymentHash: ByteVector32 get() = tags.find { it is TaggedField.PaymentHash }!!.run { (this as TaggedField.PaymentHash).hash }
 
-    val paymentSecret: ByteVector32 = tags.find { it is TaggedField.PaymentSecret }!!.run { (this as TaggedField.PaymentSecret).secret }
+    val paymentSecret: ByteVector32 get() = tags.find { it is TaggedField.PaymentSecret }!!.run { (this as TaggedField.PaymentSecret).secret }
 
-    val paymentMetadata: ByteVector? = tags.find { it is TaggedField.PaymentMetadata }?.run { (this as TaggedField.PaymentMetadata).data }
+    val paymentMetadata: ByteVector? get() = tags.find { it is TaggedField.PaymentMetadata }?.run { (this as TaggedField.PaymentMetadata).data }
 
-    val description: String? = tags.find { it is TaggedField.Description }?.run { (this as TaggedField.Description).description }
+    val description: String? get() = tags.find { it is TaggedField.Description }?.run { (this as TaggedField.Description).description }
 
-    val descriptionHash: ByteVector32? = tags.find { it is TaggedField.DescriptionHash }?.run { (this as TaggedField.DescriptionHash).hash }
+    val descriptionHash: ByteVector32? get() = tags.find { it is TaggedField.DescriptionHash }?.run { (this as TaggedField.DescriptionHash).hash }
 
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/fr/acinq/lightning/payment/Bolt11Invoice.kt
```diff
@@ -21,25 +21,25 @@ data class Bolt11Invoice(
...
-    val expirySeconds: Long? = tags.find { it is TaggedField.Expiry }?.run { (this as TaggedField.Expiry).expirySeconds }
+    val expirySeconds: Long? get() = tags.find { it is TaggedField.Expiry }?.run { (this as TaggedField.Expiry).expirySeconds }
 
-    val minFinalExpiryDelta: CltvExpiryDelta? = tags.find { it is TaggedField.MinFinalCltvExpiry }?.run { CltvExpiryDelta((this as TaggedField.MinFinalCltvExpiry).cltvExpiry.toInt()) }
+    val minFinalExpiryDelta: CltvExpiryDelta? get() = tags.find { it is TaggedField.MinFinalCltvExpiry }?.run { CltvExpiryDelta((this as TaggedField.MinFinalCltvExpiry).cltvExpiry.toInt()) }
 
     val fallbackAddress: String? = tags.find { it is TaggedField.FallbackAddress }?.run { (this as TaggedField.FallbackAddress).toAddress(prefix) }
 
-    override val features: Features = tags.find { it is TaggedField.Features }.run { Features((this as TaggedField.Features).bits) }
+    override val features: Features get() = tags.filterIsInstance<TaggedField.Features>().firstOrNull()?.run { Features(this.bits) } ?: Features.empty
```
pm47, line 42: This is the main fix.

src/commonMain/kotlin/fr/acinq/lightning/payment/Bolt11Invoice.kt
```diff
@@ -55,7 +55,7 @@ data class Bolt11Invoice(
         require(description != null || descriptionHash != null) { "there must be exactly one description tag or one description hash tag" }
     }
 
-    override fun isExpired(currentTimestampSeconds: Long): Boolean = when (expirySeconds) {
+    override fun isExpired(currentTimestampSeconds: Long): Boolean = when (val expirySeconds = expirySeconds) {
```
pm47, line 58: Unfortunately we can't smart cast lazy properties.

src/commonMain/kotlin/fr/acinq/lightning/payment/Bolt11Invoice.kt
```diff
@@ -65,14 +65,16 @@ data class Bolt11Invoice(
...
-        tags.forEach {
-            val encoded = it.encode()
-            val len = encoded.size
-            data5.add(it.tag)
-            data5.add((len / 32).toByte())
-            data5.add((len.rem(32)).toByte())
-            data5.addAll(encoded)
-        }
+        tags
+            .filterNot { it is TaggedField.Features && it.bits.isEmpty() }
```
pm47, line 69: Skip the features tag entirely when they are no enabled features.

src/commonTest/kotlin/fr/acinq/lightning/payment/Bolt11InvoiceTestsCommon.kt
```diff
@@ -498,10 +499,12 @@ class Bolt11InvoiceTestsCommon : LightningTestSuite() {
         assertEquals(pr1.paymentSecret, pr.paymentSecret)
 
         // An invoice without the payment secret feature should be rejected
-        assertTrue(Bolt11Invoice.read("lnbc40n1pw9qjvwpp5qq3w2ln6krepcslqszkrsfzwy49y0407hvks30ec6pu9s07jur3sdpstfshq5n9v9jzucm0d5s8vmm5v5s8qmmnwssyj3p6yqenwdencqzysxqrrss7ju0s4dwx6w8a95a9p2xc5vudl09gjl0w2n02sjrvffde632nxwh2l4w35nqepj4j5njhh4z65wyfc724yj6dn9wajvajfn5j7em6wsq2elakl").isFailure)
+        assertIs<Try.Failure<Throwable>>((Bolt11Invoice.read("lnbc40n1pw9qjvwpp5qq3w2ln6krepcslqszkrsfzwy49y0407hvks30ec6pu9s07jur3sdpstfshq5n9v9jzucm0d5s8vmm5v5s8qmmnwssyj3p6yqenwdencqzysxqrrss7ju0s4dwx6w8a95a9p2xc5vudl09gjl0w2n02sjrvffde632nxwh2l4w35nqepj4j5njhh4z65wyfc724yj6dn9wajvajfn5j7em6wsq2elakl")))
```
pm47, line 502: This test and the following one were actually throwing a `NullPointerException`.

PR #689, opened 2024-07-18
Unlock swap-in utxos if initial channel creation fails

### LINKED ISSUE #680: Unlock swap-in utxos if initial channel creation fails
When swap inputs are deeply confirmed, we initiate an open_channel or a splice to import those funds into a lightning channel. The first step is to "reserve" those utxos in the `SwapInManager` to avoid double-spending ourselves.

If the funding attempt fails (most of the time because we get disconnected during the funding process), we need to free those utxos to allow re-using them when reconnecting. This is correctly done for splicing, using the `replyTo` field on the `Splice.Request` command, but this isn't implemented for the initial channel creation. For simplicity, we're waiting for the user to re-open the app before re-attempting the swap, but if the app stays in the background and is never really stopped, it may take a few hours/days before the swap is attempted again if the user doesn't actively kill the app.

We should have a mechanism similar to the splice's `replyTo` for the initial channel creation (at least for the disconnection case) to allow faster seamless retries.

t-bast:
Add a `replyTo` field when opening or accepting a channel. This is used for swaps, to free utxos in case a failure happens during funding. Without this mechanism, the user needs to restart the app to be able to reuse those utxos.

Fixes #680

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/ChannelCommand.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/ChannelCommand.kt
@@ -25,6 +25,7 @@ sealed class ChannelCommand {
     data object Disconnected : ChannelCommand()
     sealed class Init : ChannelCommand() {
         data class Initiator(
+            val replyTo: CompletableDeferred<ChannelFundingResponse>,
             val fundingAmount: Satoshi,
             val pushAmount: MilliSatoshi,
             val walletInputs: List<WalletState.Utxo>,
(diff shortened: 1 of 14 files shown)


pm47 (approved):

pm47:
@dpad85 `Peer.spliceOut/spliceCpfp/requestInboundLiquidity` now return a `ChannelFundingResponse` instead of `ChannelCommand.Commitment.Splice.Response`. It's a simple rename, no functional changes.

PR #693, opened 2024-08-21
Increase CLTV delta for Bolt12 invoices

thomash-acinq:
Take into account `minFinalCltvExpiryDelta` in the blinded route CLTV delta.


vincenzopalazzo (approved):
LGTM, thanks!

PR #694, opened 2024-08-21
Include `min_final_cltv_expiry_delta` in blinded paths

t-bast:
If a payer uses the current block height as the expiry for the payments they send to us, we will reject it because we need to have a few blocks to potentially force-close and get our funds back safely.

With blinded paths, we don't need to rely on payers to add a safety margin: we can directly encode it in the blinded path's total expiry delta.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/payment/OfferManager.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/payment/OfferManager.kt
@@ -132,6 +132,8 @@ class OfferManager(val nodeParams: NodeParams, val walletParams: WalletParams, v
     private fun receiveInvoiceRequest(decrypted: OnionMessages.DecryptedMessage, remoteChannelUpdates: List<ChannelUpdate>, currentBlockHeight: Int): OnionMessageAction.SendMessage? {
         val offer = localOffers[decrypted.pathId]!!
         val request = decrypted.content.records.get<OnionMessagePayloadTlv.InvoiceRequest>()?.let { OfferTypes.InvoiceRequest.validate(it.tlvs).right }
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/fr/acinq/lightning/payment/OfferManager.kt
```diff
@@ -159,18 +161,25 @@ class OfferManager(val nodeParams: NodeParams, val walletParams: WalletParams, v
...
-                    minHtlc = remoteChannelUpdates.minOfOrNull { it.htlcMinimumMsat } ?: 1.msat,
-                    maxHtlc = amount,
+                    // We include our min_final_cltv_expiry_delta in the path, but we *don't* include it in the payment_relay field
+                    // for our trampoline node (below). This ensures that we will receive payments with at least this final expiry delta.
+                    // This ensures that even when payers haven't received the latest block(s) or don't include a safety margin in the
+                    // expiry they use, we can still safely receive their payment.
+                    cltvExpiryDelta = cltvExpiryDelta + nodeParams.minFinalCltvExpiryDelta,
+                    // We must use the most restrictive minimum HTLC value between local and remote.
+                    minHtlc = (listOf(nodeParams.htlcMinimum) + remoteChannelUpdates.map { it.htlcMinimumMsat }).max(),
+                    // Payments are allowed to overpay at most two times the invoice amount.
```
thomash-acinq, line 175: That makes sense for BOLT11 but for BOLT12 the payer already sets the amount in the invoice request.

thomash-acinq (approved):

thomash-acinq:
Should we fail the invoice request if the amount is too low?
https://github.com/ACINQ/lightning-kmp/pull/696

thomash-acinq (approved):

PR #695, opened 2024-08-22
Use bitcoin-kmp 0.20.0

sstone:
Includes support for testnet4.


pm47:
Tested against eclair, are there any blockers?

PR #697, opened 2024-09-09
Remove support for `push_amount`

t-bast:
With the updated on-the-fly funding protocol, we're using liquidity ads instead of an unofficial `push_amount` TLV. We can thus remove usage of `push_amount`s everywhere, which simplifies the codebase.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
@@ -66,6 +66,14 @@ object Deserialization {
         lastSent = readLightningMessage() as ChannelReady
     )
 
+    private fun Input.readWaitForFundingSigned() = WaitForFundingSigned(
+        channelParams = readChannelParams(),
+        signingSession = readInteractiveTxSigningSession(),
+        remoteSecondPerCommitmentPoint = readPublicKey(),
+        liquidityPurchase = readNullable { readLiquidityPurchase() },
+        channelOrigin = readNullable { readChannelOrigin() }
+    )
+
     private fun Input.readWaitForFundingSignedWithPushAmount(): WaitForFundingSigned {
         val channelParams = readChannelParams()
         val signingSession = readInteractiveTxSigningSession()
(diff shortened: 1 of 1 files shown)


src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
```diff
@@ -64,30 +66,54 @@ object Deserialization {
         lastSent = readLightningMessage() as ChannelReady
     )
 
+    private fun Input.readWaitForFundingSignedWithPushAmount(): WaitForFundingSigned {
+        val channelParams = readChannelParams()
+        val signingSession = readInteractiveTxSigningSession()
+        // We previously included a local_push_amount and a remote_push_amount.
+        readNumber()
+        readNumber()
```
pm47, line 74: At this state, the initial `LocalCommit`/`RemoteCommit` objects were already fully generated, and we don't need local/remote `push_msat` to validate signatures.

We would only need local/remote `push_msat` if we were to do a rbf, which we are not doing in Phoenix due to zeroconf.

So we can safely discard those values.

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
```diff
@@ -64,30 +66,54 @@ object Deserialization {
         lastSent = readLightningMessage() as ChannelReady
     )
 
+    private fun Input.readWaitForFundingSignedWithPushAmount(): WaitForFundingSigned {
```
pm47, line 69: Nit: the ordering is weird, I think it should be:
1. `readWaitForFundingSigned` 
2. `readWaitForFundingSignedWithPushAmount` 
3. `readWaitForFundingSignedLegacy`

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
```diff
@@ -64,30 +66,54 @@ object Deserialization {
...
+        val remoteSecondPerCommitmentPoint = readPublicKey()
+        val channelOrigin = readNullable { readChannelOrigin() }
+        return WaitForFundingSigned(channelParams, signingSession, remoteSecondPerCommitmentPoint, liquidityPurchase = null, channelOrigin)
+    }
+
+    private fun Input.readWaitForFundingConfirmedWithPushAmount(): WaitForFundingConfirmed {
+        val commitments = readCommitments()
+        // We previously included a local_push_amount and a remote_push_amount.
+        readNumber()
+        readNumber()
```
pm47, line 104: Same remark as above, those are safe to discard.

pm47 (approved):
LGTM

Tested against eclair

src/commonMain/kotlin/fr/acinq/lightning/serialization/v4/Deserialization.kt
pm47: Those are just review notes confirming that it is okay to discard those two fields after reading, which is not obvious. In particular it wasn't obvious that we wouldn't need this information to validate the signatures received in that state.

pm47 (approved):

PR #701, opened 2024-09-17
Upgrade to kotlin 2.0.20

pm47:
And other dependencies.

We are not using the new `kotlin.Uuid`, as they are still experimental.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,11 +1,13 @@
+import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
 import org.jetbrains.kotlin.gradle.targets.native.tasks.KotlinNativeSimulatorTest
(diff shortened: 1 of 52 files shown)


sstone (approved):

build.gradle.kts
```diff
@@ -30,16 +32,17 @@ kotlin {
     val bitcoinKmpVersion = "0.20.0" // when upgrading bitcoin-kmp, keep secpJniJvmVersion in sync!
     val secpJniJvmVersion = "0.15.0"
 
-    val serializationVersion = "1.6.2"
-    val coroutineVersion = "1.7.3"
+    val serializationVersion = "1.7.2"
```
dpad85, line 35:
```suggestion
    val serializationVersion = "1.7.1"
```

We have to use v1.7.1 instead of v1.7.2 for now.

That's because kotlinx.serialisation v1.7.2 requires Kotlin 2.0.20 [1], but on Phoenix we can only use Kotlin 2.0.10 (because of the SKIE dependency).

---

[1] the reason is that it requires the `kotlin.uuid.UUID` class which has been introduced in [2.0.20](https://kotlinlang.org/docs/whatsnew2020.html#support-for-uuids-in-the-common-kotlin-standard-library)

sstone (approved):

Issue #321, opened 2022-02-21
link error: cannot find -lcurl


remyers:
Fresh install of Ubuntu 20.04 I get the following link error using `./gradlew :build` after following build steps from BUILD.md .  Any tips why the curl library is not being found?

```
> Task :linkDebugTestLinux
e: /home/remyers/.konan/dependencies/x86_64-unknown-linux-gnu-gcc-8.3.0-glibc-2.19-kernel-4.9-2/x86_64-unknown-linux-gnu/bin/ld.gold invocation reported errors
The /home/remyers/.konan/dependencies/x86_64-unknown-linux-gnu-gcc-8.3.0-glibc-2.19-kernel-4.9-2/x86_64-unknown-linux-gnu/bin/ld.gold command returned non-zero exit code: 1.
output:
/home/remyers/.konan/dependencies/x86_64-unknown-linux-gnu-gcc-8.3.0-glibc-2.19-kernel-4.9-2/x86_64-unknown-linux-gnu/bin/ld.gold: error: cannot find -lcurl
/tmp/konan_temp3788192268220484282/result.o:out:function libcurl_curl_global_init_wrapper23: error: undefined reference to 'curl_global_init'
/tmp/konan_temp3788192268220484282/result.o:out:function libcurl_curl_slist_append_wrapper27: error: undefined reference to 'curl_slist_append'
...cut
/tmp/konan_temp3788192268220484282/result.o:out:knifunptr_libcurl42_curl_easy_getinfo: error: undefined reference to 'curl_easy_getinfo'

> Task :linkDebugTestLinux FAILED
```

It seems like libcurl is installed 

```
$ /sbin/ldconfig  -p | grep curl
        libcurl.so.4 (libc6,x86-64) => /lib/x86_64-linux-gnu/libcurl.so.4
        libcurl-gnutls.so.4 (libc6,x86-64) => /lib/x86_64-linux-gnu/libcurl-gnutls.so.4
```

sstone:
Our tests use `ktor-client` which on linux depends on `ktor-client-curl-linuxx64`.
It can be fixed with `sudo apt install libcurl4-openssl-dev`

remyers:
@sstone I couldn't find anything in the build docs about that dependency. Should I do a PR to add it, or perhaps there is a link to the ktor-client dependencies I should add to BUILD.md ?

sstone:
[...]

PR #702, opened 2024-09-18
Update build docs to install libcurl4-openssh-dev on linux

remyers:
This updates the docs to address issue #321 .

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -16,6 +16,12 @@ brew install libtool
 brew install gmp
 ```
 
+For Linux, our tests use ktor-client which depends on [ktor-client-curl-linuxx64](https://mvnrepository.com/artifact/io.ktor/ktor-client-curl-linuxx64) so you should install:
+
+```sh
+apt install libcurl4-openssh-dev
+```
+
 ## Build
 
 You should start by cloning the repository locally:


BUILD.md
```diff
@@ -16,6 +16,12 @@ brew install libtool
 brew install gmp
 ```
 
+For Linux, our tests use ktor-client which depends on [ktor-client-curl-linuxx64](https://mvnrepository.com/artifact/io.ktor/ktor-client-curl-linuxx64) so you should install:
+
+```sh
+apt install libcurl4-openssh-dev
```
sstone, line 22: ktor has documented what is needed for linux here:https://ktor.io/docs/client-engines.html#curl and they suggest installing the gnutls version of libcurl4, we should add a link to this.

sstone (changes requested):

BUILD.md
remyers: Fixed in [806c569](https://github.com/ACINQ/lightning-kmp/pull/702/commits/806c5690847ef3262f7e5d5696fdccefc0807ae9). How does that look?

sstone (approved):

remyers:
Checks didn't run - can someone restart the action? I can't seem to do it.

sstone:
It's because we explicitly ignore changes to some non-code files (*.md, .gitignore, ...) so for now the only options is to for an admin to merge this PR :(
I'll try and find a way to skip checks for PR that do not change anything that needs to be rebuilt

PR #709, opened 2024-10-08
Split fees between liquidity ads and swap-in

pm47:
A channel open triggered by a swap-in is technically also a liquidity operation (with 1 sat liquidity required), thus causing two consecutive db events: a liquidity purchase followed by a swap-in. Both actions happen with the same on-chain transaction and we must be careful not to count mining fees twice.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
@@ -874,18 +874,14 @@ data class Normal(
             action.fundingTx.signedTx?.let { add(ChannelAction.Blockchain.PublishTx(it, ChannelAction.Blockchain.PublishTx.Type.FundingTx)) }
             add(ChannelAction.Blockchain.SendWatch(watchConfirmed))
             add(ChannelAction.Message.Send(action.localSigs))
-            // If we received or sent funds as part of the splice, we will add a corresponding entry to our incoming/outgoing payments db
-            addAll(origins.map { origin ->
-                ChannelAction.Storage.StoreIncomingPayment.ViaSpliceIn(
-                    amountReceived = origin.amountReceived(),
-                    serviceFee = origin.fees.serviceFee.toMilliSatoshi(),
-                    miningFee = origin.fees.miningFee,
-                    localInputs = action.fundingTx.sharedTx.tx.localInputs.map { it.outPoint }.toSet(),
-                    txId = action.fundingTx.txId,
-                    origin = origin
-                )
-            })
-            // If we added some funds ourselves it's a swap-in
+            // If we purchased liquidity as part of the channel creation, we will add it to our payments db
+            liquidityPurchase?.let { purchase ->
+                // We only count fees corresponding to the liquidity purchase (i.e. their inputs).
+                add(ChannelAction.Storage.StoreOutgoingPayment.ViaInboundLiquidityRequest(txId = action.fundingTx.txId, purchase = purchase))
+                add(ChannelAction.EmitEvent(LiquidityEvents.Purchased(purchase)))
+            }
+            // NB: the following assume that there can't be a splice-in and a splice-out simultaneously, because we attribute
+            // all local mining fees to both.
             if (action.fundingTx.sharedTx.tx.localInputs.isNotEmpty()) add(
                 ChannelAction.Storage.StoreIncomingPayment.ViaSpliceIn(
                     amountReceived = action.fundingTx.sharedTx.tx.localInputs.map { i -> i.txOut.amount }.sum().toMilliSatoshi() - action.fundingTx.sharedTx.tx.localFees,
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/fr/acinq/lightning/channel/states/WaitForFundingSigned.kt
```diff
@@ -119,26 +119,27 @@ data class WaitForFundingSigned(
...
             // If we receive funds as part of the channel creation, we will add it to our payments db
             if (action.commitment.localCommit.spec.toLocal > 0.msat) add(
+                // We only count the mining fees corresponding to the inputs and outputs we paid for in the interactive-tx transaction
+                // (i.e. our wallet inputs and the shared output).
                 ChannelAction.Storage.StoreIncomingPayment.ViaNewChannel(
                     amountReceived = action.commitment.localCommit.spec.toLocal,
-                    serviceFee = channelOrigin?.fees?.serviceFee?.toMilliSatoshi() ?: 0.msat,
-                    miningFee = channelOrigin?.fees?.miningFee ?: action.fundingTx.sharedTx.tx.localFees.truncateToSatoshi(),
+                    serviceFee = 0.msat,
+                    miningFee = action.fundingTx.sharedTx.tx.localFees.truncateToSatoshi(),
```
pm47, line 137: Note how we are not using the `channelOrigin` anymore. There is a certain amount of duplication regarding fee calculation here, I'm not sure how we can simplify. We could revisit later.

src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
```diff
@@ -874,26 +874,22 @@ data class Normal(
...
-                ChannelAction.Storage.StoreIncomingPayment.ViaSpliceIn(
-                    amountReceived = origin.amountReceived(),
-                    serviceFee = origin.fees.serviceFee.toMilliSatoshi(),
-                    miningFee = origin.fees.miningFee,
-                    localInputs = action.fundingTx.sharedTx.tx.localInputs.map { it.outPoint }.toSet(),
-                    txId = action.fundingTx.txId,
-                    origin = origin
-                )
-            })
-            // If we added some funds ourselves it's a swap-in
```
pm47, line 888: There was two separate `ViaSpliceIn` sections, I'm not sure why.

src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
```diff
@@ -874,26 +874,22 @@ data class Normal(
...
+            // all local mining fees to both.
             if (action.fundingTx.sharedTx.tx.localInputs.isNotEmpty()) add(
                 ChannelAction.Storage.StoreIncomingPayment.ViaSpliceIn(
                     amountReceived = action.fundingTx.sharedTx.tx.localInputs.map { i -> i.txOut.amount }.sum().toMilliSatoshi() - action.fundingTx.sharedTx.tx.localFees,
                     serviceFee = 0.msat,
                     miningFee = action.fundingTx.sharedTx.tx.localFees.truncateToSatoshi(),
                     localInputs = action.fundingTx.sharedTx.tx.localInputs.map { it.outPoint }.toSet(),
                     txId = action.fundingTx.txId,
-                    origin = null
+                    origin = origins.filterIsInstance<Origin.OnChainWallet>().firstOrNull()
```
pm47, line 892: The model allows multiple origins but I don't think we can have several of the `OnChainWallet` type. We must have an origin otherwise the event will not be stored in db due to https://github.com/ACINQ/lightning-kmp/blob/9bfdfe61e2634c7b1d225fe6a0d09c6622dd1d48/src/commonMain/kotlin/fr/acinq/lightning/payment/IncomingPaymentHandler.kt#L148.

Note that I have a full rework of the incoming payments db on another branch where this particular part is cleaned up.

src/commonMain/kotlin/fr/acinq/lightning/channel/states/WaitForFundingSigned.kt
```diff
@@ -119,26 +119,27 @@ data class WaitForFundingSigned(
...
             add(ChannelAction.Blockchain.SendWatch(watchConfirmed))
             add(ChannelAction.Message.Send(action.localSigs))
+            // If we purchased liquidity as part of the channel creation, we will add it to our payments db
+            liquidityPurchase?.let { purchase ->
+                if (channelParams.localParams.isChannelOpener) {
+                    // We only count fees corresponding to the liquidity purchase (i.e. their inputs).
+                    add(ChannelAction.Storage.StoreOutgoingPayment.ViaInboundLiquidityRequest(txId = action.fundingTx.txId, purchase = purchase))
+                    add(ChannelAction.EmitEvent(LiquidityEvents.Purchased(purchase)))
+                }
+            }
```
pm47, line 129: We move this event up because it makes more sense to have it happen "before".

PR #712, opened 2024-10-09
Use testnet4 for mempool space unit tests

sstone:
Tests for our mempool space client used testnet3 and were disabled in https://github.com/ACINQ/lightning-kmp/pull/711 because of the ongoing block spam that was making them fail. Here we re-enable these tests but against mempool space's testnet4 servers (testnet4 fixes of lot of the "block spam" issues that are plaguing testnet3).

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/lightning/blockchain/mempool/MempoolSpaceClient.kt
+++ b/src/commonMain/kotlin/fr/acinq/lightning/blockchain/mempool/MempoolSpaceClient.kt
@@ -116,6 +116,7 @@ class MempoolSpaceClient(val mempoolUrl: Url, loggerFactory: LoggerFactory) : IC
     companion object {
         val OfficialMempoolMainnet = Url("https://mempool.space")
         val OfficialMempoolTestnet = Url("https://mempool.space/testnet/")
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/fr/acinq/lightning/blockchain/mempool/MempoolSpaceClient.kt
```diff
@@ -116,6 +116,7 @@ class MempoolSpaceClient(val mempoolUrl: Url, loggerFactory: LoggerFactory) : IC
     companion object {
         val OfficialMempoolMainnet = Url("https://mempool.space")
         val OfficialMempoolTestnet = Url("https://mempool.space/testnet/")
```
pm47, line 118:
```suggestion
        val OfficialMempoolTestnet3 = Url("https://mempool.space/testnet/")
```

sstone: done in https://github.com/ACINQ/lightning-kmp/pull/712/commits/9a490b146b401c88396ed6304305f3787b53590d

pm47 (approved):

PR #718, opened 2024-10-17
Increase blinded path expiry

t-bast:
Blinded paths should expire when we will start rejecting payments for the corresponding invoice: this is thus related to the bolt 12 invoice expiry that we used.

However, when we receive an HTLC, its `cltv_expiry` is set to a future block height, which must be at least `min_final_expiry_delta` in the future. Payers may add some additional margin to the current block height to protect against delays in HTLC relay and protect the privacy of the payment. So we must add a large enough `cltv_expiry_delta` to the invoice expiry to account for those.


thomash-acinq (approved):

pm47 (approved):
Tested against eclair.

PR #719, opened 2024-10-17
Add support for Bolt 12 contacts

t-bast:
Add support for contacts as specified in bLIP 42. Contacts are mutually authenticated using a 32-bytes random secret generated when first adding a node to our contacts. When paying contacts, we include our own payment information to allow them to pay us back and us to their contacts.

The benefit of this design is that offers stay private by default (they don't include any contact information). It's only when we pay someone we trust that we reveal contact information (which they are free to ignore).

The drawback of this design is that if when both nodes independently add each other to their contacts list, they generate a different contact secret: users must manually associate incoming payments to an existing contact to correctly identify incoming payments (by storing multiple secrets for such contacts). This also happens when contacts use multiple wallets, which will all use different contact secrets. I think this is an acceptable trade-off to preserve privacy by default.

More details in the bLIP: https://github.com/lightning/blips/pull/42


t-bast:
The bLIP was merged: https://github.com/lightning/blips/pull/42, so it should be ok to move forward with this feature :partying_face:

PR #724, opened 2024-10-24
Use shared input's `txOut` in `shouldSignFirst`

t-bast:
When deciding who must send `tx_signatures` first for a splice, we compute how much each node contributes to the splice transaction by adding the amount of the `tx_add_input` they have sent.

For the shared input (current channel output), instead of using the `txOut` amount, we previously summed the balance of each node in the channel. The result is the same when there are no pending HTLCs, but when there are pending HTLCs, it underestimates the amount of this input.

This is hard to reproduce (and test) because it requires both nodes to contribute to the splice while having large pending HTLCs that tilt the balance in favor of the wrong side.


remyers (approved):
LGTM!

PR #725, opened 2024-10-28
Make the project multi-modules

pm47:
Instead of using a root project.

Used https://github.com/Kotlin/kotlinx-io as reference.


settings.gradle.kts
```diff
@@ -1,12 +1,25 @@
...
+    }
+}
+
+rootProject.name = "lightning-kmp"
+
+include(":lightning-kmp-ios-crypto")
+include(":lightning-kmp-core")
+
+project(":lightning-kmp-ios-crypto").projectDir = file("./modules/ios-crypto")
+project(":lightning-kmp-core").projectDir = file("./modules/core")
```
dpad85, line 25: The projects' name `lightning-kmp-xxx` is quite large, it makes running specific gradle commands a bit annoying (for example, `./gradlew :lightning-kmp-core:publishToMavenLocal`). Why not use `core` and `ios-crypto` as names ?

dpad85 (commented):
The project now publishes a Maven artifact under artifact id `lightning-kmp-core` instead of `lightning-kmp`, which affects dependant projects.

We could copy eclair, with publishes a top-level `eclair` module and also several artifacts like `eclair-core` or `eclair-node`. So we'd have:
- `lightning-kmp` root project
- `lightning-kmp-core` module
- `lightning-kmp-db` module

If you only need the core dependency, you would run `./gradlew :lightning-kmp-core:publishToMavenLocal`. If you need core+db, you run `./gradlew publishToMavenLocal`. Not quite sure how to do make this happen in gradle though.

Another option could be to modify the MavenPublication task and rename the `lightning-kmp-core` Maven artifact to `lightning-kmp`, so that we keep the same id as before.

settings.gradle.kts
pm47: Done in 684272ed. I didn't know that I could override the artifact id.

sstone:
`modules` branch should be renamed to `snapshot/modules`. This will trigger our publishing workflow on github Actions and will let us publish snapshot versions to  https://oss.sonatype.org/content/repositories/snapshots/ (our publishing scripts will probably need to be updated too) so we can also test that downstream projects (phoenix, phoenixd) can still be built properly.

settings.gradle.kts
pm47: This has side effects on publication, had to revert changes...

pm47:
> We could copy eclair, with publishes a top-level eclair module and also several artifacts like eclair-core or eclair-node.

Actually, linking the root `eclair` project will not pull all child dependencies. We could try to do that on `lightning-kmp`, but I'm not sure it's worth it.

PR #727, opened 2024-11-13
Manage dependencies with a version catalog

pm47:
[Version catalog](https://docs.gradle.org/current/userguide/version_catalogs.html) is the best practice for managing dependencies. It makes it easier to track and update versions, and allows to centralize dependencies in a multi-module context. It is an alternative to the `buildSrc` workaround used in some of our projects.

This PR does not update libraries versions.

Each commit offers an alternative approach:
a) define the version catalog in `settings.gradle`
b) define the version catalog in `libs.versions.toml`
c) only manage versions (not full dependencies) in the version catalog

Option c) has my preference because we get all the benefit of centralizing versions, but we don't really need to define all dependencies in the version catalog if they are not used by several modules.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,0 +1,31 @@
+[versions]
+kotlin = "1.9.23"
+kotlinx-coroutines = "1.7.3"
+kotlinx-serialization = "1.6.2"
+ktor = "2.3.7"
+
+[plugins]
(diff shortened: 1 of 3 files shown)


dpad85 (commented):
I also prefer option c).

- option b) is less readable, because 1) we can't easily parse which dependency is used by which target by just reading the `build.gradle.kts` file ; 2) the toml file is very busy.
- option a) is the same, but somewhat worse since the data are put in the settings file.

PR #736, opened 2024-12-13
Correctly set `next_commitment_number` during splice reconnect

t-bast:
As pointed out in https://github.com/lightning/bolts/pull/1214, when reconnecting a partially signed `interactive-tx` session, we should set `next_commitment_number` to the current commitment number if we haven't received our peer's `commit_sig`, which tells them they need to retransmit it.

More importantly, if our peer behaves correctly and sends us the current commitment number, we must not think that they're late and halt, waiting for them to send `error`. This commit fixes that by allowing our peers to use the current commitment number when they set `next_funding_txid`.

Note that we keep retransmitting our `commit_sig` regardless of the value our peer set in `next_commitment_number`, because we need to wait for them to have an opportunity to upgrade. In a future commit we will stop sending spurious `commit_sig`.

This PR must only be merged once https://github.com/ACINQ/eclair/pull/2965 has been deployed on the ACINQ node.

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Channel.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Channel.kt
@@ -16,10 +16,12 @@ import fr.acinq.lightning.channel.Helpers.Closing.claimRevokedRemoteCommitTxOutp
 import fr.acinq.lightning.channel.Helpers.Closing.getRemotePerCommitmentSecret
 import fr.acinq.lightning.crypto.KeyManager
 import fr.acinq.lightning.db.ChannelClosingType
-import fr.acinq.lightning.logging.*
+import fr.acinq.lightning.logging.LoggingContext
+import fr.acinq.lightning.logging.MDCLogger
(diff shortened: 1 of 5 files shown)


modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Channel.kt
```diff
@@ -306,7 +308,7 @@ sealed class PersistedChannelState : ChannelState() {
             val myFirstPerCommitmentPoint = keyManager.channelKeys(state.channelParams.localParams.fundingKeyPath).commitmentPoint(0)
             ChannelReestablish(
                 channelId = channelId,
-                nextLocalCommitmentNumber = 1,
+                nextLocalCommitmentNumber = state.signingSession.reconnectNextLocalCommitmentNumber,
```
pm47, line 311: This will always be 0 right?

pm47 (approved):
LGTM, integration tests with eclair would be nice.

NB: https://github.com/ACINQ/eclair/pull/2965 isn't yet on mainnet

pm47 (approved):
Now safe to merge

PR #740, opened 2025-01-02
Store per-commitment `short_channel_id`

t-bast:
Each commitment has a different `short_channel_id`: it should thus be stored inside the `Commitment` object. We keep a `shortChannelId` field in the channel state because it contains:

- our scid alias if the channel is 0-conf
- otherwise the real scid of the initial funding transaction

This should be replaced by the local/remote aliases once we support `scid_alias` instead of using our custom `PeerId`. Note that our current `PeerId` trick is compatible with `scid_alias` since we send it using the same TLV in `channel_ready`.

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
@@ -906,6 +906,7 @@ data class Commitments(
                     }
                     is LocalFundingStatus.ConfirmedFundingTx -> when (c.localFundingStatus.shortChannelId) {
                         ShortChannelId(0) -> {
+                            // This branch is only useful to update old channel states that didn't store the short_channel_id.
                             logger.debug { "setting short_channel_id for fundingTxId=${fundingTx.txid}" }
                             c.copy(localFundingStatus = c.localFundingStatus.copy(shortChannelId = shortChannelId))
                         }
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/WaitForInit.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/WaitForInit.kt
@@ -78,7 +78,7 @@ data object WaitForInit : ChannelState() {
                         when (commitment.localFundingStatus) {
                             is LocalFundingStatus.UnconfirmedFundingTx -> WatchConfirmed(cmd.state.channelId, commitment.fundingTxId, commitment.commitInput.txOut.publicKeyScript, fundingMinDepth, WatchConfirmed.ChannelFundingDepthOk)
                             is LocalFundingStatus.ConfirmedFundingTx -> when (commitment.localFundingStatus.shortChannelId) {
-                                // If the short_channel_id isn't correctly set, we fetch the funding transaction to update it.
+                                // If the short_channel_id isn't correctly set, we put a watch on the funding transaction to compute it.
                                 ShortChannelId(0) -> WatchConfirmed(cmd.state.channelId, commitment.fundingTxId, commitment.commitInput.txOut.publicKeyScript, fundingMinDepth, WatchConfirmed.ChannelFundingDepthOk)
                                 else -> WatchSpent(cmd.state.channelId, commitment.fundingTxId, commitment.commitInput.outPoint.index.toInt(), commitment.commitInput.txOut.publicKeyScript, WatchSpent.ChannelSpent(commitment.fundingAmount))
                             }


vincenzopalazzo (approved):
LGTM

modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
```diff
@@ -261,7 +261,7 @@ data class Normal(
                         }
                     }
                     is ChannelUpdate -> {
-                        if (cmd.message.shortChannelId == shortChannelId && cmd.message.isRemote(staticParams.nodeParams.nodeId, staticParams.remoteNodeId)) {
+                        if (cmd.message.isRemote(staticParams.nodeParams.nodeId, staticParams.remoteNodeId)) {
```
pm47, line 264: I know we are already checking in `Peer`, but I think we should also call `matchesShortChannelId` here.

modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/WaitForInit.kt
```diff
@@ -68,7 +68,11 @@ data object WaitForInit : ChannelState() {
                     is ChannelStateWithCommitments -> cmd.state.commitments.active.mapNotNull { commitment ->
                         when (val fundingStatus = commitment.localFundingStatus) {
                             is LocalFundingStatus.UnconfirmedFundingTx -> fundingStatus.signedTx
-                            is LocalFundingStatus.ConfirmedFundingTx -> null
+                            is LocalFundingStatus.ConfirmedFundingTx -> when (fundingStatus.shortChannelId) {
+                                // If the short_channel_id isn't correctly set, we fetch the funding transaction to update it.
+                                ShortChannelId(0) -> fundingStatus.signedTx
+                                else -> null
+                            }
```
pm47, line 75: I see the intention, but I don't get how it works. AFAICT we only call `ChannelAction.Blockchain.PublishTx` on those txs. Shouldn't we put a `WatchConfirm` or something?

modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
pm47: I consider the channel state machine to be independent from the peer. Imo it shouldn't just take whatever is sent to it for granted. Not a huge deal

modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -890,15 +895,22 @@ data class Commitments(
...
         updateFundingStatus(fundingTx.txid) { c: Commitment, _: Long ->
             if (c.fundingTxId == fundingTx.txid) {
+                val shortChannelId = ShortChannelId(blockHeight, txIndex, c.commitInput.outPoint.index.toInt())
                 when (c.localFundingStatus) {
                     is LocalFundingStatus.UnconfirmedFundingTx -> {
                         logger.debug { "setting localFundingStatus confirmed for fundingTxId=${fundingTx.txid}" }
-                        c.copy(localFundingStatus = LocalFundingStatus.ConfirmedFundingTx(fundingTx, c.localFundingStatus.sharedTx.tx.fees, c.localFundingStatus.sharedTx.localSigs))
+                        c.copy(localFundingStatus = LocalFundingStatus.ConfirmedFundingTx(fundingTx, c.localFundingStatus.sharedTx.tx.fees, c.localFundingStatus.sharedTx.localSigs, shortChannelId))
+                    }
+                    is LocalFundingStatus.ConfirmedFundingTx -> when (c.localFundingStatus.shortChannelId) {
```
pm47, line 907:
```suggestion
                    // This branch is only useful to update old channel states that didn't store the short channel id
                    is LocalFundingStatus.ConfirmedFundingTx -> when (c.localFundingStatus.shortChannelId) {
```

modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/WaitForInit.kt
```diff
@@ -73,13 +74,13 @@ data object WaitForInit : ChannelState() {
...
-                                val fundingMinDepth = staticParams.nodeParams.minDepth(commitment.fundingAmount)
-                                WatchConfirmed(cmd.state.channelId, commitment.fundingTxId, commitment.commitInput.txOut.publicKeyScript, fundingMinDepth, WatchConfirmed.ChannelFundingDepthOk)
-                            }
-                            is LocalFundingStatus.ConfirmedFundingTx -> {
-                                WatchSpent(cmd.state.channelId, commitment.fundingTxId, commitment.commitInput.outPoint.index.toInt(), commitment.commitInput.txOut.publicKeyScript, WatchSpent.ChannelSpent(commitment.fundingAmount))
+                            is LocalFundingStatus.UnconfirmedFundingTx -> WatchConfirmed(cmd.state.channelId, commitment.fundingTxId, commitment.commitInput.txOut.publicKeyScript, fundingMinDepth, WatchConfirmed.ChannelFundingDepthOk)
+                            is LocalFundingStatus.ConfirmedFundingTx -> when (commitment.localFundingStatus.shortChannelId) {
+                                // If the short_channel_id isn't correctly set, we fetch the funding transaction to update it.
```
pm47, line 81:
```suggestion
                                // If the short_channel_id isn't correctly set, we put a watch on the funding transaction to compute it.
```

modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -190,7 +191,7 @@ sealed class LocalFundingStatus {
         override val fee: Satoshi = sharedTx.tx.fees
     }
 
-    data class ConfirmedFundingTx(override val signedTx: Transaction, override val fee: Satoshi, val localSigs: TxSignatures) : LocalFundingStatus() {
+    data class ConfirmedFundingTx(override val signedTx: Transaction, override val fee: Satoshi, val localSigs: TxSignatures, val shortChannelId: ShortChannelId) : LocalFundingStatus() {
```
pm47, line 194: Since we are not currently using the `scid`, I would probably have left it nullable and not do the update trick until we really need it. But your way makes the model cleaner.

modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/states/Normal.kt
pm47: True!

pm47 (approved):

modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
pm47: I would have left it nullable _until_ we use it (since it may never happen?)

PR #744, opened 2025-01-07
Cancel splice early when missing funds

t-bast:
When we are splicing out or doing a CPFP on an existing splice, we use our channel balance to pay fees and the LSP doesn't contribute at all. We must have enough funds to pay the mining fees, otherwise we will go through all of the `interactive-tx` protocol but fail at the end when our peer will notice that we're not paying the feerate we said we would.

We now fail early in that case, which also lets us provide a better error explaining why the splice attempt was cancelled.

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
@@ -249,7 +249,7 @@ sealed class FundingContributionFailure {
 data class FundingContributions(val inputs: List<InteractiveTxInput.Outgoing>, val outputs: List<InteractiveTxOutput.Outgoing>) {
     companion object {
         /** Compute our local splice contribution using all the funds available in our wallet. */
-        fun computeSpliceContribution(isInitiator: Boolean, commitment: Commitment, walletInputs: List<WalletState.Utxo>, localOutputs: List<TxOut>, liquidityPurchase: Boolean, targetFeerate: FeeratePerKw): Satoshi {
+        fun computeSpliceContribution(isInitiator: Boolean, commitment: Commitment, walletInputs: List<WalletState.Utxo>, localOutputs: List<TxOut>, isLiquidityPurchase: Boolean, targetFeerate: FeeratePerKw): Satoshi {
             val weight = computeWeightPaid(isInitiator, commitment, walletInputs, localOutputs)
             val fees = Transactions.weight2fee(targetFeerate, weight)
             return when {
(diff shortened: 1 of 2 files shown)


modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -249,15 +249,15 @@ sealed class FundingContributionFailure {
 data class FundingContributions(val inputs: List<InteractiveTxInput.Outgoing>, val outputs: List<InteractiveTxOutput.Outgoing>) {
     companion object {
         /** Compute our local splice contribution using all the funds available in our wallet. */
-        fun computeSpliceContribution(isInitiator: Boolean, commitment: Commitment, walletInputs: List<WalletState.Utxo>, localOutputs: List<TxOut>, targetFeerate: FeeratePerKw): Satoshi {
+        fun computeSpliceContribution(isInitiator: Boolean, commitment: Commitment, walletInputs: List<WalletState.Utxo>, localOutputs: List<TxOut>, liquidityPurchase: Boolean, targetFeerate: FeeratePerKw): Satoshi {
```
pm47, line 252: Nit

```suggestion
        fun computeSpliceContribution(isInitiator: Boolean, commitment: Commitment, walletInputs: List<WalletState.Utxo>, localOutputs: List<TxOut>, isLiquidityPurchase: Boolean, targetFeerate: FeeratePerKw): Satoshi {
```

modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/InteractiveTx.kt
```diff
@@ -249,15 +249,15 @@ sealed class FundingContributionFailure {
...
             val weight = computeWeightPaid(isInitiator, commitment, walletInputs, localOutputs)
             val fees = Transactions.weight2fee(targetFeerate, weight)
             return when {
                 // When buying inbound liquidity, we may not have enough funds in our current balance to pay on-chain fees.
                 // The maximum amount we can use for on-chain fees is our current balance, which is fine because:
                 //  - this will simply result in a splice transaction with a lower feerate than expected
                 //  - liquidity fees will be paid later from future HTLCs relayed to us
-                walletInputs.isEmpty() && localOutputs.isEmpty() -> -(fees.min(commitment.localCommit.spec.toLocal.truncateToSatoshi()))
+                walletInputs.isEmpty() && localOutputs.isEmpty() && liquidityPurchase -> -(fees.min(commitment.localCommit.spec.toLocal.truncateToSatoshi()))
```
pm47, line 260: If we're doing a splice out, `localOutputs.isEmpty()` would already be false, so this only changes the behavior for a splice cpfp ?

pm47 (approved):

PR #746, opened 2025-01-14
Add a `succeededAt` timestamp to payments

pm47:
Another follow-up to #722 and #738.

This allows separating between a _completed_ payment and a _successful_ payment.

Incoming payments are always successful when they are completed.

Outgoing payments may be completed, but unsuccessful. For example a Lightning outgoing payment is completed when all attempts have been exhausted but no route was found.


dpad85 (dismissed):

pm47:
Rebased on master.

dpad85 (approved):

PR #747, opened 2025-01-21
Include liquidity purchase in `NewChannelIncomingPayment`

pm47:
Since we started using the standard liquidity ads protocol to handle  swap-ins, the case of a new channel creation led to two db events: a `NewChannelIncomingPayment` and a `InboundLiquidityOutgoingPayment`. But both from a UX and from a technical point of view, these two operations actually happen atomically. Separating them creates integration hurdles, as they need to be merged again.

The first commit adds a `LiquidityAds.Purchase` to `NewChannelIncomingPayment`, and removes the `InboundLiquidityOutgoingPayment` db event. It also changes the meaning of `NewChannelIncomingPayment.miningFee` to be the overall mining fee (it was actually misnamed, should have been `localMiningFee`).

The second commit is just fixing consistency issues in naming.

The third commit removes `localMiningFee` from `InboundLiquidityOutgoingPayment`, for consistency with the above. Note that, while more pure from a technical point of view, having `localMiningFee` a main class attribute adds mental overhead, and has no use for integrators, who typically reason in total mining fee. In Phoenix this led to niceties such as: https://github.com/ACINQ/phoenix/blob/a07c153ad4e22b042dc38fed11fab89fd1f8a4de/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/db/payments/InboundLiquidityQueries.kt#L100-L107


pm47:
Superseded by #749.

PR #752, opened 2025-02-05
Fix feerate adjustment for on-the-fly liquidity

t-bast:
When purchasing on-the-fly liquidity without any channel balance, we inflate the feerate because we cannot contribute to the mining fees, so that the final feerate is close to the target feerate.

However, the ratios we were using weren't properly computed which resulted in transactions below 1 sat/byte, which thus couldn't be relayed. We now compute the real ratios depending on the number of inputs/outputs the LSP will contribute, and adjust the feerate based on those ratios.

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
@@ -69,15 +69,19 @@ data class AddLiquidityForIncomingPayment(val paymentAmount: MilliSatoshi, val r
 
     companion object {
         /**
-         * When purchasing liquidity for on-the-fly HTLCs, we're supposed for pay the mining fees for:
+         * When purchasing liquidity for on-the-fly HTLCs, we're supposed to pay the mining fees for:
          *  - the common transaction fields (nVersion, nLockTime, etc)
          *  - the channel output
          *  - if it's a splice, the previous channel input
-         * However, if we don't have any balance in that channel, we won't be able to contribute those mining fees.
+         *
+         * However, if we don't have any balance in that channel or no channel at all, we won't be able to contribute
+         * those mining fees.
+         *
          * Instead, we adjust the target feerate to a greater value and the LSP will pay mining fees for their inputs
          * and outputs at this adjusted feerate: if we use the correct ratio, the resulting mining fees will match
          * the initial target feerate. This won't be a loss for the LSP, because we will refund some of those mining
          * fees by subtracting them from the received HTLCs at the adjusted feerate (see [LiquidityAds.Fees.miningFee]).
+         *
          * The main difficulty is that we don't know how many inputs and outputs the LSP will contribute, which impacts
          * the ratio: at lower feerates we use the most conservative ratios that assume a single input and an optional
          * change output, which at higher feerates we use a more loose ratio to avoid penalizing the LSP too much if


modules/core/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -68,30 +69,30 @@ data class AddLiquidityForIncomingPayment(val paymentAmount: MilliSatoshi, val r
...
-         * When we open a new channel without contributing any input, we won't be able to pay on-chain fees for our
-         * weight of the funding transaction. If we don't do anything, the resulting transaction will thus have a
-         * lower feerate than requested and may not confirm.
-         *
-         * To avoid that, we ask our peer to target a higher feerate than the one we actually want. They will pay
-         * more mining fees to satisfy that feerate, while we won't pay any mining fees. We should be paying for the
-         * shared output, which doesn't add too much weight, so we add 25%. This is hacky but should result in an
-         * effective feerate that is somewhat close to the initial feerate we wanted. Note that we will pay liquidity
-         * fees based on this inflated feerate, which will refund our peer for this hack.
+         * When purchasing liquidity for on-the-fly HTLCs, we're supposed for pay the mining fees for:
```
pm47, line 72:
```suggestion
         * When purchasing liquidity for on-the-fly HTLCs, we're supposed to pay the mining fees for:
```

modules/core/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -68,30 +69,30 @@ data class AddLiquidityForIncomingPayment(val paymentAmount: MilliSatoshi, val r
...
+         *  - the common transaction fields (nVersion, nLockTime, etc)
+         *  - the channel output
+         *  - if it's a splice, the previous channel input
+         * However, if we don't have any balance in that channel, we won't be able to contribute those mining fees.
```
pm47, line 76:
```suggestion
         * However, if we don't have any balance in that channel or no channel at all, we won't be able to contribute those mining fees.
```

modules/core/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -68,30 +69,30 @@ data class AddLiquidityForIncomingPayment(val paymentAmount: MilliSatoshi, val r
...
+         * Instead, we adjust the target feerate to a greater value and the LSP will pay mining fees for their inputs
+         * and outputs at this adjusted feerate: if we use the correct ratio, the resulting mining fees will match
+         * the initial target feerate. This won't be a loss for the LSP, because we will refund some of those mining
+         * fees by subtracting them from the received HTLCs at the adjusted feerate (see [LiquidityAds.Fees.miningFee]).
+         * The main difficulty is that we don't know how many inputs and outputs the LSP will contribute, which impacts
+         * the ratio: at lower feerates we use the most conservative ratios that assume a single input and an optional
```
pm47, line 82: But we know the LSPs estimate for those in their liquidity ads announcement, right? Shouldn't we use those?

pm47: The `funding_weight` in the `funding_rate` precisely corresponds to the number of inputs and outputs that the LSP estimates it will use. I think it really makes sense to use that?

pm47: Right! I get it now.

PR #753, opened 2025-02-10
Forbid concurrent payments with the same `payment_hash`

pm47:
Concurrent payments with the same `payment_hash` must be rejected, because if any of the payments succeeds, any node in the route could claim the full amount as soon as they learn the preimage.

We only checked that a payment didn't already succeed, but that is not enough: pending payments need to be considered too.

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentHandler.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentHandler.kt
@@ -116,9 +116,13 @@ class OutgoingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
             logger.error { "contract violation: caller is recycling uuid's" }
             return Failure(request, FinalFailure.InvalidPaymentId.toPaymentFailure())
         }
-        if (db.listLightningOutgoingPayments(request.paymentHash).find { it.status is LightningOutgoingPayment.Status.Succeeded } != null) {
+        if (db.listLightningOutgoingPayments(request.paymentHash).any { it.status is LightningOutgoingPayment.Status.Succeeded }) {
             logger.error { "invoice has already been paid" }
             return Failure(request, FinalFailure.AlreadyPaid.toPaymentFailure())
+        }
+        if (db.listLightningOutgoingPayments(request.paymentHash).any { it.status is LightningOutgoingPayment.Status.Pending }) {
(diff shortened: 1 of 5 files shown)


modules/core/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentHandler.kt
```diff
@@ -116,10 +116,14 @@ class OutgoingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
             logger.error { "contract violation: caller is recycling uuid's" }
             return Failure(request, FinalFailure.InvalidPaymentId.toPaymentFailure())
         }
-        if (db.listLightningOutgoingPayments(request.paymentHash).find { it.status is LightningOutgoingPayment.Status.Succeeded } != null) {
+        if (db.listLightningOutgoingPayments(request.paymentHash).any { it.status is LightningOutgoingPayment.Status.Succeeded }) {
```
dpad85, line 119: Should we not also check the status of parts?

That's what we did in Phoenix when applying the same restriction (https://github.com/ACINQ/phoenix/blob/a1dd05d23538f0916cbd1751c826342a100fb4c4/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/SendManager.kt#L214-L221). Maybe it's not actually useful though.

pm47: Not only is it redundant, it is also confusing as it assumes impossible combinations are possible.

modules/core/src/commonMain/kotlin/fr/acinq/lightning/payment/OutgoingPaymentHandler.kt
```diff
@@ -116,10 +116,14 @@ class OutgoingPaymentHandler(val nodeParams: NodeParams, val walletParams: Walle
...
             logger.error { "invoice has already been paid" }
             return Failure(request, FinalFailure.AlreadyPaid.toPaymentFailure())
         }
+        if (db.listLightningOutgoingPayments(request.paymentHash).any { it.status is LightningOutgoingPayment.Status.Pending }) {
```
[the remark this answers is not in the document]
pm47: Totally! Done in d4673a41c1b0ad3cef78eee2851033f86ccde57e.

dpad85 (approved):

PR #754, opened 2025-02-10
(Minor) Emit `PaymentSent` event for `LightningOutgoingPayment`s

pm47:
We were only emitting those events for `OnChainOutgoingPayment`s.

This is useful for phoenixd websocket/webhook feature.

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/NodeEvents.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/NodeEvents.kt
@@ -1,8 +1,8 @@
 package fr.acinq.lightning
 
 import fr.acinq.bitcoin.ByteVector32
-import fr.acinq.bitcoin.PublicKey
 import fr.acinq.bitcoin.OutPoint
+import fr.acinq.bitcoin.PublicKey
 import fr.acinq.bitcoin.Satoshi
 import fr.acinq.lightning.blockchain.electrum.WalletState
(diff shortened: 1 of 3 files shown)


pm47:
> We're already emitting a PaymentSent (or PaymentNotSent or OfferNotPaid) event on the _eventsFlow when handling a ProcessCmdRes in Peer.kt, shouldn't we clean-up this duplication? We have more detailed reporting for off-chain payments as part of the sealed class SendPaymentResult : PeerEvent().

It's actually a different flow and a different class. The `Peer.eventsFlow` existed before `NodeEvents` and we left it for backward compat with phoenix. @dpad85 are you stil using it?

> Detecting that an off-chain payment wasn't sent or was failed is also important and something we want to have on the websocket, isn't it?

Absolutely, and the primary reason why I was looking into that is that we actually got a user request for a notification for failed (timed out) HTLCs. It was not immediately trivial to do, so I only did this minimal PR for the success case and planned to get back to it later. I also noticed there is some factorization to do in `OutgoingPaymentHandler`, and potential simplification of `Success`/`Failure` classes.

dpad85:
> It's actually a different flow and a different class. The `Peer.eventsFlow` existed before `NodeEvents` and we left it for backward compat with phoenix. @dpad85 are you still using it?

`Peer.eventsFlow` is only used in the iOS application, to provide a better UX when paying an offer. So we could probably remove it, but I would prefer if it was done later, when things are stable in Phoenix.

PR #756, opened 2025-02-11
Fix HTLC-penalty weight calculation

t-bast:
We were incorrectly computing the weight of HTLC-penalty transactions (using the weight of a main-penalty transaction, which has a smaller script). We fix that and clean-up some warnings in `Transactions.kt`.


sstone (approved):
Nice catch !

sstone (approved):

PR #757, opened 2025-02-11
Upgrade to Kotlin 2.1.10, Gradle 8.10 and target JVM 21

pm47:
And upgrade  other dependencies.

Based on the work made in https://github.com/ACINQ/lightning-kmp/pull/701, that was reverted.

We are not using the new `kotlin.Uuid`, as it is still experimental.

### DIFF
--- a/modules/core/build.gradle.kts
+++ b/modules/core/build.gradle.kts
@@ -1,3 +1,5 @@
+import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.plugin.mpp.DefaultCInteropSettings
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
 import org.jetbrains.kotlin.gradle.targets.native.tasks.KotlinNativeSimulatorTest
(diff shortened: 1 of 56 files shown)


modules/core/build.gradle.kts
```diff
@@ -128,42 +130,37 @@ kotlin {
...
+                }
             }
         }
     }
 }
 
-val deleteDokkaOutputDir by tasks.register<Delete>("deleteDokkaOutputDirectory") {
-    delete(dokkaOutputDir)
-}
-
```
pm47, line 162: @sstone This seemed superfluous, as the `clean` task does the trick.

modules/core/build.gradle.kts
```diff
@@ -14,8 +16,8 @@ val currentOs = org.gradle.internal.os.OperatingSystem.current()
 
 kotlin {
     jvm {
-        compilations.all {
-            kotlinOptions.jvmTarget = "1.8"
+        compilerOptions {
+            jvmTarget.set(JvmTarget.JVM_21)
```
sstone, line 20: Does this affect which Android platform phoenix is compatible with ?

pm47: Good point! It seems we can maybe use [desugaring](https://developer.android.com/studio/write/java11-default-support-table). Otherwise we'll be stuck on JDK 8.

cc @dpad85

dpad85: We should keep targeting 1.8, on Phoenix the minimum Android version is API 26 (Android 8).

sstone (approved):
LGTM! the publication CI scripts are not tested but I've checked  that publishing to the local  maven repo works and sources and javadocs are included so If there's an issue it will be easy to fix.

pm47:
Waiting from ACK from @dpad85 before merging.

dpad85 (approved):
Lightning-kmp builds fine on an Intel macOS 14.6 and a M3 macOS 15.3, using Xcode 16.2 (which is the latest version). We've also tested on devices with the Phoenix kotlin-2.x branch https://github.com/ACINQ/phoenix/pull/654

PR #764, opened 2025-02-14
Update ktor to 3.1.0

pm47:
It's basically a drop-in replacement.


dpad85 (approved):
Tested on iOS and Android.

However a property that we've been using has been removed, see https://github.com/ACINQ/phoenix/blob/91bd78e1d06dba1ffe32b4ec45e4af3c3d1a039e/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/managers/CurrencyManager.kt#L650

@robbiehanson that issue may have been fixed, or not be relevant anymore and needs testing

pm47:
👀  https://github.com/ktorio/ktor/blob/7808fca104aed078b73f69c9d46ca8afab04a298/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClientConfig.kt#L19-L22

PR #768, opened 2025-03-03
Use official splice messages

t-bast:
We replace our experimental version of `splice_init`, `splice_ack` and `splice_locked` by their official version. We also change the TLV fields added to `commit_sig`, `tx_add_input`, and `tx_signatures` to match the spec version and introduce the `start_batch` message.

We also implement the updated retransmission logic for `channel_ready` and `splice_locked` based on the `my_current_funding_locked` and `next_funding` TLVs added to `channel_reestablish`.

We only allow connecting to peers who support the official splicing feature.

This update should be released *after* deploying https://github.com/ACINQ/eclair/pull/2887. Once enough users have upgraded, we'll be able to remove support from the legacy splicing protocol from `eclair`, which will provide a nice code clean-up.


pm47 (approved):
Integration tests passed.

PR #770, opened 2025-03-21
Remove amount-based confirmation scaling

t-bast:
We previously scaled the number of confirmations based on the channel capacity: however, this doesn't really work with splicing, as shown by the following attack scenario:

- a malicious node opens a very large channel
- we thus wait for more confirmations to protect against reorgs
- they then send lightning payments until all of the balance is on our side of the channel
- we splice-out to shrink the channel to a minimal size
- if we thus us a smaller number of confirmations (because the channel is now small), the malicious node can perform this small reorg and publish one of their revoked commitment

There are more involved variations of this attack, which show that the real amount at stake is potentially much more than a single channel's current capacity. Instead of trying to find a complex solution that would likely have subtle edge cases, it's a lot simpler to always use a static number of confirmations that is large enough to protect against reorgs entirely.


remyers:
Concept ACK. While not harmful, the current scaling logic may give a false sense of reorg safety.

Generally, after a malicious reorg our counter party would use a revoked commitment that we are unable to create a penalty transaction for because we pruned our corresponding commitment, right? 

If we had not pruned our commitment we could respond to their revoked commitment and they would have to also reorg out our penalty tx, unless their reorg also included sufficient additional blocks to spend the revoked commitment.

If we delay when we prune the commitments from our database, then that would also increase the uncertainty of success for an attacker considering a reorg attack I think.

pm47 (approved):
LGTM

PR #771, opened 2025-03-24
Bolt11 invoices: do not re-encode amounts

sstone:
Bolt11 amounts may not be minimally encoded (the specs says they SHOULD be but it's not a hard requirement). 
We should preserve the original amount encoding otherwise we won't be able to read it: when we read a BOLT11 invoice, we:
- decode all the fields that we understand
- convert the amount  to millisatoshis
- and then serialize these fields again, using minimal encoding for the amount, to validate the checksum
=> if we decode a non-minimally encoded amount and reencode minimally, checksum verification will fail.

This means that lightning-kmp is currently stricter than the specs and will refuse non-minimally encoded BOLT invoices. AFAIK, there are no wallets that produce such invoices, but some very old invoices are encoded this way.


modules/core/src/commonMain/kotlin/fr/acinq/lightning/payment/Bolt11Invoice.kt
```diff
@@ -11,16 +11,21 @@ import fr.acinq.lightning.*
 import fr.acinq.lightning.Lightning.randomBytes32
 import fr.acinq.lightning.utils.*
 import fr.acinq.lightning.wire.LightningCodecs
+import fr.acinq.secp256k1.Hex
```
pm47, line 14:
```suggestion
```

modules/core/src/commonMain/kotlin/fr/acinq/lightning/payment/Bolt11Invoice.kt
```diff
@@ -11,16 +11,21 @@ import fr.acinq.lightning.*
...
 import kotlin.experimental.and
 
 data class Bolt11Invoice(
```
pm47, line 17: A comment is absolutely required for such non-conventional implementation.

```suggestion
/**
 * @param encodedAmount We counterintuitively store the string-serialized amount instead of the numeric value. The goal is to preserve the
 *                      validity of the signature for invoices generated by 3rd party software when the amount is not canonically encoded.
 */
data class Bolt11Invoice(
```

pm47 (dismissed):
I can't say it looks good, but I don't have a better/simpler alternative. We can quickly run into inconsistency issues if we store two fields (numeric and serialized).

PR #773, opened 2025-03-28
Send normal failures for Bolt12 payments

t-bast:
Since we are always using 1-hop blinded paths from our LSP to ourselves, there are no public intermediate nodes that could be probed inside that blinded path. The requirement to always return malformed failure is thus unnecessary (and harmful for the payer who doesn't know why the payment failed). We now return normal failures in that case, which should be forwarded by our LSP to the upstream nodes.

This also fixes an issue where non-trampoline blinded payments were not correctly failed because the extraction of the onion shared secret did not uses the `path_key` to derive the blinded onion decryption key.

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
@@ -702,7 +702,7 @@ data class Commitments(
             // we have already sent a fail/fulfill for this htlc
             CommitmentChanges.alreadyProposed(changes.localChanges.proposed, htlc.id) -> Either.Left(UnknownHtlcId(channelId, cmd.id))
             else -> {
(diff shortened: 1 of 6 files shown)


thomash-acinq:
There are no public intermediate nodes but the sender doesn't know it. By returning normal failures we would tell the sender that the recipient is a phoenix.
But maybe it's just the spec that's too restrictive and should be changed: intermediate nodes should obfuscate their errors to not leak information but the final node can return unobfuscated errors. We would need to check if some error from the recipient could leak information on the path.

thomash-acinq:
It does reduce the anonymity set from "any peer of the LSP (public nodes and wallets)" to only "wallets using this non-standard variation (so just phoenix)".

thomash-acinq (approved):

pm47 (commented):
I'm too unfamiliar with blinded payments to validate the implementation, but ACK on the approach.

@dpad85 is preparing a custom phoenix build for validation before an official release.

PR #780, opened 2025-04-15
Remove the `testing` directory

pm47:
It was useful for the `Node` demo app, which has been removed a long time ago.


pm47:
It's actually used in CI to instantiate bitcoind and electrum.

PR #787, opened 2025-05-12
Send splice `commit_sig`s as a batch

t-bast:
There has been some discussion on the splicing spec around batching the set of `commit_sig` messages that are sent while a splice is ongoing (see https://github.com/lightning/bolts/pull/1160/commits/b5b5572ede3563e856e79d650e29716b639bd792).

In this PR, we ensure that our `commit_sig` messages are sent as a batch (no other messages can be interleaved between `commit_sig` messages). We also receive `commit_sig` messages as a batch, which means we must wait for https://github.com/ACINQ/eclair/pull/3083 to be deployed before merging this PR.

We don't yet implement the `start_batch` message that was introduced in the BOLTs: this will be done once `eclair` supports it and the spec PR is merged (see #768).

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
@@ -782,20 +782,12 @@ data class Commitments(
         // while our peer is sending us a batch of commit_sig. When that happens, we simply need to discard the commit_sig that belong
         // to commitments we deactivated.
         val sigs = when (commits) {
-            is CommitSigBatch -> {
-                if (commits.batchSize < active.size) {
-                    return Either.Left(CommitSigCountMismatch(channelId, active.size, commits.batchSize))
-                }
-                commits.messages
-            }
-            is CommitSig -> {
-                if (active.size > 1) {
-                    return Either.Left(CommitSigCountMismatch(channelId, active.size, 1))
-                }
-                listOf(commits)
-            }
+            is CommitSigBatch -> commits.messages
+            is CommitSig -> listOf(commits)
         }
-
+        if (sigs.size < active.size) {
+            return Either.Left(CommitSigCountMismatch(channelId, active.size, sigs.size))
+        }
         // Signatures are sent in order (most recent first), calling `zip` will drop trailing sigs that are for deactivated/pruned commitments.
         val active1 = active.zip(sigs).map {
             when (val commitment1 = it.first.receiveCommit(channelKeys, params, changes, it.second, log)) {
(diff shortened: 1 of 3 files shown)


modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Commitments.kt
```diff
@@ -774,18 +774,30 @@ data class Commitments(
...
+                if (commits.batchSize < active.size) {
+                    return Either.Left(CommitSigCountMismatch(channelId, active.size, commits.batchSize))
+                }
+                commits.messages
+            }
+            is CommitSig -> {
+                if (active.size > 1) {
+                    return Either.Left(CommitSigCountMismatch(channelId, active.size, 1))
+                }
+                listOf(commits)
```
pm47, line 795:
```suggestion
        val sigs = when (commits) {
            is CommitSigBatch -> commits.messages
            is CommitSig -> listOf(commits)
        }

        if (sigs.size < active.size) {
            return Either.Left(CommitSigCountMismatch(channelId, active.size, sigs.size))
```

modules/core/src/commonMain/kotlin/fr/acinq/lightning/io/Peer.kt
```diff
@@ -472,14 +478,26 @@ class Peer(
...
-                        } catch (e: Throwable) {
-                            logger.warning { "cannot deserialize message: ${received.byteVector().toHex()}" }
+                        val msg = receiveMessage()
+                        when {
+                            msg is CommitSig && msg.batchSize > 1 -> {
+                                val others = (1 until msg.batchSize).mapNotNull { receiveMessage() as CommitSig }
+                                input.send(MessageReceived(peerConnection.id, CommitSigs.fromSigs(listOf(msg) + others)))
+                            }
+                            msg is LightningMessage -> input.send(MessageReceived(peerConnection.id, msg))
+                            else -> {}
```
pm47, line 500:
```suggestion
                        val msg = when (val msg = receiveMessage()) {
                            is CommitSig -> {
                                val others = (1 until msg.batchSize).mapNotNull { receiveMessage() as CommitSig }
                                CommitSigs.fromSigs(listOf(msg) + others)
                            }
                            else -> msg
                        }
                        msg?.let { input.send(MessageReceived(peerConnection.id, it)) }
```

modules/core/src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -1254,6 +1264,21 @@ data class CommitSig(
     }
 }
 
+data class CommitSigBatch(val messages: List<CommitSig>) : CommitSigs() {
+    // The read/write functions and the type field are meant for individual lightning messages.
+    // While we treat a commit_sig batch as one logical message, we will actually encode each messages individually.
+    // That's why the read/write functions are no-op.
```
pm47, line 1270: Maybe it would be better to make these methods throw an error?

PR #798, opened 2025-07-08
Refactor `ChannelKeys` and `KeyManager`

t-bast:
We refactor our `ChannelKeys` to use the same model as `eclair`. This is a first step before refactoring transactions and channel data similarly to what was done in `eclair` in preparation for taproot channels.

Backwards-compatibility is guaranteed by our unit tests for the local key manager and transactions building. We also have e2e tests with `eclair` with the complete taproot feature which validate these changes.


pm47 (dismissed):
Tested end-to-end against eclair.

PR #800, opened 2025-07-29
Refactor transactions for taproot compatibility

t-bast:
We refactor our transactions handling and our closing data to match the changes made in `eclair` to support taproot channels and commitment upgrades during splices.

The architecture matches the change made in eclair, up to changes from https://github.com/ACINQ/eclair/pull/3118


pm47 (commented):
LGTM:
- consistent with the eclair data model
- end-to-end tests pass against eclair

PR #807, opened 2025-09-16
Add a new peer storage format to handle legacy backups

pm47:
Our current peer storage data format is `encrypt(list(channel))`.

We add a new format: `list(encrypt(channel))`.

This allows handling of legacy proprietary per-channel backups: on the LSP side, we just need to concatenate legacy backups and set that as the peer storage, and then reject channel updates with the legacy backup method (we actually just reject connections from peers that do not support the peer storage backup).

Note that we do not yet remove the "restore from `channel_reestablish`" logic, so that we don't have to synchronize the phoenix update from the LSP update.


modules/core/src/commonMain/kotlin/fr/acinq/lightning/serialization/channel/Encryption.kt
```diff
@@ -95,9 +98,55 @@ object Encryption {
...
+     * Backward compatibility for an older backup mechanism where each channel data was stored individually, and restored from channel_reestablish messages.
+     */
+    private fun fromEncryptedPeerStorageLegacyChannelData(key: PrivateKey, encryptedPeerStorage: EncryptedPeerStorage): Result<Serialization.PeerStorageDeserializationResult> {
+        // The blob is a list of individually encrypted channel data (because each of them was previously stored separately)
+        val input = ByteArrayInput(encryptedPeerStorage.data.toByteArray())
+        val encryptedChannelDatas = input.readCollection { input.readDelimitedByteArray() }.toList()
+        val res: List<Result<Serialization.DeserializationResult>> = encryptedChannelDatas.map { data ->
+            runCatching { decrypt(key.value, data.drop(2).toByteArray()) }
+                .recoverCatching { decrypt(key.value, data) }
+                .map { Serialization.deserialize(it) }
```
pm47, line 132: This is the same contant as the `fun PersistedChannelState.Companion.from()` function above, but that function will be deleted once the migration is done, so factoring wasn't worth it.

PR #812, opened 2025-10-02
Upgrade kotlin to 2.2.0

pm47:
Let's take advantage of the upcoming major release to upgrade kotlin and other dependencies.

The main point of attention is a breaking change in `kotlinx-datetime`: see https://github.com/Kotlin/kotlinx-datetime?tab=readme-ov-file#deprecation-of-instant.


modules/core/build.gradle.kts
```diff
@@ -65,7 +65,6 @@ kotlin {
                 api("fr.acinq.bitcoin:bitcoin-kmp:${libs.versions.bitcoinkmp.get()}")
                 api("org.jetbrains.kotlinx:kotlinx-coroutines-core:${libs.versions.kotlinx.coroutines.get()}")
                 api("org.jetbrains.kotlinx:kotlinx-serialization-core:${libs.versions.kotlinx.serialization.get()}")
-                api("org.jetbrains.kotlinx:kotlinx-serialization-cbor:${libs.versions.kotlinx.serialization.get()}")
```
pm47, line 68: @dpad85 This will need to be added to phoenix build.gradle.kts.

PR #816, opened 2025-10-07
Fixing bugs & shortcomings in OfferPaymentMetadata.V1

robbiehanson:
There are a few bugs in `OfferPaymentMetadata.V1`. For example, say we generate an offer with the description: "Pizza 🍕":

badflow_1

The description may end up within`OfferPaymentMetadata.V1` as a `payerNoter`. This is incorrect:

badflow_3

It's not a payerNote, and the UI needs to understand this. Further, if the user does add a payerNote, then the original description will be lost. Which may be important information for the user. For example, the original description may have been "Invoice #: 472648"

## General idea

```kotlin
data class V1(
    override val offerId: ByteVector32,
    override val amount: MilliSatoshi,
    override val preimage: ByteVector32,
    val payerKey: PublicKey,
    val payerNote: String?,
    val quantity: Long,
    val createdAtMillis: Long
)

data class V2(
    override val offerId: ByteVector32,
    override val amount: MilliSatoshi,
    override val preimage: ByteVector32,
    override val createdAtSeconds: Long,
    override val relativeExpirySeconds: Long?,
    val description: String?,
    val payerKey: PublicKey?,
    val payerNote: String?,
    val quantity: Long?,
)
```

## Optimizations

There are several optimizations we can perform to shrink the size of the encoded metadata:

* the quantity field can be null (to match `quantity_opt`). This saves us 8 bytes in the common case.
* we can store the timestamp in seconds instead of milliseconds. This matches the timestamp fields in the Bolt12 invoice. And allows us to use `bigSize`, which shrinks the size from a fixed 8 bytes to 5 bytes (at least for another ~100 years).
* the `amount: MilliSatoshi` can be stored as `bigSize`, which shrinks the size from a fixed 8 bytes, to an average of 5 bytes. It only uses more bytes if the payment amount is over ~0.043 BTC, and most payments are under this amount.

The end result is that V2 is smaller than V1.

## Future-proofing

I'm experimenting in another branch (`card-payment`), and I need to create a Bolt12Invoice that expires in 30 seconds. But that's wasn't possible, because the `relativeExpiry` wasn't stored in `OfferPaymentMetadata.V1`, and the default value was hard-coded. So I fixed that issue.

Also in the branch, I'm experimenting with creating an "unsolicited Bolt 12 invoice". That is, a Bolt 12 Invoice that is generated not in response to an InvoiceRequest, but due to an attempted card payment. As such, it does not have a `payerKey`. So I've made that field optional in V2.

Note: Making it optional takes up no extra bytes.

## Alternative implementations

This PR fixes all the issues I have with V1, but it may not be the ideal solution. Ultimately, one of the core issues I see is that, while `LightningOutgoingPayment` stores the full `Bolt12Invoice`, `Bolt12IncomingPayment` only stores the `OfferPaymentMetadata`.

Perhaps the ideal solution would be:
* Make `Bolt12IncomingPayment` store the full `Bolt12Invoice`
* Change `OfferPaymentMetadata` to be simply a signature over a subset of the `offer_*` & `invoice_*` fields. And verify the sig when receiving the payment part(s).

However... that's a very big change. Implementing it will take a lot of effort. And we're all busy with other big changes at the moment. So in the meantime, I'm proposing the minor changes in V2.

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/payment/OfferPaymentMetadata.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/payment/OfferPaymentMetadata.kt
@@ -100,8 +100,6 @@ sealed class OfferPaymentMetadata {
         }
 
         companion object {
-            val minLength: Int get() = 121
-
             fun read(input: Input): V1 {
                 val offerId = LightningCodecs.bytes(input, 32).byteVector32()
                 val amount = LightningCodecs.u64(input).msat
@@ -115,6 +113,7 @@ sealed class OfferPaymentMetadata {
         }
     }
 
+    /** In this version, we encrypt the payment metadata with a key derived from our seed. */
     data class V2(
         override val offerId: ByteVector32,
         override val amount: MilliSatoshi,
@@ -125,7 +124,6 @@ sealed class OfferPaymentMetadata {
         val payerKey: PublicKey?,
         val payerNote: String?,
         val quantity: Long?,
-
     ) : OfferPaymentMetadata() {
         override val version: Byte get() = 2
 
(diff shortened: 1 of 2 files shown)


thomash-acinq (approved):
Nice improvement

PR #817, opened 2025-10-08
Increase maximum number of accepted HTLCs to 19

sstone:
https://github.com/ACINQ/lightning-kmp/pull/800 included a lot of improvements on our data structure and removed data that was either redondant or could be re-computed. This means that serialised channel states are smaller, and we can increase the number of pending HTLCs to 19 and stay below the 65 Kb size limit for LN messages.

Note that this applies to new channels only: existing ones will still use the previous value of 12 pending HTLCs.


t-bast (approved):
It's a more than 50% improvement :rocket:

PR #853, opened 2026-03-26
ElectrumMiniWalletTest: use mock testnet3 server

sstone:
testnet3 electrum servers are unreliable and cause tests and CI to fail. We replace our testnet3 server by as mock that uses data returned by a real testnet3 server.


t-bast (commented):
I think that can be quite useful. It won't catch issues caused by changes in servers though, ideally we'd want:

- to run these tests with the mock during standard CI
- but to run them with the real testnet3 servers during a nightly or weekly run

That wouldn't require much code, but I don't know if kotlin's test framework allows skipping or including specific tests when running, the way Scalatest does?

sstone:
replaced by #854

PR #854, opened 2026-03-31
Use docker instead of testnet3 for some ElectrumMiniWallet tests

sstone:
testnet3 electrum servers are unreliable which caused our CI to fail. Tests that required a testnet3 electrum servers (to verify how we handle gap in address history) have been modified to run against a local regtest docker electrum server, like our ElectrumWatcher integration tests.
We also upgrade bitcoind docker image from 0.20.1 (which did not understand addresses use in our swap-in protocol)  to 29.1.

Replaces https://github.com/ACINQ/lightning-kmp/pull/853


t-bast (commented):
LGTM, just some nits

t-bast (approved):

PR #855, opened 2026-04-02
Follow BOLT1 handling for "no reply" pings: ignore, don't reply.

rustyrussell:
```
A node receiving a `ping` message:
  - if `num_pong_bytes` is less than 65532:
    - MUST respond by sending a `pong` message, with `byteslen` equal to `num_pong_bytes`.
  - otherwise (`num_pong_bytes` is **not** less than 65532):
    - MUST ignore the `ping`.
```

This allows for message padding.

See https://github.com/ACINQ/eclair/pull/3278 for the eclair variant.


t-bast (approved):
Thanks!

PR #856, opened 2026-04-15
Simple close workflow: use our closing_complete instead of local signing nonces

sstone:
In the simple close session, which is never persisted, we just need to remember our closing_complete and the partial signatures it contains, instead of keeping our musig2 nonces and generating our partial signatures a second time to build the final closing tx when we receive our peer's partial signatures.

We also check our peer's partial signature and fail early if they're not valid, instead of building a fully sign closing tx and then verify that it is correct.

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
@@ -4,7 +4,6 @@ import fr.acinq.bitcoin.*
 import fr.acinq.bitcoin.crypto.musig2.IndividualNonce
 import fr.acinq.bitcoin.utils.Either
 import fr.acinq.bitcoin.utils.Try
-import fr.acinq.bitcoin.utils.flatMap
 import fr.acinq.bitcoin.utils.runTrying
 import fr.acinq.lightning.CltvExpiryDelta
 import fr.acinq.lightning.MilliSatoshi
@@ -313,7 +312,6 @@ object Helpers {
                 return Either.Left(CannotGenerateClosingTx(commitment.channelId))
             }
             val localFundingKey = channelKeys.fundingKey(commitment.fundingTxIndex)
-            val localNonces = Transactions.CloserNonces.generate(localFundingKey.publicKey(), commitment.remoteFundingPubkey, commitment.fundingTxId)
             val tlvs = when (commitment.commitmentFormat) {
                 Transactions.CommitmentFormat.AnchorOutputs -> TlvStream(
                     setOfNotNull(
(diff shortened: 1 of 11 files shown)


modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
```diff
@@ -289,7 +289,7 @@ object Helpers {
             feerate: FeeratePerKw,
             lockTime: Long,
             remoteNonce: IndividualNonce?
-        ): Either<ChannelException, Triple<Transactions.ClosingTxs, ClosingComplete, Transactions.CloserNonces>> {
```
t-bast, line 292: I believe you can completely remove the `CloserNonces` class from the codebase like what you did in `eclair`.

modules/core/src/commonMain/kotlin/fr/acinq/lightning/channel/Helpers.kt
```diff
@@ -451,7 +454,7 @@ object Helpers {
             commitment: FullCommitment,
             closingTxs: Transactions.ClosingTxs,
             closingSig: ClosingSig,
-            localNonces: Transactions.CloserNonces?,
+            localCosingComplete: ClosingComplete?,
```
t-bast: nit:

```suggestion
            localClosingComplete: ClosingComplete?,
```

modules/core/src/commonTest/kotlin/fr/acinq/lightning/channel/InteractiveTxTestsCommon.kt
```diff
@@ -36,114 +36,173 @@ class InteractiveTxTestsCommon : LightningTestSuite() {
         assertEquals(f.fundingParamsA.fundingAmount, fundingA + fundingB)
         assertEquals(f.fundingParamsA.fundingAmount, fundingA + fundingB)
 
-        val alice0 = InteractiveTxSession(f.nodeIdB, f.channelKeysA, f.keyManagerA.swapInOnChainWallet, f.fundingParamsA, 0, 0.msat, 0.msat, emptySet(), f.fundingContributionsA)
```
t-bast, line 39: I don't understand the changes to this test: they seem completely unrelated to the closing nonce change? And I'm not sure what they add? Can you explain?

sstone: These change re-generate test data so that the closing musig nonce is never reused. It's not needed yet and I can revert this change and wait until we upgrade to a version of `bitcoin-kmp` that includes https://github.com/ACINQ/bitcoin-kmp/pull/177.

t-bast: I don't see the point of making tests more complex (those tests are already somewhat hard to read) in order to test this, can you revert changes to this file? Testing normal flows is enough to guarantee that there is no nonce reuse (there wasn't any nonce reuse even before your change). It's not the responsibility of `lightning-kmp` to verify that nonce reuse is not possible, it's the responsibility of `bitcoin-kmp` which already tests this.

sstone: Reverted in https://github.com/ACINQ/lightning-kmp/pull/856/changes/a62a381493b77db913485bb7f79c366e051b75a1, we'll change these tests when we upgrade to single-use nonces.

t-bast (approved):

PR #858, opened 2026-05-19
Upgrade to bitcoin-kmp 0.31.0

sstone:
Main changes are:
- some static methods are been removed, users must now use instance methods (for example: `DeterministicWallet.publicKey(key)` is now `key.extendedPublicKey`) which are must easier to use from IDEs
- musig2 secret nonces can only be used once, this mostly impacts unit tests

This PR uses a snapshot version of `bitcoin-kmp`  0.31.0 and will be updated once it has been released.


modules/core/src/commonTest/kotlin/fr/acinq/lightning/channel/InteractiveTxTestsCommon.kt
```diff
@@ -36,114 +36,172 @@ class InteractiveTxTestsCommon : LightningTestSuite() {
         assertEquals(f.fundingParamsA.fundingAmount, fundingA + fundingB)
         assertEquals(f.fundingParamsA.fundingAmount, fundingA + fundingB)
 
-        val alice0 = InteractiveTxSession(f.nodeIdB, f.channelKeysA, f.keyManagerA.swapInOnChainWallet, f.fundingParamsA, 0, 0.msat, 0.msat, emptySet(), f.fundingContributionsA)
```
t-bast, line 39: I think we've discussed it before: can you revert the changes to this test? It's unrelated to the version update as far as I understand, and I don't see why it's useful (apart from being hard to review!).

The musig2 nonce reuse doesn't seem to be an issue with the previous code for that test, and I don't think we need to test nonce reuse from `lightning-kmp`, it's `bitcoin-kmp`'s responsibility and is already tested there.

sstone: I was probably not clear enough about the reason for these changes: this specific test reuses the same    interactive sessions several times with valid and invalid inputs (missing signatures etc...). 
Since `bitcoin-kmp` 0.31 forbids secret nonce reuse, without changes this test will fail at `// The resulting transaction is valid and has the right feerate.`. And the `assertNull` checks above for various invalid signatures will fail for the wrong reasons (because we're reusing a secret nonce, not because inputs are invalid).

sstone: I could not find a simple way to "clone" interactive sessions without doing the full "send inputs/outputs/complete" setup (copy/serialization/reflection won't work or would be really ugly) but I moved the "invalid signatures" checks to their own test, which may be easier to review amd makes sense ("initiator funds more than non-initiator" was doing different unrelated checks in the same test).

t-bast: Thanks, it's easier to review that way and only focus on the error cases :+1:

t-bast (approved):

PR #860, opened 2026-06-30
Implement TLS for linux sockets

sstone:
We add a cinterop wrapper for [ruslt-ffi](https://github.com/rustls/rustls-ffi)  to implement TLS for linux sockets. 
[ruslt-ffi](https://github.com/rustls/rustls-ffi) provides c-binding for [rustls](https://docs.rs/rustls/latest/rustls/), a TLS implementation written in rust.

[ruslt-ffi](https://github.com/rustls/rustls-ffi) was chosen because it is much simpler to use that openssl, and can be compiled into a single no-dependencies static library. This PR includes compiled `rustls-ffi` binaries for linux X64 and linux Arm64 to keep our build process simple. Please note that these libraries are fairly big (about 33 Mb).


modules/core/src/macosMain/kotlin/fr/acinq/lightning/io/rustls/KtorTlsTcpSocket.kt
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * A TLS connection to a server, layered over a raw Ktor TCP [Socket]. rustls runs
+ * the TLS state machine; this class pumps ciphertext between rustls and the socket
+ * and exposes plaintext [write]/[read].
+ *
+ * Not safe for concurrent use; drive it from a single coroutine.
+ */
+@OptIn(ExperimentalForeignApi::class)
+class KtorTlsTcpSocket (
```
pm47: This class (and the file) should be called `RustTlsTcpSocket` since it is not using the ktor impl:

```suggestion
class RustTlsTcpSocket (
```

sstone: done in https://github.com/ACINQ/lightning-kmp/pull/860/changes/97a2fafba03cbffa71b127bab9525dcf93714be0

sstone:
I checked that this solution also works for macos x64 and arm64 (i.e. all tests pass) but using Apple's library (see https://github.com/ACINQ/lightning-kmp/pull/861) would be even better.

gradle.properties
```diff
@@ -11,3 +11,4 @@ kotlin.mpp.enableCInteropCommonization=true
 # dokka migration to v2
 org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled
 org.jetbrains.dokka.experimental.gradle.pluginMode.noWarn=true
+kotlin.native.cacheKind.linuxX64=none
```
pm47: Nit: should be moved 3 lines above, in the kotlin section

pm47:
Needs rebase :pray:

gradle.properties
sstone: Done in https://github.com/ACINQ/lightning-kmp/pull/860/changes/251b1d98e1cb64f3f13807baffb32dc8828f6f2d

sstone:
Rebased on master.

modules/core/build.gradle.kts
```diff
@@ -282,17 +297,6 @@ tasks.withType<AbstractTestTask> {
...
-tasks
-    .filterIsInstance<KotlinNativeTest>()
-    .filter { it.name == "linuxX64Test" }
-    .map {
-        it.filter.excludeTestsMatching("*IntegrationTest")
-        it.filter.excludeTestsMatching("*ElectrumClientTest")
-        it.filter.excludeTestsMatching("*ElectrumMiniWalletTest")
-        it.filter.excludeTestsMatching("*SwapInWalletTestsCommon")
-    }
-
```
pm47, line 295: Nice

pm47 (approved):

PR #861, opened 2026-07-01
Move `iosMain` code to `appleMain`

pm47:
Those platforms are almost the same and code can be factored.

As a nice side effect, this adds TCP TLS support and optimized `Chacha20Poly` to `macOs`.

PR #864, opened 2026-07-02
Simplify and improve `runSuspendTest`

pm47:
Follow-up to #861.


modules/core/src/commonTest/kotlin/fr/acinq/lightning/tests/utils/runTest.kt
```diff
@@ -1,26 +1,29 @@
...
+    }
+    try {
+        withTimeout(timeout) { testJob.join() }
+    } catch (e: TimeoutCancellationException) {
+        testJob.cancel()
+        // wait for the test's coroutines to complete their cancellation cleanup: anything
+        // they print after the test is reported finished confuses gradle's test runner
+        testJob.join()
+        fail("test did not complete within $timeout")
     }
```
sstone: This will swallow any `CancellationException` as normal child cancellation (test silently passed). Notably, with bitcoind down, `ElectrumMiniWalletIntegrationTest`'s inner `withTimeout(10.seconds)` (still present
at that commit) would throw a `TimeoutCancellationException`.

```suggestion
    // withTimeoutOrNull only swallows its own timeout: an exception thrown by the test body — including a
    // TimeoutCancellationException from a withTimeout inside the test — propagates and fails the test
    withTimeoutOrNull(timeout) {
        test()
        // the test body has returned: cancel whatever it left running in its scope, otherwise this scope
        // would never complete; withTimeoutOrNull only resumes once those coroutines have finished their
        // cancellation cleanup, so nothing they print can be misattributed by gradle's test runner
        coroutineContext.cancelChildren()
    } ?: fail("test did not complete within $timeout")
```

modules/core/src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWalletIntegrationTest.kt
```diff
@@ -35,7 +28,7 @@ class ElectrumMiniWalletIntegrationTest : LightningTestSuite() {
     private val keyManager = LocalKeyManager(MnemonicCode.toSeed(mnemonics, "").toByteVector(), Chain.Regtest, TestConstants.aliceSwapInServerXpub)
 
     init {
-        runSuspendBlocking {
+        runSuspendTest {
```
sstone: Suggestion for the whole init block:
```scala
    init {
        runSuspendTest(timeout = 10.seconds) {
            while (runTrying { bitcoincli.getNetworkInfo() }.isFailure) {
                delay(0.5.seconds)
            }
        }
        runSuspendTest {
            val address = bitcoincli.getNewAddress()
            val address0 = keyManager.swapInOnChainWallet.getSwapInProtocol(0).address(Chain.Regtest)
            bitcoincli.sendToAddress(address0, 1.0)
            val address2 = keyManager.swapInOnChainWallet.getSwapInProtocol(2).address(Chain.Regtest)
            bitcoincli.sendToAddress(address2, 1.0)
        }
    }
```

sstone (changes requested):

modules/core/src/commonTest/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumMiniWalletIntegrationTest.kt
pm47: Done in 0da7f886.

modules/core/src/commonTest/kotlin/fr/acinq/lightning/tests/utils/runTest.kt
pm47: Done in 0da7f886.

sstone (approved):

PR #871, opened 2026-08-26
Store the right `commit_sig` signature

t-bast:
Signatures can either be ECDSA signatures for non-taproot channels, or partial musig2 signatures for taproot channels. We make sure we don't get confused between the two types of signatures and always store and validate the one that matches our commitment format, which guarantees that we're able to broadcast our commitment transaction.

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
@@ -12,6 +12,7 @@ import fr.acinq.lightning.channel.ChannelFlags
 import fr.acinq.lightning.channel.ChannelSpendSignature
 import fr.acinq.lightning.channel.ChannelType
 import fr.acinq.lightning.router.Announcements
+import fr.acinq.lightning.transactions.Transactions
 import fr.acinq.lightning.utils.*
 import fr.acinq.secp256k1.Hex
 import io.ktor.utils.io.charsets.*
@@ -1314,8 +1315,12 @@ data class CommitSig(
     override val type: Long get() = CommitSig.type
 
     val partialSignature: ChannelSpendSignature.PartialSignatureWithNonce? = tlvStream.get<CommitSigTlv.PartialSignatureWithNonce>()?.psig
-    val sigOrPartialSig: ChannelSpendSignature = partialSignature ?: signature
     val fundingTxId: TxId? = tlvStream.get<CommitSigTlv.FundingTx>()?.txId
+
+    fun signatureFor(commitmentFormat: Transactions.CommitmentFormat): ChannelSpendSignature? = when (commitmentFormat) {
+        Transactions.CommitmentFormat.AnchorOutputs -> signature
+        Transactions.CommitmentFormat.SimpleTaprootChannels -> partialSignature
+    }
 
     override fun write(out: Output) {
         LightningCodecs.writeBytes(channelId, out)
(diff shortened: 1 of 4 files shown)


modules/core/src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
```diff
@@ -1314,9 +1315,13 @@ data class CommitSig(
     override val type: Long get() = CommitSig.type
 
     val partialSignature: ChannelSpendSignature.PartialSignatureWithNonce? = tlvStream.get<CommitSigTlv.PartialSignatureWithNonce>()?.psig
-    val sigOrPartialSig: ChannelSpendSignature = partialSignature ?: signature
     val fundingTxId: TxId? = tlvStream.get<CommitSigTlv.FundingTx>()?.txId
 
+    fun signatureFor(commitmentFormat: Transactions.CommitmentFormat): ChannelSpendSignature? = when (commitmentFormat) {
+        Transactions.CommitmentFormat.AnchorOutputs -> signature
```
sstone, line 1321: nit: to be consistent with `eclair` we could also use:
```suggestion
        Transactions.CommitmentFormat.AnchorOutputs -> if (partialSignature == null) signature else null
```

sstone (commented):
we should also update `Helpers.receiveClosingSig()` with:

```kotlin
            val closingTxsWithSig = when (commitment.commitmentFormat) {
                Transactions.CommitmentFormat.AnchorOutputs -> {
                    listOfNotNull(
                        closingSig.closerAndCloseeOutputsSig?.let { sig -> closingTxs.localAndRemote?.let { tx -> Pair(tx, ChannelSpendSignature.IndividualSignature(sig)) } },
                        closingSig.closerOutputOnlySig?.let { sig -> closingTxs.localOnly?.let { tx -> Pair(tx, ChannelSpendSignature.IndividualSignature(sig)) } },
                        closingSig.closeeOutputOnlySig?.let { sig -> closingTxs.remoteOnly?.let { tx -> Pair(tx, ChannelSpendSignature.IndividualSignature(sig)) } },
                    )
                }

                Transactions.CommitmentFormat.SimpleTaprootChannels -> {
                    listOfNotNull(
                        closingSig.closerAndCloseeOutputsPartialSig?.let { sig -> remoteNonce?.let { nonce -> closingTxs.localAndRemote?.let { tx -> Pair(tx, ChannelSpendSignature.PartialSignatureWithNonce(sig, nonce)) } } },
                        closingSig.closerOutputOnlyPartialSig?.let { sig -> remoteNonce?.let { nonce -> closingTxs.localOnly?.let { tx -> Pair(tx, ChannelSpendSignature.PartialSignatureWithNonce(sig, nonce)) } } },
                        closingSig.closeeOutputOnlyPartialSig?.let { sig -> remoteNonce?.let { nonce -> closingTxs.remoteOnly?.let { tx -> Pair(tx, ChannelSpendSignature.PartialSignatureWithNonce(sig, nonce)) } } },
                    )
                }
            }

```

modules/core/src/commonMain/kotlin/fr/acinq/lightning/wire/LightningMessages.kt
t-bast: I honestly don't see why it's useful, it's more code and it's a bit clunky? As long as we have added unit tests that verify this, especially in `lightning-kmp` we don't need to be as aggressive about force-closing. The BOLTs don't say that we must reject this case, I think that in `eclair` it makes sense because we want to be more aggressive towards attackers, but `lightning-kmp` only connects to LSPs that we choose, so we can opt for simpler code?

t-bast:
> we should also update Helpers.receiveClosingSig() with

Good catch, done in https://github.com/ACINQ/lightning-kmp/pull/871/commits/1d5e0042e87516f849cea5d1434dba7f5e5083a0

sstone (approved):

PR #873, opened 2026-09-11
Expose payment failure categories

ca-ruz:
## Summary

  This PR adds a stable high-level category API for payment failures.

  It exposes categories on:

  - `FinalFailure`
  - `LightningOutgoingPayment.Part.Status.Failed.Failure`
  - `OutgoingPaymentFailure`
  - `Bolt12InvoiceRequestFailure`

  The category enum is shared as `PaymentFailureCategory`, so callers can classify both regular outgoing payment failures and BOLT 12 invoice request failures with the same API.

  ## Context

  This came out of debugging outgoing Lightning payment failures in Ambrosia, which uses phoenixd as a Lightning backend.

  In that investigation, phoenixd accepted outgoing payment attempts, selected a local channel, sent an HTLC, and then received `UpdateFailHtlc`. The existing phoenixd API exposed a human-readable `reason`, but downstream applications could not reliably classify the failure without parsing strings or duplicating lightning-kmp internals.

  A ACINQ/phoenixd#240 PR initially added failure categorization in phoenixd, but review feedback correctly pointed out that phoenixd should remain a thin wrapper and that this classification belongs in lightning-kmp.

  There was also feedback to handle the BOLT 12 case, where paying an offer may fail before a payable invoice is obtained. This PR therefore exposes the same category type for `Bolt12InvoiceRequestFailure`.

  ## Why This Is Useful

  Although the need was discovered while debugging Ambrosia, this is useful for any application built on lightning-kmp, including phoenixd and other wallets/services.

  Callers can now expose structured diagnostics, improve logs, and choose better user-facing messages without depending on localized strings or fragile class-name mappings.

  For example, clients can distinguish:

  - invalid request or payment parameters
  - insufficient local balance
  - unavailable local channels
  - insufficient routing fees
  - CLTV/expiry issues
  - recipient liquidity problems
  - recipient rejection or unreachability
  - remote route failures
  - exhausted/interrupted retries
  - unknown failures

  ## Compatibility

  This is a derived API only.

  - no payment behavior changes
  - no database migration
  - no serialization format change
  - existing failure types and messages remain unchanged

  ## Tests

  Added focused tests covering category mappings for final failures, payment part failures, aggregate `OutgoingPaymentFailure` explanations, and BOLT 12 invoice request failures.

  The tests include the cases that motivated the phoenixd work:

  - `TrampolineFeeInsufficient`
  - `FeeInsufficient`
  - `TemporaryChannelFailure`
  - `UnknownNextPeer`
  - `PaymentTimeout`
  - `RetryExhausted`
  - local insufficient balance
  - BOLT 12 invoice request timeout

  Tested with:

  ```bash
  ./gradlew :lightning-kmp-core:jvmTest \
    --tests fr.acinq.lightning.payment.OutgoingPaymentFailureTestsCommon \
    --tests fr.acinq.lightning.payment.OfferManagerTestsCommon

### DIFF
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/db/PaymentsDb.kt
@@ -450,6 +450,23 @@ data class LightningOutgoingPayment(
                  * Applications should define their own localized message for each of these failure cases.
                  */
                 sealed class Failure {
+                    val category: OutgoingPaymentFailure.Category
+                        get() = when (this) {
+                            PaymentAmountTooSmall -> OutgoingPaymentFailure.Category.LocalValidation
+                            PaymentAmountTooBig -> OutgoingPaymentFailure.Category.LocalValidation
+                            NotEnoughFunds -> OutgoingPaymentFailure.Category.LocalBalance
+                            NotEnoughFees -> OutgoingPaymentFailure.Category.Fee
+                            PaymentExpiryTooBig -> OutgoingPaymentFailure.Category.Cltv
+                            TooManyPendingPayments -> OutgoingPaymentFailure.Category.Retry
+                            ChannelIsSplicing -> OutgoingPaymentFailure.Category.LocalChannel
+                            ChannelIsClosing -> OutgoingPaymentFailure.Category.LocalChannel
+                            TemporaryRemoteFailure -> OutgoingPaymentFailure.Category.Remote
+                            RecipientLiquidityIssue -> OutgoingPaymentFailure.Category.Liquidity
+                            RecipientIsOffline -> OutgoingPaymentFailure.Category.Recipient
+                            RecipientRejectedPayment -> OutgoingPaymentFailure.Category.Recipient
+                            is Uninterpretable -> OutgoingPaymentFailure.Category.Unknown
+                        }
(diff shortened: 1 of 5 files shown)


dpad85 (commented):
I think the categories are too fine-grained while not giving actionable information. It's tempting to just repeat the payment error under another label.

Instead, I suggest fewer categories, based on the source and what the user can actually do.

Just by looking at the category you'd get a rough idea of where it happened and what can be done. Then, for details and to provide pertinent advice to the user, the API/Wallet would check the underlying error (I think you would need to do that anyway, even with more categories).

Possibilities are:
- it's on your end, but you can retry if you do something;
- it's on your end, and you cannot retry;
- it's not on your end, but you can retry;
- it's not on your end, and you cannot retry.

We would end up with 6 categories:

- `LocalFatal`: this payment cannot be done. E.g., you already paid it.
- `LocalTransient`: adjust something (the amount, wait for channels to be established, ...) then you can try again
- `InflightTransient`: something happened on the way, just try again
- `NotEnoughFee`: increase the fee, and try again.
- `RemoteFatal`: the payment cannot be done. E.g., the recipient actively rejected it.
- `Unknown`: deserves its own category, but in effect the user should likely try again

Not sure about `NotEnoughFee`, maybe it could be put under `InflightTransient`.

PR #876, opened 2026-09-16
Remove the status cache in `ElectrumWatcher`

pm47:
The watcher only fetched the history of a script hash when its status changed. When a watch was added on a script hash we were already subscribed to, transactions already in the history were never evaluated against the new watch, and the server never notifies us again since the status doesn't change.

The status cache is removed entirely: this was its only use, and it saved a few requests at most.


modules/core/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
```diff
@@ -153,19 +158,25 @@ class ElectrumWatcher(val client: IElectrumClient, val scope: CoroutineScope, lo
             )
             if (state.isConnected) {
                 val response = client.startScriptHashSubscription(scriptHash)
-                processScripHashSubscriptionResponse(response)
+                // We may already be subscribed to this script hash, (e.g. we watched a funding output for confirmation, and now watch
```
sstone: There is no guard around `getScriptHashHistory()` and we will subscribe again to script hasheds we're already subscribed to, I propose the following changes:
```diff
diff --git a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
index ff2024b2..6ba1495d 100644
--- a/modules/core/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
+++ b/modules/core/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
@@ -153,17 +153,21 @@ class ElectrumWatcher(val client: IElectrumClient, val scope: CoroutineScope, lo
                     scriptHash
                 }
             }
+            // Subscriptions are per-connection: we reset them whenever we (re)connect, so this tells us whether the server is
+            // already sending us notifications for that script hash.
+            val alreadySubscribed = state.scriptHashSubscriptions.contains(scriptHash)
             state = state.copy(
                 watches = state.watches + watch, scriptHashSubscriptions = state.scriptHashSubscriptions + scriptHash
             )
             if (state.isConnected) {
-                val response = client.startScriptHashSubscription(scriptHash)
-                // We may already be subscribed to this script hash, (e.g. we watched a funding output for confirmation, and now watch
-                // it for spending): we must check the existing history.
-                if (response.status != null) {
-                    val history = client.getScriptHashHistory(scriptHash)
-                    processScripHashHistory(history, listOf(watch))
-                    state = state.copy(idleSince = currentTimestampMillis())
+                runCatching {
+                    // We may already be subscribed to this script hash, (e.g. we watched a funding output for confirmation, and now watch
+                    // it for spending): we must check the existing history, but we don't subscribe again
+                    if (alreadySubscribed || client.startScriptHashSubscription(scriptHash).status != null) {
+                        val history = client.getScriptHashHistory(scriptHash)
+                        processScripHashHistory(history, listOf(watch))
+                        state = state.copy(idleSince = currentTimestampMillis())
+                    }
                 }
             }
         }

```

pm47: Applied (slightly modified) in [9f07fa5](https://github.com/ACINQ/lightning-kmp/pull/876/commits/9f07fa59fe21ee9a4130e039bbd59d253d3c3a95).

sstone: `alreadySubscribed` must be computed before `state = state.copy(
                 watches = state.watches + watch, scriptHashSubscriptions = state.scriptHashSubscriptions + scriptHash)` as in the patch above,  with 9f07fa5 it's always true that's why `ElectrumWatcherIntegrationTest` fails

pm47: Embarrassing... Fixed in 9f07fa59.

sstone (approved):

PR #880, opened 2026-09-17
Retry pending publishes on every new block, not just on reconnection

sstone:
When publishing a delayed transaction, we now fetch its parent to get the script of the output it spends. If that fetch fails we put the transaction back into the publish queue, but that queue was only drained on reconnection: `getTx` returns null after two 10s timeouts or on any electrum error, without losing the connection, so on a long-lived connection the transaction was dropped and no `ParentTxConfirmed` watch was ever registered.

We now also drain the publish queue whenever we receive a new block, which bounds the retry delay to one block instead of one reconnection.

The publish handler is extracted into `processPublish`, which lets the queue be drained by calling it directly instead of re-sending to our own mailbox: that self-send could deadlock the watcher if the queue exceeded the mailbox buffer. The queue is cleared before being replayed, since `processPublish` puts failed transactions back into it.


modules/core/src/commonMain/kotlin/fr/acinq/lightning/blockchain/electrum/ElectrumWatcher.kt
```diff
@@ -158,6 +158,53 @@ class ElectrumWatcher(val client: IElectrumClient, val scope: CoroutineScope, lo
...
+                        state = state.copy(sent = state.sent + tx)
+                    }
+                }
+            }
+        }
+
+        suspend fun processPublishQueue() {
+            val pending = state.publishQueue
+            state = state.copy(publishQueue = setOf())
+            pending.forEach { processPublish(it) }
```
pm47, line 205: The current mutable `state` approach makes us do that kind of ugly things. It's a separate concern though.

pm47 (approved):

